PR #4, opened 2022-04-26
sketch uniform conversion with host assistance

graydon:
Implements a uniform host-assisted conversion interface both for ValTypes and those which need a little more help.

### DIFF
--- a/src/host.rs
+++ b/src/host.rs
@@ -1,12 +1,78 @@
+use stellar_xdr::ScObjectType;
+
+use crate::{or_abort::OrAbort, ValType};
+
 use super::{Object, Val};
 use core::any;
+
+// HostConvertable is similar to ValType but also covers types with conversions
+// that need a Host -- those that might require allocating an Object. ValType
+// covers types that can always be directly converted to Val with no Host.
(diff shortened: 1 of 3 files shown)


graydon:
cc @leighmcculloch

src/host.rs
```diff
@@ -1,12 +1,78 @@
...
+
+use crate::{or_abort::OrAbort, ValType};
+
 use super::{Object, Val};
 use core::any;
 
+// HostConvertable is similar to ValType but also covers types with conversions
+// that need a Host -- those that might require allocating an Object. ValType
+// covers types that can always be directly converted to Val with no Host.
+pub trait HostConvertable: Sized {
```
leighmcculloch, line 11: Looks good. Given these traits and types used for conversions are all zero-cost, why wouldn't we have a single trait for types that require the host or not? Types that don't need the host just wouldn't use them. From the SDK's pov this would provide a simpler interface to application developers.

leighmcculloch: Also, I assume by these types not being ValType, this means they can't be used inside Maps/Vecs? I think that was the issue @jonjove originally experienced with our old model where types were treated differently.

PR #14, opened 2022-04-30
Update the xdr

leighmcculloch:
### What

Update stellar-xdr to the latest commit. Update locations that used variable length arrays and opaques to use the `VecM` type.

### Why

Keep it up-to-date. Also, I'm starting work on the host interface, and I need the host and sdk libs to be using the same version of the XDR lib.

### Known limitations

The VecM type has some traits and functions implemented like TryFrom from a few types (Vec, slice, arrays, etc), AsRef (for slices), and a few functions like `iter`, `as_vec`, `to_vec`. And I noticed that `()` is commonly used as an error type in this crate, so I've also added a From impl for converting the errors from the xdr lib into `()` so we can map automatically. 

Please let me know if there are other conveniences I should add, specifically for the VecM.

### DIFF
--- a/src/host_context.rs
+++ b/src/host_context.rs
@@ -74,7 +74,7 @@ impl HostContext {
                 Tag::Symbol => {
                     let sym: Symbol = unsafe { <Symbol as ValType>::unchecked_from_val(val) };
                     let str: String = sym.into_iter().collect();
-                    Ok(ScVal::ScvSymbol(str.as_bytes().into()))
+                    Ok(ScVal::ScvSymbol(str.as_bytes().try_into()?))
                 }
                 Tag::BitSet => Ok(ScVal::ScvBitset(val.get_payload())),
                 Tag::Status => {
@@ -127,7 +127,7 @@ impl HostContext {
                         for e in vv.iter() {
                             sv.push(self.from_host_val(e.val)?);
                         }
-                        Ok(ScObject::ScoVec(ScVec(sv)))
+                        Ok(ScObject::ScoVec(ScVec(sv.try_into()?)))
                     }
                     HostObject::Map(mm) => {
                         let mut mv = Vec::new();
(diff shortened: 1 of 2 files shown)


src/host_context.rs
```diff
@@ -178,13 +178,13 @@ impl HostContext {
...
             ScObject::ScoString(s) => {
-                let ss = match String::from_utf8(s.clone()) {
+                let ss = match String::from_utf8(s.clone().into()) {
                     Ok(ss) => ss,
                     Err(_) => return Err(()),
                 };
                 self.add_host_object(ss)
             }
-            ScObject::ScoBinary(b) => self.add_host_object(b.clone()),
+            ScObject::ScoBinary(b) => self.add_host_object::<Vec<u8>>(b.clone().into()),
```
leighmcculloch, line 187: @graydon Is there a trait I can implement, or a fn that is convention you'd recommend for making these types of calls (`.clone().into()`) more convenient, so we do not need to always specify the destination host object type?

src/host_context.rs
```diff
@@ -368,10 +368,10 @@ mod test {
     }
 
     #[test]
-    fn vec_host_objs() {
+    fn vec_host_objs() -> Result<(), ()> {
         let mut host = HostContext::default();
-        let scvec0: ScVec = ScVec(vec![ScVal::ScvU32(1)]);
-        let scvec1: ScVec = ScVec(vec![ScVal::ScvU32(1)]);
+        let scvec0: ScVec = ScVec(vec![ScVal::ScvU32(1)].try_into()?);
+        let scvec1: ScVec = ScVec(vec![ScVal::ScvU32(1)].try_into()?);
```
leighmcculloch, line 374: It's a-shame we can't avoid the `.try_into()?` because the vec on the left is of const size. 🤔 @graydon is there anyway to take advantage of that?

PR #21, opened 2022-05-04
Host guest env reorg

graydon:
This is a major reorg of the host crate -- so major it even renames the root crate to stellar-contract-env! We'll need to rename this repo if we go along with this.

Along the way, the following (with motives):

  - Switches from one crate with cfg-controlled bits to 4 crates in a dependency diamond. This is to make both configs testable without having to toggle the cfg; this is silly but a consequence of how feature flags work in rust, there's no way to have two turned on at once at the root so we need two separate crates (and another with the common parts). This is actually [the recommended approach to mutually exclusive features](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features). Bleh.
  - Do some renaming for fewer compound words. We need a name for the interface between host and guest as well as the implementations of the interface as seen from host and guest perspectives. I tried a lot of words and the simplest I cold find was `Host`, `Guest` and `Env` for the interface they both implement.
  - Make a handle-containing-and-environment type (actually two: `EnvVal` and `EnvObj`) so that operator overloading, `Ord` and `Eq` can work at least on `EnvVal` and `EnvObj` types. The latter are needed so that we can reach through an `Env` to resolve or compare objects deeply when needed; the former are needed because we store them in `HostMap`s and such and need a uniform type, and also because even in the guest we need some type to return from functions that return maybe-an-object like the i64 injection function.
  - Implement `Ord` and `Eq` on the `EnvVal` and `EnvObj` types, in terms of the presence of an `Env`, by reaching through the Env. This should let users in the guest deep-compare objects the same way the host does.
  - Implement the `Guest` type as a unit struct, defining `Object = EnvObj<Guest>` when `cfg(guest)`.
  - Implement the `Host` type as the former `HostContext` and `Object = EnvObj<Weak<Host>>` when `cfg(host)`
 
This very roughly seems to all hold together, but it's late and I didn't get it entirely .. ergonomic. In particular I expect @leighmcculloch will want to fiddle a lot with the conversion functions and maybe actually add a wrapper around `Rc<Host>` and implement `Env` for that too (or instead of `WeakHost`) for when users manually construct or pass-around a strongly-referenced Host. So there's stuff to do. But overall I _think_ this gets us a fair ways along in the direction we were discussing today.


stellar-contract-env/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "stellar-contract-env"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+stellar-contract-env-guest = { path = "../stellar-contract-env-guest", optional = true }
+stellar-contract-env-host = { path = "../stellar-contract-env-host", optional = true }
+
+[features]
+default = ["host"]
```
leighmcculloch, line 11: We should leave this default out I think, and force the importer to specify a feature. This is actually better for one-or-the-other features I think, because we don't have to disable the default feature set.

stellar-contract-env-common/src/env.rs
```diff
@@ -0,0 +1,57 @@
+use super::Val;
+use core::any;
+
+// This trait needs to be implemented by any type that plays the role of the
+// host for a contract. In a contract test or core setting this will be a full
+// HostContext, whereas in a contract's wasm build it will be an empty
+// type that calls through to global wasm imports provided at
+// wasm-module-instantiation time.
+pub trait Env: Sized + Clone {
```
leighmcculloch, line 9: This looks great.

leighmcculloch (approved):

leighmcculloch:
> We'll need to rename this repo if we go along with this.

Requesting a rename in https://github.com/stellar/terraform/pull/787.

PR #26, opened 2022-05-06
Allow EnvObj to be created by sdk

leighmcculloch:
### What

Allow EnvObj to be created by the sdk.

### Why

The SDK gets RawVal's back from the Env (Host/Guest), but needs to convert them into EnvObjs, but there's no way for it to do that because the type is privately an EnvVal.

### Known limitations

N/A


leighmcculloch:
@graydon I needed to make EnvObj constructible on the SDK side since the Env only provides back RawVal's. Initially I did commit 7d7275402677aa15ed6430988d9ca3c046a57035 which didn't change much and just added a conversion function, but it felt a bit awkward to be specifying the tag value here again. I think defining EnvObj as explicitly having a RawObj makes it a bit more natural to use? Happy to discard the second commit and keep just the first commit if you prefer.

PR #27, opened 2022-05-07
Add TryFrom<RawVal> to RawValType

leighmcculloch:
### What

Add TryFrom<RawVal> to RawValType.

### Why

So that anywhere RawValType is used as a generic parameter we can assume that the type specified for the parameter also has TryFrom<RawVal>, which all RawValType's do.

### Known limitations

N/A


graydon:
I feel like I was here before and then there was a problem and I had to remove it, but I don't remember what it was so .. let's try!

PR #33, opened 2022-05-09
Add impl to `obj_to_u64`, `obj_to_i64`

jayz22:
### What

Add a couple of host functions:
- `obj_to_u64`
- `obj_to_i64`

### Why
Provide implementation for essential host functions. 

### Known limitations
This is just a start, still many more to do.

### DIFF
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -168,7 +168,9 @@ impl<E: Env> TryFrom<EnvVal<E, RawVal>> for i64 {
         if ev.val.is_positive_i64() {
             Ok(unsafe { ev.val.unchecked_as_positive_i64() })
         } else if Object::val_is_obj_type(ev.val, ScObjectType::I64) {
-            Ok(ev.env.obj_to_i64(ev.val))
+            Ok(ev
+                .env
+                .obj_to_i64(unsafe { Object::unchecked_from_val(ev.val) }))
         } else {
             Err(())
         }
@@ -196,7 +198,9 @@ impl<E: Env> TryFrom<EnvVal<E, RawVal>> for u64 {
         if ev.val.is_positive_i64() {
             Ok(unsafe { ev.val.unchecked_as_positive_i64() } as u64)
         } else if Object::val_is_obj_type(ev.val, ScObjectType::U64) {
-            Ok(ev.env.obj_to_u64(ev.val))
+            Ok(ev
+                .env
+                .obj_to_u64(unsafe { Object::unchecked_from_val(ev.val) }))
         } else {
             Err(())
         }
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -2,7 +2,6 @@ use crate::{BitSet, Object, Status, Symbol, Tag, TagType, TaggedVal, Val};
 
 use super::{
     raw_val::{RawVal, RawValConvertible},
-    xdr::ScObjectType,
     Env,
 };
 use core::{cmp::Ordering, fmt::Debug};
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -167,12 +167,9 @@ impl<E: Env> TryFrom<EnvVal<E, RawVal>> for i64 {
     fn try_from(ev: EnvVal<E, RawVal>) -> Result<Self, Self::Error> {
         if ev.val.is_positive_i64() {
             Ok(unsafe { ev.val.unchecked_as_positive_i64() })
-        } else if Object::val_is_obj_type(ev.val, ScObjectType::I64) {
-            Ok(ev
-                .env
-                .obj_to_i64(unsafe { Object::unchecked_from_val(ev.val) }))
         } else {
-            Err(())
+            let ev: EnvVal<E, Object> = ev.try_into()?;
+            Ok(ev.env.obj_to_i64(ev.val))
         }
     }
 }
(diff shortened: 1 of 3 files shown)


leighmcculloch (commented):
👏🏻 Excited that we're starting to get host fns filled out, because we need them for the SDK tests to start being meaningful again. I left a few comments inline.

stellar-contract-env-host/src/host.rs
```diff
@@ -299,15 +299,31 @@ impl Env for Host {
     }
 
     fn obj_to_u64(&mut self, u: RawVal) -> u64 {
-        todo!()
+        unsafe {
+            self.unchecked_visit_val_obj(u, move |ho| {
+                if let Some(HostObject::U64(num)) = ho {
```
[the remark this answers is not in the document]
jayz22: I see your point. If I understood correct, the assumption host is making is the `u: RawVal` is a `RawObj`, by calling the unsafe method `unchecked_visit_val_obj`. If `u` is not `Tag::Object`, the result from this method will match to `None` which triggers the the `!panic`. So it feels to me ok, unless my understanding is missing. 
My first implementation (in the 1st commit which you also reviewed), goes from `RawVal -> RawObj -> ScObject->u64`, which doesn't involve the unsafe method but takes an extra hop. I then saw a couple of host functions already exist (`vec_len` and `vec_push`) that uses the `unchecked_visit_val_obj` so I decided to change my implementation. Do you think that was a safer approach?

jayz22:
Converting this to draft for now as I'll be adding more host functions and iterating on existing ones. Feedback is welcome at any time.

leighmcculloch:
@jayz22 I think it is fine if you want to micro-commit this stuff to main. I would find value in having access to those i64 fns earlier than other ones.

jayz22:
> @jayz22 I think it is fine if you want to micro-commit this stuff to main. I would find value in having access to those i64 fns earlier than other ones.

I agree, it's probably better to do micro-commits and to iterate on it more quickly. I added one more fn (`vec_get`) and um-marked the draft. Still waiting on @graydon 's input on the unsafe fn usage.

stellar-contract-env-host/src/host.rs
```diff
@@ -351,7 +367,25 @@ impl Env for Host {
     }
 
     fn vec_get(&mut self, v: RawVal, i: RawVal) -> RawVal {
-        todo!()
+        let index = if u32::is_val_type(i) {
+            unsafe { u32::unchecked_from_val(i) }
+        } else {
+            panic!("index has wrong type, must be u32")
+        };
```
leighmcculloch, line 374: I think if you're going to do this in a checked way (you're checking with is_val_type, you may as well use:
```suggestion
        let index: u32 = i.try_into().expect("index has wrong type, must be u32");
```

jayz22: That's a great suggestion, fixed!

stellar-contract-env-host/src/host.rs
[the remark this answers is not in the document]
graydon: I think a couple things here:

1. Host should probably have a systematic treatment for "fetch an object by ID with an expected host object type, and fail if it's the wrong type" so we don't have to repeat that pattern everywhere.
2. In general it should not be possible for guest code to get the host to panic -- this would make it possible for a smart contract to crash stellar-core -- so we need a general pattern where the host function implementation is actually a partial function returning Result<...> and the entrypoint from Env panics on the error case when running locally / linking directly to the SDK and traps the VM when being invoked through a call from the VM. This is a little subtle and may involve some changes to the trait and method signatures. I'll try to put something together.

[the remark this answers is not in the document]
leighmcculloch: > 2\. This is a little subtle and may involve some changes to the trait and method signatures. I'll try to put something together.

I just merged https://github.com/stellar/rs-stellar-contract-env/pull/39 that changed the signatures, so recommend updating your local if you just started doing this!

stellar-contract-env-common/src/env_val.rs
```diff
@@ -168,7 +168,9 @@ impl<E: Env> TryFrom<EnvVal<E, RawVal>> for i64 {
         if ev.val.is_positive_i64() {
             Ok(unsafe { ev.val.unchecked_as_positive_i64() })
         } else if Object::val_is_obj_type(ev.val, ScObjectType::I64) {
-            Ok(ev.env.obj_to_i64(ev.val))
+            Ok(ev
+                .env
+                .obj_to_i64(unsafe { Object::unchecked_from_val(ev.val) }))
         } else {
             Err(())
         }
```
leighmcculloch, line 176: This code doesn't need to do a check and then unchecked from val, because you can use the TryInto trait impl that exists, and then rely on the host checking I64 specifically. This keeps at least one branch out of the guest WASM code.

```rust
if ev.val.is_positive_i64() {
    Ok(unsafe { ev.val.unchecked_as_positive_i64() })
} else {
    let ev: EnvVal<E, Object> = ev.try_into()?;
    Ok(ev.env.obj_to_i64(ev.val))
}
```

jayz22: Thanks, updated.

leighmcculloch (approved):
🎉 Looks great. If you're ready to merge, hit the `Enable Auto Merge` button or simply `Squash & Merge` if that isn't showing.

jayz22:
> 🎉 Looks great. If you're ready to merge, hit the `Enable Auto Merge` button or simply `Squash & Merge` if that isn't showing. GitHub will squash all your commits as part of the merge, so no need to do that manually.

Thanks @leighmcculloch, I don't think I have the write access to merge:
"Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests."

PR #39, opened 2022-05-10
Make ret values RawObj that are guaranteed to be

leighmcculloch:
### What

Make return values `RawObj` that are guaranteed to be that type.

### Why

It simplifies the SDK since the SDK will be assuming the values are RawObj's anyway, and this allows it to do so without unchecked/unsafe blocks of code.

Note that it is important that host fn inputs are only `RawVal`s because the host should not make any assumptions about what the guest has provided as arguments.

### Known limitations

N/A

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -1,4 +1,4 @@
-use super::RawVal;
+use super::{RawObj, RawVal};
 use core::any;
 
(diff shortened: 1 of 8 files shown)


graydon (approved):
LGTM!

leighmcculloch:
The corresponding update to the SDK is this: https://github.com/stellar/rs-stellar-contract-sdk/pull/76

PR #41, opened 2022-05-10
Implement and use CheckedEnv, add safe & typed Host::visit_obj

graydon:
A few changes in here:

1. Upgrade a bunch of host function input types to RawObj. They're statically checked in the loopback use case and they're checked by the arg-conversion path in the dispatch functions, trapping on failure to convert (i.e. if the user provides a non-object val).
2. Adds real (if extremely rudimentary) error types to the host and common crates.
3. Adds a new `CheckedEnv` trait that is just like `Env` (same method names and base return types) but it returns `Result<T, Self::Error>` so that we can implement a fallible version of `Env` on the host and not panic.
4. Adds a blanket impl of `Env` for `CheckedEnv` that just unwraps its results, so that when you use `Host: CheckedEnv` in the loopback case the user gets the same promise as they get when running in the VM guest case: any error in the host _will not return_ to the guest. Panic in loopback, trap when running in VM.
5. Adds a new type-directed `HostObjectType::try_extract` function that tries to match a host object type's variant to the case the trait can extract Self from.
6. Adds a new helper function `Host::visit_obj` that combines all of the above, giving you a fallible method that takes a closure that, if you put a type signature of some host object type in the closure parameters, will attempt to extract the type in question and return an error on either host index out-of-bounds _or_ host object type mismatch, and all those errors will trap at the VM rather than panic.
7. Switches Host to implement CheckedEnv, and redoes the example host functions to illustrate.

I'm sorry this is a lot. As with many changes so far, there's a big pile of interrelated structural stuff to do that all has to happen at once, but it should make writing host functions much more straightforward.

Feel free to enhance the actual Error type as needed for clarity. I just made a kinda silly General category that carries a static string slice.


stellar-contract-env-common/src/env.rs
```diff
@@ -77,29 +77,29 @@ macro_rules! call_macro_with_all_host_functions {
 
             mod map "m" {
                 {"$_", fn map_new() -> RawObj }
-                {"$0", fn map_put(m:RawVal, k:RawVal, v:RawVal) -> RawObj}
-                {"$1", fn map_get(m:RawVal, k:RawVal) -> RawVal}
-                {"$2", fn map_del(m:RawVal, k:RawVal) -> RawObj}
-                {"$3", fn map_len(m:RawVal) -> RawVal}
-                {"$4", fn map_keys(m:RawVal) -> RawObj}
-                {"$5", fn map_has(m:RawVal,k:RawVal) -> RawVal}
+                {"$0", fn map_put(m:RawObj, k:RawVal, v:RawVal) -> RawObj}
```
leighmcculloch, line 80: 🎉

stellar-contract-env-host/tests/integration.rs
```diff
@@ -7,7 +7,7 @@ fn vec_as_seen_by_user() -> Result<(), ()> {
 
     let vec1a = host.vec_new().in_env(&host);
     let vec1b = host
-        .vec_push(vec1a.as_ref().val, int1.as_ref().val)
+        .vec_push(vec1a.clone().into(), int1.as_ref().val)
```
leighmcculloch, line 10: Why is clone preferred over as_ref on the host side?

leighmcculloch (commented):
👏🏻

PR #42, opened 2022-05-10
Shift generic param from EnvValType fns to type

leighmcculloch:
### What

Shift generic param from EnvValType fns to type.

### Why

It lends itself to some better ergonomics in the SDK when implementing EnvValType for SDK types across all Envs values.

### Known limitations

N/A


leighmcculloch:
Superseded by #45

PR #43, opened 2022-05-10
Add AsRef RawVal to EnvObj and UnimplementedEnv

leighmcculloch:
### What

Add accessors for `RawVal` and `RawObj` to `EnvObj` and an implementation of `Env`, `UnimplementedEnv`.

### Why

The `EnvObj` has a `RawVal` / `RawObj` within it, and providing access to it is convenient.

The `UnimplementedEnv` is helpful in tests when we need an `Env` but we don't need an `Env` that actually does anything, which is helpful for testing some lower level fns. In the common crate we don't have access to the `Host` `Env` that we normally use in higher level tests.

### Known limitations

N/A

### DIFF
--- a/stellar-contract-env-common/src/unimplemented_env.rs
+++ b/stellar-contract-env-common/src/unimplemented_env.rs
@@ -1,0 +1,86 @@
+use super::{call_macro_with_all_host_functions, Env, EnvBase, RawObj, RawVal};
+use core::any;
+use paste::paste;
+
+#[derive(Clone, Default)]
+pub struct UnimplementedEnv;
+
+impl EnvBase for UnimplementedEnv {
+    fn as_mut_any(&mut self) -> &mut dyn any::Any {
+        self
+    }
+
+    fn check_same_env(&self, _other: &Self) {}
+}
+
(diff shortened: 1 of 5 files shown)


stellar-contract-env-common/src/unimplemented_env.rs
```diff
@@ -0,0 +1,86 @@
...
+//
+//  {fn $fn_id:ident $args:tt -> $ret:ty}
+//
+// and produces the the corresponding method declaration to be used in the Env
+// trait.
+macro_rules! host_function_helper {
+    {fn $fn_id:ident($($arg:ident:$type:ty),*) -> $ret:ty}
+    =>
+    {
+        fn $fn_id(&self, $(paste!{[<_ $arg>]}:$type),*) -> $ret {
```
graydon, line 27: You can just use `_:$type` here and avoid the paste! dep

(if you really want to keep paste, I think it can move to a build-dep not a dep)

leighmcculloch: Ah nice. Good idea. I don't want paste. I was only using it because concat_ident macro is unstable and I couldn't use it.

PR #45, opened 2022-05-11
Reorganize RawVal wrappers around TaggedVal, remove duplication

graydon:
This addresses the "duplicate wrapper hierarchy" we were developing between RawObj / EnvObj and RawVal / EnvVal. It adds a little more machinery to substitute (the diff is, remarkably, exactly 550 lines removed and 550 added) but I think it's cognitively much less problematic:

1.  Adds `trait TagType`, a bunch of concrete unit types that impl it, and `TaggedVal<T:TagType>(RawVal, PhantomData)` as a generic "wrapper for a RawVal that we know the tag of"
2. Renames all the existing wrappers, such as `type Symbol = TaggedVal<TagSymbol>`
3. Does this to `RawObj` in particular, renaming it back to `type Object = TaggedVal<TagObject>`
4. Moves the previous inherent methods on the old newtypes to inherent methods on these `TaggedVal` specializations. Turns out this is one place you are allowed to do specialization in Rust. Who knew?
5. Adds a new `Val` trait that can `AsRef` and `AsMut` through to a `RawVal` and impls this for both `RawVal` and `TaggedVal`.
6. Makes `EnvVal` parametric in not just the `E:Env` but also `V:Val`.
7. Removes `EnvObj` entirely, since it's just `EnvVal<E, Object>` a.k.a. `EnvVal<E, TaggedVal<TagObject>>`
8. Mops up all the fallout in tests and host methods and such

Overall I _think_ this is an improvement. Error messages get longer -- not so good -- with all the type aliases that don't get cut off at newtypes; but besides that I think the structure is simpler and it's more obvious where to put a thing and how to do conversions. Plus we can just use as_ref() in a lot more places, without cloning things that might contain refcounted hosts.

Idk, it could be "not enough improved to warrant", but I think it's a step in the right direction. @leighmcculloch wdyt?


stellar-contract-env-common/src/env_val.rs
```diff
@@ -1,159 +1,220 @@
...
 
-impl<E: Env> From<EnvVal<E>> for RawVal {
-    #[inline(always)]
-    fn from(ev: EnvVal<E>) -> Self {
-        ev.val
+impl<E: Env, T: TagType> EnvVal<E, TaggedVal<T>> {
+    // This just lets callers disambiguate by name when they call as_ref()
+    // in a context that might want &TaggedVal<T> or &RawRef; it saves
+    // them writing <_ as AsRef<RawVal>>::as_ref(foo)
+    pub fn as_raw_ref(&self) -> &RawVal {
```
leighmcculloch, line 29: Given that there is no cost difference between a ref to a val, and a val, we should probably just forego the ref?
```suggestion
    pub fn as_val(&self) -> RawVal {
```
Wdyt?

stellar-contract-env-common/src/env_val.rs
```diff
@@ -1,159 +1,220 @@
...
+    fn as_ref(&self) -> &TaggedVal<T> {
+        &self.val
+    }
+}
+
+impl<E: Env, T: TagType> AsMut<TaggedVal<T>> for EnvVal<E, TaggedVal<T>> {
+    fn as_mut(&mut self) -> &mut TaggedVal<T> {
+        &mut self.val
+    }
+}
```
leighmcculloch, line 56: This is great 🎉

stellar-contract-env-common/src/tagged_val.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+impl TagType for TagSymbol {
+    const TAG: Tag = Tag::Symbol;
+}
+impl TagType for TagBitSet {
+    const TAG: Tag = Tag::BitSet;
+}
+impl TagType for TagStatus {
+    const TAG: Tag = Tag::Status;
+}
```
leighmcculloch, line 50: 🤯

leighmcculloch (approved):

PR #47, opened 2022-05-11
Remove EnvObj

leighmcculloch:
### What

Remove EnvObj.

### Why

It was replaced by using `EnvVal<TaggedVal<TagObject>>` in #45 and is no longer needed.

### Known limitations

N/A


graydon:
Oops sorry missed it!

PR #54, opened 2022-05-11
Split EnvValConvertible into Into<EnvVal>, IntoEnvVal, TryFrom<EnvVal>

leighmcculloch:
### What

Make `EnvValConvertible` blanket impl:

- `IntoEnvVal`, which is blanket impl for all `Into<EnvVal>`
- `TryFrom<EnvVal>`

And add explicit `IntoEnvVal` impls for all `RawVal`s and `TaggedVal`s.

### Why

I noticed that in the SDK I was implementing both TryFrom and EnvValConvertible with the same logic for types. We should just ask types to implement each fn once, and EnvValConvertible can rely on the presence of the TryFrom when going from EnvVal to an impl of EnvValConvertible since the entire value, including the Env, is held in a single value. All types should be able to provide a TryFrom in this direction.

I also noticed that most types in the SDK are going to be able to impl the reverse by simply implementing `Into<EnvVal>`.

Overall this simplifies the SDK because we can use a blanket impl to provide the `EnvValConvertible` for types implementing the `Into` and `TryFrom`.

There is a small increased code for `RawVal` and `TaggedVal` as each have to manually impl `IntoEnvVal`, and `TryFrom`, but that is not a big deal considering they are both limited to a small number of types that will not change or grow. Unlikely the SDK that is likely to have many new types.

### Known limitations

We used Option<> on the EnvValConvertible type, or RawValConvertible type, because it saved like 1 byte of code size. By introducing From's into the path I think we omit that saving, but we are using the From trait in so many places I think we're probably splitting hairs worrying about this level of optimization and it'll be lost in the noise of other things anyway.

### DIFF
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -96,12 +96,11 @@ pub trait EnvValConvertible<E: Env, V: Val>:
     fn into_val(self, env: &E) -> V {
         Self::into_env_val(self, env).val
     }
-    fn try_from_val(env: &E, v: V) -> Option<Self> {
+    fn try_from_val(env: &E, v: V) -> Result<Self, Self::Error> {
         Self::try_from(EnvVal {
             env: env.clone(),
             val: v,
         })
-        .ok()
     }
 }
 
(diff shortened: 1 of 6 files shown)


leighmcculloch:
This change results in SDK types implementing three fns:

Screen Shot 2022-05-11 at 10 03 42 PM

graydon:
I'm not going to pretend I have insight into the optimal path in this corner of the design space but it looks fine to me, happy if it makes your life easier!

PR #55, opened 2022-05-13
Better public vm invoke

graydon:
This just tidies up the interface for core.

### DIFF
--- a/stellar-contract-env-host/src/vm.rs
+++ b/stellar-contract-env-host/src/vm.rs
@@ -1,7 +1,13 @@
 mod dispatch;
 mod func_info;
 
-use super::{host, Host, RawVal};
+use std::error::Error;
+
+use super::{
+    host,
+    xdr::{ScVal, ScVec},
+    Host, RawVal,
+};
 use func_info::HOST_FUNCTIONS;
 use wasmi::{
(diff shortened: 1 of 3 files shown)


stellar-contract-env-host/src/vm.rs
```diff
@@ -142,4 +148,21 @@ impl VM {
...
+    /// XDR ScVal arguments into Host-specific RawVals and converting the
+    /// RawVal result back to an ScVal.
+    pub fn invoke_function(
+        &self,
+        host: &mut Host,
+        func: &str,
+        args: &ScVec,
+    ) -> Result<ScVal, Box<dyn Error>> {
+        let mut raw_args: Vec<RawVal> = Vec::new();
+        for scv in args.0.iter() {
```
leighmcculloch, line 162: This looks annoying that you need to always be `.0`ing on types that wrap Vec's in the XDR. I'll surface impls of Vec's API up into all types that wrap a Vec to make it better.

leighmcculloch: Created issue for this: https://github.com/stellar/rs-stellar-xdr/issues/45

graydon: Doesn't matter to me at all, but I won't stop you if it bugs you!

PR #59, opened 2022-05-14
Change types of some BigInt fns

leighmcculloch:
### What

Change types of some BigInt fns from Object to RawVal or builtins.

### Why

The values are not objects and are either i32's encoded in RawVals, or they are i64/u64 that can be transmitted over the boundary as they are.

### Known limitations

N/A

cc @graydon @jayz22

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -127,7 +127,7 @@ macro_rules! call_macro_with_all_host_functions {
             }
 
             mod bigint "b" {
-                {"$_", fn bigint_from_u64(x:Object) -> Object}
+                {"$_", fn bigint_from_u64(x:u64) -> Object}
                 {"$0", fn bigint_add(x:Object,y:Object) -> Object}
                 {"$1", fn bigint_sub(x:Object,y:Object) -> Object}
                 {"$2", fn bigint_mul(x:Object,y:Object) -> Object}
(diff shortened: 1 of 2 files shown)


stellar-contract-env-common/src/env.rs
```diff
@@ -138,16 +138,16 @@ macro_rules! call_macro_with_all_host_functions {
                 {"$7", fn bigint_xor(x:Object,y:Object) -> Object}
                 {"$8", fn bigint_shl(x:Object,y:RawVal) -> Object}
                 {"$9", fn bigint_shr(x:Object,y:RawVal) -> Object}
-                {"$A", fn bigint_cmp(x:Object,y:Object) -> Object}
-                {"$B", fn bigint_is_zero(x:Object) -> Object}
+                {"$A", fn bigint_cmp(x:Object,y:Object) -> RawVal}
```
leighmcculloch, line 141: Should be a -1, 0, 1, I assume a i32.

stellar-contract-env-common/src/env.rs
```diff
@@ -138,16 +138,16 @@ macro_rules! call_macro_with_all_host_functions {
...
+                {"$B", fn bigint_is_zero(x:Object) -> RawVal}
```
leighmcculloch, line 142: Should be a bool.

stellar-contract-env-common/src/env.rs
```diff
@@ -138,16 +138,16 @@ macro_rules! call_macro_with_all_host_functions {
...
                 {"$C", fn bigint_neg(x:Object) -> Object}
                 {"$D", fn bigint_not(x:Object) -> Object}
                 {"$E", fn bigint_gcd(x:Object) -> Object}
                 {"$F", fn bigint_lcm(x:Object,y:Object) -> Object}
                 {"$G", fn bigint_pow(x:Object,y:Object) -> Object}
                 {"$H", fn bigint_pow_mod(p:Object,q:Object,m:Object) -> Object}
                 {"$I", fn bigint_sqrt(x:Object) -> Object}
-                {"$J", fn bigint_bits(x:Object) -> Object}
+                {"$J", fn bigint_bits(x:Object) -> RawVal}
```
leighmcculloch, line 150: Should be an u32 I guess.

Issue #75, opened 2022-05-20
Add way to deep clone an Env


leighmcculloch:
As described in https://github.com/stellar/rs-stellar-contract-sdk/issues/52#issuecomment-1132194212 there are some testing scenarios where taking a snapshot of the host `Env` and copy it and replay from a point is helpful, and not having this can make some testing less ergonomic/easy.

Probably something like a `deep_clone` fn like how @graydon describes in https://github.com/stellar/rs-stellar-contract-sdk/issues/52#issuecomment-1132343405.

PR #77, opened 2022-05-20
Add EnvBase::deep_clone, implement on Host.

graydon:
Address issue #75

### DIFF
--- a/stellar-contract-env-common/src/unimplemented_env.rs
+++ b/stellar-contract-env-common/src/unimplemented_env.rs
@@ -10,6 +10,10 @@ impl EnvBase for UnimplementedEnv {
     }
 
     fn check_same_env(&self, _other: &Self) {}
+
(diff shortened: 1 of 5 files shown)


stellar-contract-env-common/src/unimplemented_env.rs
```diff
@@ -10,6 +10,10 @@ impl EnvBase for UnimplementedEnv {
     }
 
     fn check_same_env(&self, _other: &Self) {}
+
+    fn deep_clone(&self) -> Self {
+        Self
```
leighmcculloch, line 15: Given that it would be a logical error to call this on a Guest, is there a way for us to make it a compile error to that ever to happen?

leighmcculloch: This comment was intended to be on the Guest impl.

leighmcculloch (approved):

PR #80, opened 2022-05-24
Add more hostfns; use `rstest` for tests

jayz22:
### What
Add more host functions:
- Vec pop, front, back
- Corresponding test cases.

Use crate `rstest`, use fixture in our tests.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-host/Cargo.toml
+++ b/stellar-contract-env-host/Cargo.toml
@@ -14,8 +14,5 @@ num-rational = "0.4"
 thiserror = "1.0.31"
 wasmi = { version = "0.11.0", optional = true }
 
-[dev-dependencies]
-rstest = "0.13.0"
-
 [features]
 vm = ["wasmi", "stellar-contract-env-common/vm"]
--- a/stellar-contract-env-host/Cargo.toml
+++ b/stellar-contract-env-host/Cargo.toml
@@ -13,6 +13,7 @@ num-bigint = "0.4"
 num-rational = "0.4"
 thiserror = "1.0.31"
 wasmi = { version = "0.11.0", optional = true }
+rstest = "0.13.0"
 
 [features]
 vm = ["wasmi", "stellar-contract-env-common/vm"]
--- a/stellar-contract-env-host/Cargo.toml
+++ b/stellar-contract-env-host/Cargo.toml
@@ -13,6 +13,8 @@ num-bigint = "0.4"
 num-rational = "0.4"
 thiserror = "1.0.31"
 wasmi = { version = "0.11.0", optional = true }
+
+[dev-dependencies]
 rstest = "0.13.0"
 
 [features]
--- a/stellar-contract-env-host/src/test.rs
+++ b/stellar-contract-env-host/src/test.rs
@@ -1,15 +1,20 @@
-use stellar_contract_env_common::RawValConvertible;
+use rstest::{fixture, rstest};
+use stellar_contract_env_common::{CheckedEnv, RawValConvertible};
 
 use crate::{
     xdr::{ScObject, ScObjectType, ScVal, ScVec},
     Host, IntoEnvVal, Object, Tag,
 };
 
-#[test]
-fn u64_roundtrip() {
-    let host = Host::default();
+#[fixture]
+fn default_host() -> Host {
+    Host::default()
+}
+
+#[rstest]
+fn u64_roundtrip(default_host: Host) {
     let u: u64 = 38473_u64; // This will be treated as a ScVal::Object::U64
-    let v = u.into_env_val(&host);
+    let v = u.into_env_val(&default_host);
     let obj: Object = v.val.try_into().unwrap();
     assert!(obj.is_obj_type(ScObjectType::U64));
     assert_eq!(obj.get_handle(), 0);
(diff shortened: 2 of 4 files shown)


stellar-contract-env-host/src/test.rs
```diff
@@ -89,3 +89,31 @@ fn vec_as_seen_by_host() -> Result<(), ()> {
     assert_eq!(val0, val1);
     Ok(())
 }
+
+#[test]
+fn vec_front_and_back() -> Result<(), ()> {
+    let mut host = Host::default();
+    let scvec0: ScVec = ScVec(vec![ScVal::U32(1), ScVal::U32(2), ScVal::U32(3)].try_into()?);
```
leighmcculloch, line 96: You shouldn't need the ScVec:
```suggestion
    let scvec0: ScVec = vec![ScVal::U32(1), ScVal::U32(2), ScVal::U32(3)].try_into()?;
```
Let me know if that doesn't work because the direct conversation was added last week.

stellar-contract-env-host/src/test.rs
```diff
@@ -89,3 +89,31 @@ fn vec_as_seen_by_host() -> Result<(), ()> {
...
+
+#[test]
+#[should_panic(expected = "value does not exist")]
+fn empty_vec_front_and_back() {
+    let mut host = Host::default();
+    let scvec0: ScVec = ScVec(vec![].try_into().unwrap());
+    let scobj0 = ScObject::Vec(scvec0);
+    let obj0 = host.to_host_obj(&scobj0).unwrap();
+    host.vec_front(*obj0.as_ref()).unwrap();
+    host.vec_back(*obj0.as_ref()).unwrap();
```
leighmcculloch, line 118: Only the front call is being tested in this test because the test is failing immediately. You'll either need to make two tests, or use catch_unwind to catch the exception and continue.

stellar-contract-env-host/src/test.rs
```diff
@@ -89,3 +89,31 @@ fn vec_as_seen_by_host() -> Result<(), ()> {
...
+    let scvec0: ScVec = ScVec(vec![ScVal::U32(1), ScVal::U32(2), ScVal::U32(3)].try_into()?);
+    let scobj0 = ScObject::Vec(scvec0);
+    let obj0 = host.to_host_obj(&scobj0).unwrap();
+    let front = unsafe {
+        <i32 as RawValConvertible>::unchecked_from_val(host.vec_front(*obj0.as_ref()).unwrap())
+    };
+    let back = unsafe {
+        <i32 as RawValConvertible>::unchecked_from_val(host.vec_back(*obj0.as_ref()).unwrap())
+    };
+    assert_eq!(front, 1);
```
leighmcculloch, line 105: Instead of doing the unsafe unchecked you could compare the value returned with `ScVal::I32(1)`, etc. It might help keep the test focused on what it needs to test.

leighmcculloch (commented):
Woohoo. Great to see some more of these functions being implemented. A few comments inline.

stellar-contract-env-host/src/test.rs
jayz22: Good point! 
`catch_unwind` doesn't work complaining some types not being unwind safe. Even if it does, in order to compare error message (not just checking if `is_err()`), we would need to add `PartialEq` trait in a few places including in the XDR.
Anyway I think the better approach is to test each case separately. See update.
I've added `rstest` and used its `fixture` to avoid code repetition.

stellar-contract-env-host/src/test.rs
jayz22: Hmm not sure about this one. 
I think here comparing the raw value is what we want. `ScVal::I32` is what we used to construct the host object, the value gets unpacked from `ScVal`. Retrieving elements from the vector object gets back the `RawVal`, which doesn't know anything about `ScVal` any more.

stellar-contract-env-host/Cargo.toml
```diff
@@ -13,6 +13,7 @@ num-bigint = "0.4"
 num-rational = "0.4"
 thiserror = "1.0.31"
 wasmi = { version = "0.11.0", optional = true }
+rstest = "0.13.0"
```
leighmcculloch, line 16: Can you add this to the `dev-dependencies`? It should not be required outside tests.

stellar-contract-env-host/src/test.rs
leighmcculloch: Ah yes. 👍🏻

stellar-contract-env-host/src/test.rs
```diff
@@ -89,3 +97,49 @@ fn vec_as_seen_by_host() -> Result<(), ()> {
...
+    ScObject::Vec(scvec)
+}
+
+#[fixture]
+fn vec_obj() -> ScObject {
+    let scvec: ScVec = vec![ScVal::U32(1), ScVal::U32(2), ScVal::U32(3)]
+        .try_into()
+        .unwrap();
+    ScObject::Vec(scvec)
+}
```
leighmcculloch, line 113: Fwiw I think the tests were easier to read before fixtures were added. Tests that are largely self contained are easier to read and maintain. DRY doesn't have the same benefits in tests.

stellar-contract-env-host/Cargo.toml
jayz22: Yes, done.

stellar-contract-env-host/src/test.rs
jayz22: Hmm to me it makes writing tests easier with fewer lines of code. Since we will have many of these data structure tests operating on the same initial setup, I thought it might be worth.  But I definitely see your point. If you feel strongly about it I'm happy to revert it back.

leighmcculloch: I would opt for test simplicity and test isolation at this point.

jayz22: Reverted it back.

leighmcculloch (approved):

jayz22:
@leighmcculloch I don't have permission to merge.

PR #81, opened 2022-05-24
Move ScVal<>RawVal conversion to common

leighmcculloch:
### What

Move ScVal<>RawVal conversion out of the host and into common.

### Why

The host does not need to own conversion between these types as conversion between these types is mostly independent of the host. The host does need to be able to map objects into object handles though and the Host can provide an implementation of the trait for doing that.

While I was working on the CLI I need to be able to convert between these values, and so I at least need to make public the functions for doing the conversions. Moving them out into common felt natural.

Close https://github.com/stellar/rs-stellar-contract-env/issues/106

### Known limitations

N/A


leighmcculloch:
Ugh, I got pretty lost doing this and turns out I don't need it. But maybe it is still useful for some other reason to get the RawVal <> ScVal conversion outside the Host? Thoughts @graydon?

graydon:
I _think_ I'm ok with this .. need to study in a little more detail tomorrow, too wiped out tonight.

stellar-contract-env-common/src/env_val.rs
```diff
@@ -231,6 +252,98 @@ impl<E: Env> IntoEnvVal<E, RawVal> for u64 {
...
+                    if tag_static.is_type(ScStatic::True) {
+                        Ok(ScVal::Static(ScStatic::True))
+                    } else if tag_static.is_type(ScStatic::False) {
+                        Ok(ScVal::Static(ScStatic::False))
+                    } else if tag_static.is_type(ScStatic::Void) {
+                        Ok(ScVal::Static(ScStatic::Void))
+                    } else if tag_static.is_type(ScStatic::LedgerKeyContractCodeWasm) {
+                        Ok(ScVal::Static(ScStatic::LedgerKeyContractCodeWasm))
+                    } else {
+                        Err(ConversionError)
```
leighmcculloch, line 284: Unfortunately because these conversion can be used outside the host they can no longer return nice meaningful host errors. I don't see a way around this, so I'm going to press ahead with this, and maybe we can improve it somehow. Lmk if you see a way.

cc @graydon

graydon (approved):
LGTM

stellar-contract-env-common/src/env_val.rs
graydon: I put a lot of machinery for richer errors in EnvBase over in #243, maybe it's usable here?

PR #82, opened 2022-05-25
Add remaining vec hostfns and tests

jayz22:
### What

Add remaining `vec` hostfns:
- put
- get
- del
- take
- insert
- append

Fixed:
- pop

The only one omitted is `vec_drop` which is redundant with `vec_take`.

Add their corresponding test cases.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-host/src/test.rs
+++ b/stellar-contract-env-host/src/test.rs
@@ -2,7 +2,7 @@ use stellar_contract_env_common::{CheckedEnv, RawValConvertible};
 
 use crate::{
     xdr::{ScObject, ScObjectType, ScVal, ScVec},
-    Host, IntoEnvVal, Object, Tag,
+    Host, IntoEnvVal, Object, RawVal, Tag,
 };
 
 #[test]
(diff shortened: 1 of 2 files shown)


stellar-contract-env-host/src/test.rs
```diff
@@ -129,3 +129,226 @@ fn empty_vec_back() {
...
+#[test]
+fn vec_put_and_get() {
+    let mut host = Host::default();
+    let scvec: ScVec = vec![ScVal::U32(1), ScVal::U32(2), ScVal::U32(3)]
+        .try_into()
+        .unwrap();
+    let scobj = ScObject::Vec(scvec);
+    let obj = host.to_host_obj(&scobj).unwrap();
+    let i: RawVal = 1_u32.into();
+    let obj1 = host.vec_put(*obj.as_ref(), i, 9_u32.into()).unwrap();
```
leighmcculloch, line 142: Since you're not really wanting a ref to obj, using as_ref then copying is a bit of a roundabout way to do the conversion. We have `to_tagged()` that will convert the obj into a `TaggedVal<TagObject>`, but I can definitely see it is super not obvious that `to_tagged()` would get you what you want since we use the alias `Object` everywhere instead `TaggedVal<TagObject>`. So I just added `to_object()` in https://github.com/stellar/rs-stellar-contract-env/pull/84 that does the same thing with a better name, and I think it will fit your intent clearer here:
```suggestion
    let obj1 = host.vec_put(obj.to_object(), i, 9_u32.into()).unwrap();
```

leighmcculloch (approved):

stellar-contract-env-host/src/test.rs
jayz22: Ah, I missed the `to_tagged()`, you are right. Thanks for adding the `to_object()`.

leighmcculloch (approved):

PR #87, opened 2022-05-28
Conversions for (T0,T1,...)<>Rawval

leighmcculloch:
### What

Add conversions for tuples to and from RawVal.

### Why

To make it convenient to pass tuples across the host/guest boundary, without having to manually craft vecs.

Close stellar/rs-stellar-contract-sdk#107

cc @jonjove 

### Known limitations

N/A


leighmcculloch:
I want to add a test for these tuples, but it isn't trivial to do so without a mocked implementation of the `Env`. We should figure that out soon.

PR #88, opened 2022-05-28
Experiment with Into/TryFrom EnvVal impls that target the tagged type

leighmcculloch:
### What

Have types implement Into/TryFrom EnvVal<_, TaggedVal<T>> where T is the tag object that is 1-1 with the type.

### Why

For all types we've been implementing conversions to and from EnvVal<_, RawVal>, but for many types we could implement to and from the specific tag object tagged val, and then macro an implementation from there to RawVal.

This potentially allows types and code to not need to dance between Object and RawVal so much.

This also potentially makes it harder to use into_env_val and try_from_env_val and their related traits, because there will now be multiple Val impls for each.

### Known limitations

This is just an experiment. This might cause a code size increase, but unclear since it should be mostly pass through.

### DIFF
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -1,6 +1,6 @@
 use stellar_xdr::ScObjectType;
 
-use crate::{BitSet, Object, Status, Symbol, Tag, TagType, TaggedVal, Val};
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Closing as really unclear on benefits this would bring, and it introduces type uncertainty whenever using into_env_val, which isn't great. Might reopen later.

PR #92, opened 2022-06-02
Add cross-contract invoking hostfns, contract data hostfns, tests

jayz22:
### What

Add host functions for cross-contract invocation:
- `call0` through `call4`, sharing an implementation in private fn `call_n`
- Add remaining contract data hostfns: {`put`, `has`, `del`}`_contract_data`.
- Add a few tests. 

### Why

### Known limitations

Need to write more tests.


stellar-contract-env-host/src/host.rs
```diff
@@ -701,9 +777,30 @@ impl CheckedEnv for Host {
...
+            func,
+            &[
+                self.from_host_val(a)?,
+                self.from_host_val(b)?,
+                self.from_host_val(c)?,
+            ],
+        )
+    }
+
+    #[cfg(not(feature = "vm"))]
```
leighmcculloch, line 800: FYI so you're not defining the function twice you can move the cfg inside the function for the contents.

stellar-contract-env-host/src/storage.rs
```diff
@@ -10,7 +10,7 @@ pub struct Key {
     pub key: ScVal,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
leighmcculloch, line 13: I don't think there's a reason we can't add Eq either?
```suggestion
#[derive(Clone, Debug, Eq, PartialEq)]
```

stellar-contract-env-host/src/host.rs
jayz22: Thanks, did not know that. Fixed.

stellar-contract-env-host/src/storage.rs
jayz22: Fixed.

stellar-contract-env-host/src/host.rs
```diff
@@ -393,6 +397,42 @@ impl Host {
...
+            ScVal::Object(Some(ob)) => *ob,
+            _ => return Err(HostError::General("not an object")),
+        };
+        let code = match scobj {
+            ScObject::Binary(b) => b,
+            _ => return Err(HostError::General("not a binary object")),
+        };
+        let vm = Vm::new(&self, id, code.as_slice())?;
+        // Resolve the function symbol and invoke contract call
+        let sym = unsafe { <Symbol as RawValConvertible>::unchecked_from_val(func) };
```
graydon, line 428: You can actually set the argument type for `func` to be `Symbol`, and then it will automatically cause a conversion error if a user calls one of the `call_{1,2,3,4}` functions with a non-symbol value, and you can get rid of this `unchecked_from_val`.

stellar-contract-env-host/src/host.rs
```diff
@@ -393,6 +397,42 @@ impl Host {
...
+        let res = vm.invoke_function(
+            RefCell::new(self.clone()),
```
graydon, line 430: Sorry I forgot to remove this use of RefCell also.

stellar-contract-env-host/src/test.rs
```diff
@@ -352,3 +361,101 @@ fn vec_append_empty() {
...
+    let sym = Symbol::from_str("add");
+    let scobj = ScObject::Binary([0; 32].try_into()?);
+    let obj = host.to_host_obj(&scobj).unwrap();
+    let res = host
+        .call2(obj.to_object(), sym.into(), 1i32.into(), 2i32.into())
+        .unwrap();
+
+    assert!(res.is::<i32>());
+    assert!(res.get_tag() == Tag::I32);
+    let i = unsafe { <i32 as RawValConvertible>::unchecked_from_val(res) };
```
graydon, line 458: I think you can do `let i : i32 = res.try_into().unwrap()`

graydon (changes requested):
Couple minor changes left, but looking good!

stellar-contract-env-host/src/test.rs
jayz22: Thanks, fixed.

stellar-contract-env-host/src/host.rs
jayz22: No problem, have to replace the function call anyways :)

stellar-contract-env-host/src/host.rs
jayz22: That's great. Fixed!

graydon:
Great, thanks!

PR #108, opened 2022-06-10
Add hostfns signatures promised in CAP-51

jayz22:
### What

Add host functions signatures promised in [CAP-51](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0051.md). No impl.  

### Why

### Known limitations

Implementation will be done later.

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -86,6 +86,8 @@ macro_rules! call_macro_with_all_host_functions {
                 {"$3", fn map_len(m:Object) -> RawVal}
                 {"$4", fn map_keys(m:Object) -> Object}
                 {"$5", fn map_has(m:Object,k:RawVal) -> RawVal}
+                {"$6", fn map_lower_bound(m:Object,k:RawVal) -> RawVal}
+                {"$7", fn map_upper_bound(m:Object,k:RawVal) -> RawVal}
             }
 
             mod vec "v" {
@@ -103,6 +105,7 @@ macro_rules! call_macro_with_all_host_functions {
                 {"$9", fn vec_back(v:Object) -> RawVal}
                 {"$A", fn vec_insert(v:Object, i:RawVal, x:RawVal) -> Object}
                 {"$B", fn vec_append(v1:Object, v2:Object) -> Object}
+                {"$C", fn vec_slice(v:Object, i:RawVal) -> Object}
             }
 
             mod ledger "l" {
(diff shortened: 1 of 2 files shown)


stellar-contract-env-common/src/env.rs
```diff
@@ -103,6 +105,7 @@ macro_rules! call_macro_with_all_host_functions {
                 {"$9", fn vec_back(v:Object) -> RawVal}
                 {"$A", fn vec_insert(v:Object, i:RawVal, x:RawVal) -> Object}
                 {"$B", fn vec_append(v1:Object, v2:Object) -> Object}
+                {"$C", fn vec_slice(v:Object, i:RawVal) -> Object}
```
leighmcculloch, line 108: ❗ Slice should have three parameters, it should have an index and a length, but it looks like it only has an index here.

stellar-contract-env-common/src/env.rs
```diff
@@ -86,6 +86,8 @@ macro_rules! call_macro_with_all_host_functions {
                 {"$3", fn map_len(m:Object) -> RawVal}
                 {"$4", fn map_keys(m:Object) -> Object}
                 {"$5", fn map_has(m:Object,k:RawVal) -> RawVal}
+                {"$6", fn map_lower_bound(m:Object,k:RawVal) -> RawVal}
```
leighmcculloch, line 89: ❓ Seeing the function definition here prompts some questions about the map type and CAP-15:
- How will an ScoMap with multiple entries of the same key be treated?
   - Will core consider such a map invalid and fail any transaction that attempts to input such a map? 
- In what order will the map be iterated by these functions, will it be:
   - Undefined or bucket order and essentially unpredictable to the user, although consistent if repeated for a map not altered?
   - Random order so as to make the unpredictable nature of the order clearer and precent reliance on happenstance ordering? (e.g. like Go)
   - Ordered by the keys based on the keys ordering?
      - If it is ordered by the keys how will keys of different types be ordered?
   - Ordered based on insertion, which is convenient in many use cases? (e.g. Ruby, JavaScript)

cc @graydon

stellar-contract-env-common/src/env.rs
```diff
@@ -151,6 +154,29 @@ macro_rules! call_macro_with_all_host_functions {
                 {"$L", fn bigint_to_i64(x:Object) -> i64}
                 {"$M", fn bigint_from_i64(x:i64) -> Object}
             }
+
+            mod binary "x" {
+                {"$_", fn serialize_to_binary(x:Object) -> Object}
```
leighmcculloch, line 159: ❓ Do the serialization functions serialize to an ScVal? Or to the raw type?

For example, does an i32 serialize as an XDR integer, or as an XDR integer within an ScVal union with the I32 discriminant? I assume it's the latter, but unclear.

stellar-contract-env-common/src/env.rs
```diff
@@ -151,6 +154,29 @@ macro_rules! call_macro_with_all_host_functions {
...
+                {"$0", fn deserialize_from_binary(x:Object) -> Object}
+                {"$1", fn binary_new() -> Object}
+                {"$2", fn binary_get(x:Object,i:RawVal) -> Object}
+                {"$3", fn binary_pop(x:Object) -> Object}
+                {"$4", fn binary_push(x:Object,v:RawVal) -> Object}
+                {"$5", fn binary_insert(x:Object,i:RawVal,v:RawVal) -> Object}
+                {"$6", fn binary_remove(x:Object,i:RawVal) -> Object}
+                {"$7", fn binary_len(x:Object) -> RawVal}
```
leighmcculloch, line 167: ❓ Could we add functions for binary to parallel vec? For example: slice, append, put, etc.

stellar-contract-env-common/src/env.rs
```diff
@@ -151,6 +154,29 @@ macro_rules! call_macro_with_all_host_functions {
...
+                {"$8", fn binary_copy_to_guest_mem(x:Object,i:RawVal,j:RawVal,l:RawVal) -> RawVal}
+                {"$9", fn binary_copy_from_guest_mem(x:Object,i:RawVal,j:RawVal,l:RawVal) -> RawVal}
+                {"$A", fn binary_to_hash(x:Object) -> Object}
```
leighmcculloch, line 170: ❓ This is probably more a question for the cap, but could we invert the binary conversion functions so they live with their type? For example, make this `hash_from_binary`, and `hash_to_binary`. Since hash is the specialization and has its own function I think the specialization function for conversion should live with the other specialization functions for that type. Same for public key.

leighmcculloch (commented):
One comment for a fn that needs it's signature fixed (❗), otherwise looks good to merge. 

I also left some questions (❓) but none are intended to block merging this PR.

stellar-contract-env-common/src/env.rs
jayz22: Ah, good catch, fixed! (I somehow had "vec_split" in mind when I wrote this)

stellar-contract-env-common/src/env.rs
jayz22: Actually, both the serialize and the deserialize functions are intended to operate on `SCObject`s (and the function argument type `Object` reflects that). My rationale is all other non-object `SCVal` types can be trivially converted between bytes and an i64. Do you feel we should expose this to take in an `SCVal`?

stellar-contract-env-common/src/env.rs
jayz22: Yes.

stellar-contract-env-common/src/env.rs
leighmcculloch: Yup

stellar-contract-env-common/src/env.rs
jayz22: This is updated. Now the binary functions mirror vec ones.

stellar-contract-env-common/src/env.rs
jayz22: Fixed.

jayz22:
@leighmcculloch I think this can be merged? In case someone may need to work off of it.

leighmcculloch (approved):

PR #114, opened 2022-06-15
Support parsing and propagating doc-comments on env macros

### LINKED ISSUE #51: Env trait and impls get fn comments describing fns
We should add comments to the Env trait fns as we implement them because some of their behaviors are ambiguous knowing only the fn name.

graydon:
This lets you write doc-comments on the module and function inputs to the env macros, which are then copied through to some of the resulting expansions (the native modules and the trait definitions). The result is that you write something like:

image

and you get IDE documentation of the generated trait functions like

image

I included a couple examples.

Close #51


leighmcculloch (approved):
💯 🎉 🙏🏻

PR #116, opened 2022-06-16
Error handling and fallible contract calling

jayz22:
### What
- Implement conversion logic between `HostError` and `Status`
- Implement the fallible version of contract calling -- `try_call`
- Add tests to call contract within contract and verify the failing contract receives the correct status code. 
- Impl a binary function needed to unblock testing.

### Why

See https://github.com/stellar/rs-stellar-xdr/pull/85

### Known limitations
- Need to expose the `try_call` function through macro, so that it can be picked up by the sdk and guest. Currently only in the host. 
- Need to make sure `ScStatus` is only allowed as a return value, can't be stored.


Cargo.lock
```diff
@@ -189,7 +189,7 @@ version = "0.0.0"
 [[package]]
 name = "stellar-xdr"
 version = "0.0.0"
-source = "git+https://github.com/stellar/rs-stellar-xdr?rev=b2d367f04706af7f21a7a6a7abb4920e18dacadb#b2d367f04706af7f21a7a6a7abb4920e18dacadb"
```
graydon, line 192: Can you get the change into the XDR repo first?

stellar-contract-env-host/src/host.rs
```diff
@@ -33,33 +36,120 @@ use thiserror::Error;
...
 #[derive(Error, Debug)]
 pub enum HostError {
+    /// General error.
     #[error("general host error: {0}")]
     General(&'static str),
-    #[error("XDR error")]
-    XDRError(#[from] xdr::Error),
+    /// An error with specific status code.
+    #[error("host error: {0}, status: {:?}")]
+    WithStatus(String, ScStatus),
```
graydon, line 44: I don't think the String is required here -- the status code should have a method that produces the symbolic name, which ought to be enough for diagnosis?

graydon (approved):
Couple nits but otherwise this is looking really good!

graydon (approved):

PR #132, opened 2022-06-23
read tickets

graydon:
This is a WIP implementation of "read tickets", a mechanism for associating read permissions with objects and enforcing that only context frames granted access to an object can read it. It doesn't currently perform the permission-transfer necessary when handing an object handle from a contract caller to callee, so at the moment it _completely_ isolates objects allocated by frame X from those allocated by frame Y, but I'll finish this up tomorrow to allow such transfers.

I _think_ this is about the cheapest/simplest way to enforce this sort of thing. I sketched a few designs based on frame-local aliases / forwarding proxies and they just wind up messier and using more memory & CPU. Most of the time this just adds ~2 words to every object and does a couple bitmask tests.

- [ ] Ensure that the read ticket mask for every object allocated in an aborted subtransaction is set to "unreadable"
- [ ] Figure out how read tickets relate to debug events -- probably "should ignore them when formatting values for debug strings"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,12 +15,37 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "bit-set"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e11e16035ea35e4e5997b393eacbf6f63983188f7a2ad25bfb13465f5ad59de"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.6.3"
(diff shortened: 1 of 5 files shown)


leighmcculloch (commented):
@graydon Was this ready to merge and we forgot to review / merge it?

jonjove:
@leighmcculloch my understanding of this PR was that it was a proof-of-concept, but we weren't going to review it until after the current iteration. @graydon correct me if I'm wrong.

graydon:
Yeah this is a "come back to it in next iteration" change. I _think_ it's roughly the right shape (given the ugly problem space) but there's a little care required and it doesn't really block anything just yet.

graydon:
possibly ticket numbers should be "the next free object handle", and we can integrate this work with https://github.com/stellar/rs-soroban-env/issues/681

anupsdf:
Setting target milestone to `Testnet Phase 1` after discussing with @graydon

graydon:
Continuing in #930

PR #133, opened 2022-06-23
Add host functions for CAP-52

jonjove:
### What

Removes stale `get_last_operation_result` host function that no longer exists. Adds host functions specified in CAP-52 without implementations.

### Why

The CAP-52 host functions are need for the implementation of CAP-54.

### Known limitations

New host functions aren't actually implemented yet.

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -64,6 +64,9 @@ macro_rules! call_macro_with_all_host_functions {
 
             mod context "x" {
                 {"_", fn log_value(v:RawVal) -> RawVal }
+                /// Get the binary contractID of the contract which invoked the
+                /// running contract. Traps if the running contract was not
+                /// invoked by a contract.
                 {"0", fn get_invoking_contract() -> Object }
                 {"1", fn obj_cmp(a:RawVal, b:RawVal) -> i64 }
             }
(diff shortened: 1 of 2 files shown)


stellar-contract-env-common/src/env.rs
```diff
@@ -183,6 +183,13 @@ macro_rules! call_macro_with_all_host_functions {
                 {"_", fn compute_hash_sha256(x:Object) -> Object}
                 {"0", fn verify_sig_ed25519(x:Object, k:Object, s:Object) -> RawVal}
             }
+
+            mod account "a" {
+                {"_", fn get_low_threshold(a:Object) -> RawVal}
```
leighmcculloch, line 188: Could you add an `account_` prefix to these? We're using an object name prefix on host fns for most object types.

jonjove: Done.

leighmcculloch (commented):
I forgot to ask, can you add comments to the new functions so we don't need to double back and add them later? The macro contents support doc comments. Have a look at https://github.com/stellar/rs-stellar-contract-env/pull/131/files#diff-2706694cc99e5d5001390a8e291d5ea22144fd9a7b1c6f872c4cfaec61bbc0b9R87 which adds comments for some existing types.

jonjove:
@leighmcculloch done.

stellar-contract-env-common/src/env.rs
```diff
@@ -183,6 +186,24 @@ macro_rules! call_macro_with_all_host_functions {
                 {"_", fn compute_hash_sha256(x:Object) -> Object}
                 {"0", fn verify_sig_ed25519(x:Object, k:Object, s:Object) -> RawVal}
             }
+
+            mod account "a" {
+                /// Get the low threshold for the account with ed25519 public
+                /// key a (a is binary). Traps if no such account exists.
```
leighmcculloch, line 192: >Traps if no such account exists.

Any trap needs to be avoidable, so I guess we need a function such as `account_does_exist`.

leighmcculloch (approved):

stellar-contract-env-common/src/env.rs
jonjove: Agreed. I realized we need the same for `has_invoking_contract`. I'll get to that when I actually start consuming these functions next week.

PR #134, opened 2022-06-23
Add testutils feature and use it to emulate a Frame

jonjove:
### What

Adds a new feature that facilitates testing.

### Why

Makes it possible to call contract data functions from native tests.

### Known limitations

It would be better if we didn't need to use this "testutils" feature and instead could do this through `cfg(test)`.


graydon (approved):
Talked to jon about this today, looks like this is what everyone does (and it's fine by me).

PR #131, opened 2022-06-23
Implement `map` and `binary` host functions

jayz22:
### What

- [x] map
- [x] binary
- [ ] bignum
- [ ] special binary and interop 
- [ ] whatever left in CAP51

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

PR #136, opened 2022-06-24
Add impl of map_put, map_has, map_get

leighmcculloch:
### What

Add impl of map_put, map_has, map_get.

### Why

Need it for the SDK for user-defined type encoding as maps.

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -597,11 +597,18 @@ impl CheckedEnv for Host {
     }
 
     fn map_put(&self, m: Object, k: RawVal, v: RawVal) -> Result<Object, HostError> {
-        todo!()
+        let k = self.associate_raw_val(k);
+        let v = self.associate_raw_val(v);
+        let mnew = self.visit_obj(m, |hm: &HostMap| Ok(hm.update(k, v)))?;
+        Ok(self.add_host_object(mnew)?.into())
     }
 
     fn map_get(&self, m: Object, k: RawVal) -> Result<RawVal, HostError> {
-        todo!()
+        let k = self.associate_raw_val(k);
+        self.visit_obj(m, move |hm: &HostMap| match hm.get(&k) {
+            None => Err(HostError::General("key not found in map")),
+            Some(v) => Ok(v.to_raw()),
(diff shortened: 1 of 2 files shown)


leighmcculloch:
@jayz22 Not sure if I'm doing this correctly, please take a look when you have a moment even if it is post-merge.

stellar-contract-env-host/src/host.rs
```diff
@@ -597,11 +597,18 @@ impl CheckedEnv for Host {
     }
 
     fn map_put(&self, m: Object, k: RawVal, v: RawVal) -> Result<Object, HostError> {
-        todo!()
+        let k = self.associate_raw_val(k);
+        let v = self.associate_raw_val(v);
+        let mnew = self.visit_obj(m, |hm: &HostMap| Ok(hm.update(k, v)))?;
```
jayz22, line 602: Cool, I didn't see the `update` function and used `insert` instead. This looks much better.

leighmcculloch:
@jayz22 I realize now that you implemented map already in #131. I missed it because I didn't see "map" in the title oops. In any case it looks like my implementation is almost identical to yours. Sorry for the conflict!

jayz22 (commented):
Looks great! 
I've added most of the map and binary function in https://github.com/stellar/rs-stellar-contract-env/pull/131 . 
Feel free to copy them if needed. Will try to get that one merged soon.

stellar-contract-env-host/src/host.rs
leighmcculloch: Ah interesting, I didn't see insert. It looks like they do almost the same thing. The API for these collection types are interesting and not completely intuitive to me. i.e. I find it confusing that 'update' returns a copy and doesn't actually update the existing map, whilst 'insert' does. 🤷🏻

jayz22: Haha my thoughts exactly.

jayz22: Yeah, `update` is just a `clone + insert`: https://docs.rs/im/latest/src/im/ord/map.rs.html#832-836

PR #142, opened 2022-06-25
Misc gardening

graydon:
A few separate fixes here, probably worth reading by @leighmcculloch and @jayz22 (and maybe @sisuresh since it touches some of your tests) but nothing substantially weird or bad:

  - Move from ()-as-an-error in common to a specific named `ConversionError` struct
  - Make the Debug impls much better in common, using enum names and so forth
  - Make the budget arithmetic more efficient to check on the fly and partially update (as events grow)
  - Fix a couple error codes
  - Mostly: remove a ton of unwrap and should_panic in test code

One notable aspect of the last point is moving to `assert_matches!(...)` macros and matching on the error enum cases rather than the string messages in the error codes. I think this is a better thing to lean on for testing, not least because it lets us do semantic-rename of cases easily in the future, and also tests exactly the site at which the error is generated, and tests the code that the user is going to see out at the ScStatus level (keeping in mind that the informative extra strings are not propagated to the guest).

### DIFF
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -1,6 +1,8 @@
 use stellar_xdr::ScObjectType;
 
-use crate::{BitSet, Object, Status, Symbol, Tag, TagType, TaggedVal, Val};
+use crate::{
+    raw_val::ConversionError, BitSet, Object, Status, Symbol, Tag, TagType, TaggedVal, Val,
+};
 
 use super::{
     raw_val::{RawVal, RawValConvertible},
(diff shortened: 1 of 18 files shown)


graydon:
(I'm self-approving and landing this because none of it is at all architectural, it's just gardening; if you really object to any of it let me know and we can discuss / back out whatever later, but it'll be conflict-city if I don't land asap)

stellar-contract-env-common/src/env_val.rs
```diff
@@ -131,15 +133,15 @@ impl<E: Env> From<EnvVal<E, RawVal>> for RawVal {
 }
 
 impl<E: Env, T: TagType> TryFrom<EnvVal<E, RawVal>> for TaggedVal<T> {
-    type Error = ();
+    type Error = ConversionError;
```
leighmcculloch, line 136: Is it practical to have a unique error type per type so that it is extra clear what conversion is failing? I think it might be worth making it different for non-wasm builds too and include an indication between what type the value actually is to speed up debugging.

graydon: Possibly. I think we'll do another systematic pass over errors in the future before we finalize things to pick the right granularity -- the work here (along with the PR @jayz22 landed) is just to get us to a world where we're using typed errors split up into roughly the right shape.

jayz22: Created https://github.com/stellar/rs-stellar-contract-env/issues/158

PR #147, opened 2022-06-29
Add host functions: binary memcpy, BigInt

jayz22:
### What

### Why

### Known limitations

No tests yet.


stellar-contract-env-host/src/host.rs
```diff
@@ -1441,8 +1488,54 @@ impl CheckedEnv for Host {
...
+            }
+            let vnew: Vec<u8> = self.with_current_frame(|frame| match frame {
+                #[cfg(feature = "vm")]
+                Frame::ContractVM(vm) => vm.with_memory_access(|mem| {
+                    let mut vnew = hv.clone();
+                    /* we would potentially let the vec grow */
+                    if (i + l) > vnew.len() {
+                        vnew.resize(i + l, 0);
+                    }
+                    let _ = mem.get_into(j, &mut vnew.as_mut_slice()[i..(i + l)])?;
```
jonjove, line 1648: Nitpick: why do you need `let _ =` here?

stellar-contract-env-host/src/host.rs
```diff
@@ -1441,8 +1488,54 @@ impl CheckedEnv for Host {
...
+        // length of the slice to copy
+        let l: usize = u32::try_from(l).map_err(|_| {
+            HostError::WithStatus(
+                String::from("l must be u32"),
+                ScStatus::HostFunctionError(ScHostFnErrorCode::InputArgsWrongType),
+            )
+        })? as usize;
+
+        let res = self.visit_obj(b, move |hv: &Vec<u8>| {
+            if i > u32::MAX as usize - l {
```
jonjove, line 1634: I stared at this for a while wondering if it was safe, and I think it is _if_ all the earlier conversions are done safely. This might end up clearer if you did all the calculations in `u32`, then safely converted the final results to `usize`.

stellar-contract-env-host/src/host.rs
```diff
@@ -1432,7 +1432,54 @@ impl CheckedEnv for Host {
...
+                return Err(HostError::WithStatus(
+                    String::from("index out of bound"),
+                    ScStatus::HostObjectError(ScHostObjErrorCode::VecIndexOutOfBound),
+                ));
+            }
+
+            self.with_current_frame(|frame| match frame {
+                #[cfg(feature = "vm")]
+                Frame::ContractVM(vm) => vm.with_memory_access(|mem| {
+                    let _ = mem.set(j, &hv.as_slice()[i..(i + l)])?;
```
jonjove, line 1593: Nitpick: same thing about `let _ =`.

stellar-contract-env-host/src/host.rs
```diff
@@ -1432,7 +1432,54 @@ impl CheckedEnv for Host {
...
         l: RawVal,
     ) -> Result<RawVal, HostError> {
-        todo!()
+        // offset in the host binary
+        let i: usize = u32::try_from(i).map_err(|_| {
+            HostError::WithStatus(
+                String::from("i must be u32"),
+                ScStatus::HostFunctionError(ScHostFnErrorCode::InputArgsWrongType),
+            )
+        })? as usize;
```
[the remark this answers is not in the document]
jayz22: I see what you saying. Though I still feel we should be able to make the assumption that usize is at least 32-bits (i.e. no one will run the host crate in a 16-bit microcontroller). Otherwise, almost every other design decision in the host will fail, such as `RawVal` types.  I think there should be a top-level, global check that ensures that `usize.BITS >= 32` (or a similar check). cc @graydon

stellar-contract-env-host/src/host.rs
jayz22: Fixed.

stellar-contract-env-host/src/host.rs
```diff
@@ -1432,7 +1432,54 @@ impl CheckedEnv for Host {
         j: RawVal,
         l: RawVal,
     ) -> Result<RawVal, HostError> {
-        todo!()
+        // offset in the host binary
+        let i: usize = u32::try_from(i).map_err(|_| {
```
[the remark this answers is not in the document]
jayz22: Fixed.

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,131 @@ impl CheckedEnv for Host {
...
-    fn binary_copy_to_guest_mem(
+    fn binary_copy_to_linear_memory(
         &self,
         b: Object,
         i: RawVal,
         j: RawVal,
         l: RawVal,
     ) -> Result<RawVal, HostError> {
-        todo!()
+        //[offset in the host binary, pos in the linear memory, length of the slice to copy]
```
leighmcculloch, line 1559: We could name the parameters with these names instead of `i`, `j`, `l`, then the comment would be less necessary.

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,131 @@ impl CheckedEnv for Host {
...
+        #[cfg(not(feature = "vm"))]
```
leighmcculloch, line 1560: Why do some of the host functions require the full VM? It feels like we're entering a mode where the host crate should just always be in VM mode, and we should remove the feature because it always needs to be on. cc @graydon @jonjove 

(This comment doesn't need to block this PR.)

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,131 @@ impl CheckedEnv for Host {
...
+        todo!();
+        #[cfg(feature = "vm")]
+        if let [i, j, l] = [i, j, l]
+            .iter()
+            .map(|v| u32::try_from(*v))
+            .collect::<Result<Vec<u32>, ConversionError>>()?
+            .as_slice()
+        {
+            self.visit_obj(b, move |hv: &Vec<u8>| {
+                let end_idx = i.checked_add(*l).ok_or(HostError::WithStatus(
```
leighmcculloch, line 1570: If you use `ok_or_else` you can lazily create the host error, rather than always creating the host error.
```suggestion
                let end_idx = i.checked_add(*l).ok_or_else(|| HostError::WithStatus(
```

leighmcculloch (approved):
A couple comments, nothing blocking. I think @graydon should 👀 this too, but I'm approving so that we can keep moving forward.

stellar-contract-env-host/src/host.rs
jayz22: Thanks, fixed.

stellar-contract-env-host/src/host.rs
jayz22: Thanks, I learned something :)

stellar-contract-env-host/src/host.rs
leighmcculloch: I was meaning can we name the parameters in the function itself, all the way up in the host fn definition in the macro? It looks like ijl populate in a few places, including into the SDK PR, which isn't ideal.

jayz22: I see. Are you just referring to these functions, or host function parameter naming in general? Since all current hostfn parameters are 1 char (at most 2) in length and it helps keeping things concise and compact (not sure if there is impact on code size too). I'm not sure if it's worth changing these names to break from the convention. Although I do see these indices are a bit less obvious than the other hostfns.
It would be look like:
`{"2", fn binary_copy_from_linear_memory(b:Object, offset_hv:RawVal, pos_lm:RawVal, len:RawVal) -> Object}`

leighmcculloch: >Are you just referring to these functions, or host function parameter naming in general? 

Most other functions are easily understood even if the arguments are single letters. For example, the following is reasonably understood that `a` will be an account and `s` will be a signer.
https://github.com/stellar/rs-stellar-contract-env/blob/e9128da4635e23053084e24359d42f9348690bad/stellar-contract-env-common/src/env.rs#L302

But some functions it is very unclear. The letters `i`, `j`, and `l` don't appear to align at all semantically with what is intended to be passed in to the function. I also think it is unlikely that if we used names like `o`, `p`, and `l`, that it would help. Some host functions need more descriptive names. We've used more descriptive names for some functions before, so I don't think we're breaking from convention. For example:
https://github.com/stellar/rs-stellar-contract-env/blob/e9128da4635e23053084e24359d42f9348690bad/stellar-contract-env-common/src/env.rs#L159-L160
https://github.com/stellar/rs-stellar-contract-env/blob/e9128da4635e23053084e24359d42f9348690bad/stellar-contract-env-common/src/env.rs#L166

leighmcculloch: Might I suggest something like:

```
{"2", fn binary_copy_from_linear_memory(b:Object, b_pos:RawVal, lm_pos:RawVal, len:RawVal) -> Object}
```

jayz22: > Some host functions need more descriptive names. We've used more descriptive names for some functions before, so I don't think we're breaking from convention. For example:

Ah yes, didn't see those. 

Thank you, I will update those names.

stellar-contract-env-host/src/host.rs
jayz22: Yeah these ones do, since they need access to the linear memory, which is part of the VM. 
I am certainly in favor of removing the VM feature toggle.

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,137 @@ impl CheckedEnv for Host {
...
+            let mut vnew = self.visit_obj(b, |hv: &Vec<u8>| Ok(hv.clone()))?;
+            let end_idx = b_pos.checked_add(*len).ok_or_else(|| {
+                HostError::WithStatus(
+                    String::from("u32 overflow"),
+                    ScStatus::HostFunctionError(ScHostFnErrorCode::InputArgsInvalid),
+                )
+            })? as usize;
+            self.with_current_frame(|frame| match frame {
+                Frame::ContractVM(vm) => vm.with_memory_access(|mem| {
+                    /* we would potentially let the vec grow */
```
jonjove, line 1628: Nitpick: this is a one-line comment so use the `//` style

jonjove (approved):
I have some minor style comments, and one bigger maintainability comment. I don't really want to block this PR to fix the maintainability comment since it fits into the bigger picture of #154. But let's get this all cleaned up in another PR.

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,137 @@ impl CheckedEnv for Host {
...
-        l: RawVal,
+        b_pos: RawVal,
+        lm_pos: RawVal,
+        len: RawVal,
     ) -> Result<RawVal, HostError> {
-        todo!()
+        #[cfg(not(feature = "vm"))]
+        todo!();
+        #[cfg(feature = "vm")]
+        if let [b_pos, lm_pos, len] = [b_pos, lm_pos, len]
```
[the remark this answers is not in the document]
jayz22: Thank you for the suggestion, I will incorporate this when fixing the other issue.

stellar-contract-env-host/src/host.rs
jayz22: Fixed.

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,137 @@ impl CheckedEnv for Host {
...
-        i: RawVal,
-        j: RawVal,
-        l: RawVal,
+        b_pos: RawVal,
+        lm_pos: RawVal,
+        len: RawVal,
     ) -> Result<RawVal, HostError> {
-        todo!()
+        #[cfg(not(feature = "vm"))]
+        todo!();
```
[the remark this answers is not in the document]
jayz22: Fixed them here.

stellar-contract-env-host/src/host.rs
graydon: To me, having the "vm" feature be explicit has been helpful in making sure that the host can be meaningfully configured/built in its "native" mode (where guest calls host directly) -- I am certain we're not passing through any VM infrasturcture along the way because it's all compiled-out. So I'd _like_ to maintain it, but won't insist if it's getting too painful.

That said, in this case making the "vm" feature mandatory wouldn't make this (or _any_ of the linear-memory-accessing functions) make any more sense for the native-build (guest == host) case. Indeed, it's good that we got stopped here to consider it, because this is one of the cases where we're at risk of making the native build incoherent. These functions simply don't make sense as-written for native builds. They either need to have different signatures or else be avoided in native builds, be performed SDK-side.

Think about what the parameters to these functions mean: they're 32-bit numbers that represent portions of the guest memory. In a native build, guest memory is host memory. This causes two issues:

  1. That guest/host memory is probably 64 bit, not 32 bit.
  2. Even if you make it handle 64-bit correctly, it's not entirely clear we'll be able to have (or keep) a non-UB interface to converting from a u64 back to a pointer (under the forthcoming [strict provenance](https://github.com/rust-lang/rust/issues/95228) model in Rust, anyway).

I'm not 100% sure what the best way around this is right now.

The simplest thing -- and what I'd suggest trying "for now" -- would be to define it to work with arguments of multiple integer types -- u32, u63, or even u64-in-an-object -- and implement the non-VM case by converting the values back to a raw pointers (like just using `u64 as *mut u8`) and doing an unsafe memcpy with the resulting pointers. This will work in today's rust but may be harder to support in the future (or require more elaborate ceremony to tell the compiler "we know what we're doing" -- the strict provenance docs _say_ they're going to keep supporting this kind of conversion, somehow). In the VM case, just trap on anything non-u32.

There might be other more-awkward options to explore in the future, such as injecting a `base pointer` into the host when it's constructed and requiring all passed pointers to guest-to-binary / binary-to-guest functions be within the same provenance as the base pointer, or something, and then doing the guest-to-guest operations strictly SDK-side.

jayz22:
My apologies for one more forced push. Previously my git config was not set properly so my `git pull` was doing a rebase by default and introduced some artifacts in the commit history. It should be configured properly going forward.

stellar-contract-env-host/src/host.rs
```diff
@@ -1549,30 +1549,137 @@ impl CheckedEnv for Host {
...
+        if let [b_pos, lm_pos, len] = [b_pos, lm_pos, len]
+            .iter()
+            .map(|v| u32::try_from(*v))
+            .collect::<Result<Vec<u32>, ConversionError>>()?
+            .as_slice()
+        {
+            self.visit_obj(b, move |hv: &Vec<u8>| {
+                let end_idx = b_pos.checked_add(*len).ok_or_else(|| {
+                    HostError::WithStatus(
+                        String::from("u32 overflow"),
```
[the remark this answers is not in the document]
leighmcculloch: ☝🏻

leighmcculloch (approved):
I think this is fine to merge now if you want, there are two comments marked with emojis below that I think could be addressed either now or separately.

stellar-contract-env-host/src/host.rs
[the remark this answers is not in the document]
jayz22: Thanks for the suggestion. I will improve this in a follow up PR (broader error handling related refactoring and improvements).

jayz22:
> I think this is fine to merge now if you want, there are two comments marked with emojis below that I think could be addressed either now or separately.

Yeah I think it's good to merge it now (after fixing the build). I will address the two stylistic comments as part of a broader cleanup/enhancement. 
The main thing I need to address is Graydon's comment https://github.com/stellar/rs-stellar-contract-env/pull/147/#discussion_r916089138 , and created https://github.com/stellar/rs-stellar-contract-env/issues/179.

jayz22:
> Yeah I think it's good to merge it now (after fixing the build)

@leighmcculloch Should be good to go now.

stellar-contract-env-host/src/host.rs
graydon: @leighmcculloch I'm not totally sure this _can_ be done SDK-side, at least not with a lot of mode-specific logic (not sure how much of that you have already); does the SDK have the ability to retrieve a "true" host object pointer through which it could locate the interior of a Binary, rather than by handle?

leighmcculloch: We could add functions to the `Host` to expose that I guess. The SDK can call fns added to Host that are not part of the `Env` trait.

PR #148, opened 2022-06-30
Re-export the exact version of im_rc used in the host.

graydon:
Theoretically we can try to make downstream clients such as stellar-core use "the same" `im_rc` version through matching `Cargo.toml` file entries; but it is less error-prone and friendlier to multi-versioning if we just re-export the version used by the host from the host itself.


leighmcculloch:
I would expect that if stellar-core used a different version of im_rc that we'd get compile errors when attempting to use them together since the types won't be interchangeable. Does that not happen? What's the error-prone situation we're attempting to prevent?

PR #159, opened 2022-07-03
Pass `Object` type to bigint `shl`, `shr`, `pow` fn signatures

jayz22:
### What

Pass in bigint objects to `bigint_shl/shr/pow` fn signatures, and provide sign and overflow checks within these functions. 

### Why

The `num_bigint` api of `shl`, `shr`, `pow` has internal checks for [sign](https://docs.rs/num-bigint/latest/src/num_bigint/biguint/shift.rs.html#13) and [overflow](https://docs.rs/num-bigint/latest/src/num_bigint/biguint/power.rs.html#25), which may panic if the inputs do not fit. Instead of restricting our hostfn api (e.g taking a `u32` as the exponent in `bigint_pow`), we should still pass those parameters as `BigInt` and check their validity from within the host functions. This makes the bigint hostfn apis more uniform and easier to maintain. 

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -171,29 +171,55 @@ macro_rules! call_macro_with_all_host_functions {
             }
 
             mod bigint "g" {
+                /// Constructs a BigInt from an u64.
                 {"_", fn bigint_from_u64(x:u64) -> Object}
+                /// Converts a BigInt to an u64. Traps if the value cannot fit into u64.
                 {"0", fn bigint_to_u64(x:Object) -> u64}
+                /// Constructs a BigInt from an i64.
                 {"1", fn bigint_from_i64(x:i64) -> Object}
+                /// Converts a BigInt to an i64. Traps if the value cannot fit into i64.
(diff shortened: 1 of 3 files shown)


leighmcculloch:
What's the actual valid range for these values?

jayz22:
> What's the actual valid range for these values?

There is no explicit range limits on these values. Limits come from memory and pointer size. The bigint is stored as an array of u64 and array size cannot exceed `usize::MAX`.

leighmcculloch:
This needs to be well defined as part of the CAP that introduces the type, rather than implementation defined.

leighmcculloch:
I note in the description of this PR:

>Pass in bigint objects to `bigint_shl/shr/pow`

I note in the CAP that the following comments are being made:

>1. The operand of the left and right shift operations (`bigint_shl` and `bigint_shr` cannot be negative and cannot exceed size of `u64`. 
>2. The exponent in `bigint_pow` cannot be negative and cannot exceed size of `u64`.

If the limit for these is U64, why not use the U64 object type instead of the BigInt object type?

leighmcculloch (approved):

Issue #160, opened 2022-07-05
Improve cost model / budget machinery


graydon:
The current budget.rs cost model is inadequate for serious use. It needs:

  - [ ] To be able to load a batch of parameters from `CONFIG_SETTING` LEs
  - [X] To use a more uniform and lightweight interface for defining multiple cost centers (currently one has to define a number of new variables and functions in a number of structures and traits)
  - [X] To handle cost functions that are parameterized by constants but have nonlinear relationships to inputs (eg. charging log_16(sizeof(map)) CPU for a map lookup)
  - [X] To handle systematic/automatic calibration of costs by measurement -- both taking measurements over a range of input values and then fitting cost-function parameters to those measurements

graydon:
Wednesday update:
  - factored out the measurement and reporting code
  - connected it to model-fitting directly
  - added memory-allocation measurement

It's now fairly easy to add new calibration benchmarks, as illustrated.

The good news is that the vector-allocation benchmark shows low costs in line with expectations for vector operations: the per-cell CPU instruction cost is about the same as 2 WASM instructions would be, and the per-cell memory cost is exactly 16 bytes as expected (1 word for the value + 1 word for the host weak pointer).

graydon:
See https://github.com/stellar/rs-stellar-contract-env/pull/164 -- still needs to be wired up to the CONFIG_SETTING LEs  but it's getting there.

graydon:
Initial work is done here, remainder (which is involved) to be continued in #305

PR #164, opened 2022-07-06
Overhaul the budget system with general cost models, add model fitting.

graydon:
I'm out of time for this evening, but I promised @jayz22 a little material and this is as far as I got. It compiles and half-works, but doesn't work enough to merge just yet. You're welcome to try to brush it up into something mergeable, or just stare at it and get an idea for the direction I think we should be going / where I'm going to try to take it next week.

(this addresses #160)

What's here:
  - General cost models rather than just linear factors. A cost model is a parameterized function with a constant, linear, log, log-base, and quadratic term. These parameters will eventually be loaded from ledger `CONFIG_SETTING` entries, but for now they're all zero except the one I added for tracking CPU insns per WASM insn.
  - Uniform interface to multiple cost models in multiple different dimensions. All the data structures are just keyed by a single type code for each cost type (it's an explicit enum in this PR -- it should go into the XDR), and the interface for charging costs is just calling `budget.charge(CostType::Foo, n)?` where `n` is the input domain for the cost model. It'll look up the cost model for each dimension (cpu and memory), transform `n` to a cost under the cost model, charge the dimension, and return an error if the dimension is over budget.
  - Model fitting for offline calibration of the cost models. I added enough machinery to _almost_ convince myself that we can empirically measure the costs of host functions using calibration benchmarks and then feed those benchmarks into a stock gradient-descent optimizer to do least-squares curve fitting of the model to the data (code is in the PR here, but I did test it and it was able to recover at least a linear and a quadratic curve from sample data I ran it on). I _think_ this is a viable path for calibrating cost models, but I'll need to return to it later.

Comments welcome! It's very rough, but I think it's close to the shape we'll want going forward.

### DIFF
--- a/stellar-contract-env-host/Cargo.toml
+++ b/stellar-contract-env-host/Cargo.toml
@@ -22,6 +22,9 @@ num-integer = "0.1.45"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
+optimization = "0.2.0"
+env_logger = "0.9.0"
+log = "0.4.17"
 
 [features]
 vm = ["wasmi", "parity-wasm", "stellar-contract-env-common/vm"]
@@ -29,7 +32,9 @@ testutils = []
 
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 perf-event = "0.4.7"
+tracking-allocator = "0.4.0"
 tabwriter = "1.2.1"
+thousands = "0.2.0"
 
 [[bench]]
 required-features = ["vm"]
@@ -37,3 +42,9 @@ harness = false
 bench = true
 name = "calibrate_wasm_insns"
 path = "benches/calibrate_wasm_insns.rs"
+
+[[bench]]
+harness = false
+bench = true
+name = "calibrate_host_ops"
+path = "benches/calibrate_host_ops.rs"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,30 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "assert_matches"
 version = "1.5.0"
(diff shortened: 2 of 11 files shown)


graydon:
Wednesday update:
  - factored out the measurement and reporting code
  - connected it to model-fitting directly
  - added memory-allocation measurement

It's now fairly easy to add new calibration benchmarks, as illustrated.

The good news is that the vector-allocation benchmark shows low costs in line with expectations for vector operations: the per-cell CPU instruction cost is about the same as 2 WASM instructions would be, and the per-cell memory cost is exactly 16 bytes as expected (1 word for the value + 1 word for the host weak pointer).

stellar-contract-env-host/Cargo.toml
```diff
@@ -22,18 +22,30 @@ num-integer = "0.1.45"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
+optimization = "0.2.0"
+env_logger = "0.9.0"
+log = "0.4.17"
 
 [features]
 vm = ["wasmi", "parity-wasm", "stellar-contract-env-common/vm"]
+default = ["vm"]
```
leighmcculloch, line 31: 👀 Can we leave this off? I think the day we make this default, we should get rid of the feature.

leighmcculloch (commented):
One request, otherwise LGTM. 🎉 

I don't have an opinion on the cost model parts.

For the CLI it would be nice to add another function alongside modify_budget so that the CLI can call a function that doesn't be given a mutable value. But we can add this later.

stellar-contract-env-host/Cargo.toml
graydon: Oh, yeah, that came along by mistake. Nice catch.

graydon: fixed.

graydon:
@leighmcculloch re "For the CLI it would be nice to add another function alongside modify_budget so that the CLI can call a function that doesn't be given a mutable value": added `host.get_budget(|budget| ... )` and renamed `modify_budget` to `get_budget_mut`. also cleaned up the budget interface a bit so you can call `budget.get_input(CostType::WasmInsnExec)` -- or whatever, pass each cost type you like -- to see/report the cost-model _inputs_ separately from the accumulated memory/cpu output costs charged to the budget.

leighmcculloch (approved):

PR #181, opened 2022-07-08
Include data types in `ConversionError`

jayz22:
### What

Solves https://github.com/stellar/rs-stellar-contract-env/issues/158

~This draft is the solution I had in mind. Looked very close, until I hit a Rust-related road block.. See inline comments. Help appreciated.~

Main changes: 
- A new `RawVal::ConversionError<T>{ f: RawVal, t: PhantomData<T>}` that is able to display the correct "from" and "to type" information when a type conversion error occurs.
- Add a new enum case `HostError::ConversionError(String)` that bubbles up the 'RawVal::ConversionError<T>` with only a string. 
- Add a simple `tuple` test case to demonstrate it works.

### Why


stellar-contract-env-common/src/raw_val.rs
```diff
@@ -61,7 +61,18 @@ impl AsMut<RawVal> for RawVal {
...
+impl<F, T> core::fmt::Display for ConversionError<F, T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "Conversion from {} to {} failed",
+            std::any::type_name::<F>(),
+            std::any::type_name::<T>()
+        )
+    }
+}
```
jayz22, line 75: The idea is to pass in types `F` and `T` (denoting the data *from* and *to* for the conversion) into ConversionError but with no actual data. Then use the `std::any::type_name::<T>` to print out a string representation of the type `T`.  Implement `core::fmt::Display` or `Debug` for `ConversionError` so that it can get back a string. 

The code as presented complains about unused parameter `F` and `T` in `pub struct ConversionError<F, T>`. 
I tried `core::fmt:Debug` as well and it wouldn't work either due to a different error. 
I tried putting in `PhantomData` inside `ConversionError`, i.e. 

`pub struct ConversionError<F, T> {
    _f: PhantomData<F>,
    _t: PhantomData<T>,
}`

But couldn't figure out how to make the rest of it work. I will try a bit more tomorrow. 

@leighmcculloch Does this align with what you had in mind?  
Suggestions are welcome. cc @graydon @jonjove

stellar-contract-env-common/src/env_val.rs
```diff
@@ -209,7 +209,7 @@ impl<E: Env> IntoEnvVal<E, RawVal> for i64 {
 }
 
 impl<E: Env> TryFrom<EnvVal<E, RawVal>> for u64 {
-    type Error = ConversionError;
+    type Error = ConversionError<EnvVal<E, RawVal>, u64>;
```
leighmcculloch, line 212: In my suggestion below what I mean is make this as follows, to indicate that it is a ConversionError destined for a `u64`:
```suggestion
    type Error = ConversionError<u64>;
```

stellar-contract-env-common/src/raw_val.rs
```diff
@@ -61,7 +61,18 @@ impl AsMut<RawVal> for RawVal {
 // away, the same way `()` would, while remaining a separate type to allow
 // conversion to a more-structured error code at a higher level.
 #[derive(Debug)]
-pub struct ConversionError;
+pub struct ConversionError<F, T>;
```
leighmcculloch, line 64: In my suggestion below what I mean is make this as follows:

```suggestion
pub struct ConversionError<T> {
    #[cfg(not(target_family = "wasm"))]
    /// Source is the RawVal that was attempted to be converted into T, but that failed conversion.
    pub source: RawVal;
}
```

stellar-contract-env-common/src/raw_val.rs
```diff
@@ -61,7 +61,18 @@ impl AsMut<RawVal> for RawVal {
...
+
+impl<F, T> core::fmt::Display for ConversionError<F, T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "Conversion from {} to {} failed",
+            std::any::type_name::<F>(),
+            std::any::type_name::<T>()
```
leighmcculloch, line 72: In my suggestion below what I mean is make it something like this:
```suggestion
            "Conversion from {} to {} failed",
            self.source,
            std::any::type_name::<T>()
```

leighmcculloch (commented):
This looks really cool. I replied to your main comment below. The other comments I left are describing at each point what I might do. I'm not sure if what I'm suggesting is best though, as I'd probably need to play with the type system a bit to confirm if it makes sense.

stellar-contract-env-common/src/raw_val.rs
jayz22: Thanks @leighmcculloch , I like your suggestion of passing in a `RawVal` and leverage the printing feature of it. 

I made the changes you suggested and got it basically working, see updated description. 

I'm not sure if I fully understand what you meant by:

> Will having generics in the error make it difficult to wrap this error type?

But I think I addressed it in my code. See bullet point #2 in the description.

stellar-contract-env-common/src/raw_val.rs
```diff
@@ -61,7 +61,21 @@ impl AsMut<RawVal> for RawVal {
 // away, the same way `()` would, while remaining a separate type to allow
 // conversion to a more-structured error code at a higher level.
 #[derive(Debug)]
-pub struct ConversionError;
+pub struct ConversionError<T> {
+    pub f: RawVal,
+    pub t: PhantomData<T>,
```
leighmcculloch, line 66: Could we give these fields names? I find short names are useful when the field name would repeat the type (e.g. if the type name was already From, a field name of from is not useful), or when the field is constrained to a small scope, neither of which is the case here.
```suggestion
    pub from: RawVal,
    pub to: PhantomData<T>,
```

stellar-contract-env-common/src/tuple.rs
```diff
@@ -5,29 +5,31 @@ use crate::{
...
+use core::marker::PhantomData;
+
 macro_rules! impl_for_tuple {
     ( $count:literal $($typ:ident $idx:tt)+ ) => {
         impl<E: Env, $($typ),*> TryFrom<EnvVal<E, RawVal>> for ($($typ),*)
         where
             $($typ: TryFrom<EnvVal<E, RawVal>>),*
         {
-            type Error = ConversionError;
+            type Error = ConversionError<($($typ),*)>;
```
leighmcculloch, line 16: If you merge main into your branch I think you'll need to change this to:
```suggestion
            type Error = ConversionError<($($typ,)*)>;
```
A fix for tuples was merged so that 1-ary tuples are supported, which requires the trailing comma to distinguish a value in parenthesis from a tuple that uses parenthesis.

stellar-contract-env-host/src/host.rs
```diff
@@ -48,6 +49,8 @@ pub enum HostError {
     /// An error with specific status code.
     #[error("host error: {0}, status: {:?}")]
     WithStatus(String, ScStatus),
+    #[error("{0}")]
+    ConversionError(String),
```
leighmcculloch, line 53: Nice idea 👏🏻.

leighmcculloch (commented):
I like it, and the conversion to the string version is a way to address the upstream impact nicely too.

Will we get any difficulty if we start nesting these in the SDK do you think? For example, if I have a user-defined type (Udt) that contains a u32, and a u64, what should the Udt's `Error` type be? Should it simply be `ConversionError<Udt>` and then map conversion errors of the internal types to its type? I guess so, I think that makes sense, although the source error will be lost. I guess that's not a huge problem, and the source problem would hopefully be visible since the RawVal would be outputted.

You need to merge main in, and there's at least one change that'll be required after doing that.

One request about naming the fields.

stellar-contract-env-common/src/raw_val.rs
jayz22: @leighmcculloch I had to remove the `#[cfg(not(target_family = "wasm"))]` guard because the `ConversionError` being constructed and used in multiple places and I can't find an elegant way to toggle it without putting this guard everywhere. 
> The reason to put all this behind a not-wasm cfg is so that when testing locally this nice error is outputted, but then on chain all this code disappears.

Is it a requirement for this to not show up on-chain?

stellar-contract-env-common/src/tuple.rs
jayz22: Fixed.

stellar-contract-env-common/src/raw_val.rs
jayz22: Fixed.

stellar-contract-env-common/src/raw_val.rs
leighmcculloch: When you introduce the cfg, what size difference do the wasm files have? That's the only thing we can easily measure right at this moment.

I'm inclined to say if there's no substantial size difference run with it without the cfg, and create an issue for evaluating the execution cost in September.

jayz22:
@leighmcculloch I addressed all your comments. Just one comment above regarding `target_family = "wasm"` attribute. 

Re: nesting errors in UDT. Yes you are right, in this case it will just be a`ConversionError<Udt>` and the error will read "Conversion error from RawVal to Udt" with the internal error lost. I agree this is not ideal and I'll continue to think of ways to improve it. 

One minor thing, I think the `Makefile` needs to be updated? `make all` didn't catch the build error for me which occurred in the build pipeline: `cargo check --profile dev --target wasm32-unknown-unknown --no-default-features`.

leighmcculloch:
> One minor thing, I think the `Makefile` needs to be updated? `make all` didn't catch the build error for me which occurred in the build pipeline: `cargo check --profile dev --target wasm32-unknown-unknown --no-default-features`.

Could you add the following line to the Makefile's build task?
```diff
	cargo hack --feature-powerset check
+	cargo hack --feature-powerset check --target wasm32-unknown-unknown
```
You might need to exclude the host package from that build. If you need to do that you can use `--workspace --exclude=...`.

stellar-contract-env-common/src/raw_val.rs
jayz22: I see, then I agree it is important for it to not show up wasm-build.

leighmcculloch (commented):
I think this looks good. I don't feel like I've totally given up on the idea of embedding the actual RawVal value inside the ConversionError, but understand the road blocks with doing that weren't trivial.

Before we merge this I think we just need to make sure it's not a mountain to climb to get the SDK updated too, just so we don't block the SDK from other updates until it's update too.

jayz22:
I agree it is far useful to include the `RawVal` inside the `ConversionError`. I have (again) switched back to that. The only notable change is I need to add the `Eq, PartialEq` to `RawVal` as it's needed for the ConversionError.
The sdk side changes are here: https://github.com/stellar/rs-stellar-contract-sdk/pull/208 (not fully working yet).

stellar-contract-env-common/src/raw_val.rs
```diff
@@ -42,7 +42,7 @@ pub enum Tag {
 }
 
 #[repr(transparent)]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, PartialEq, Eq)]
```
leighmcculloch, line 45: Can you add a comment pointing out that RawVal equality if based on the RawVal alone and not the host, so two equal objects may be referenced by inequal RawVals. I think this behavior is appropriate, it's too inconvenient to never be able to compare them and we shouldn't need a host to do comparisons.

jayz22: @leighmcculloch I added the comment.

jonjove: I agree that `obj_cmp` is the proper way to compare objects, but people will inadvertently use this by accident and then there will be bugs. I'm really not sure I'm comfortable with this change.

leighmcculloch: Removing the Eq from ConversionError will make it harder to write tests for any function that can return the error.

leighmcculloch: @graydon A while ago you mentioned the possibility for interning host object types. Is that something planned? If host objects are interned then this problem would no longer be relevant.

jayz22:
Closing this as the underlying event and error has been overhaul. Will continue with new approach in https://github.com/stellar/rs-stellar-contract-env/pull/246.

PR #182, opened 2022-07-08
Add conversions to Result<_, Status>

leighmcculloch:
### What

Add conversions from `RawVal` to `Result<RawVal, Status>` and `Result<_, Status>`.

### Why

Simplify the process of converting a `RawVal` that could either be a value or a `Status` into that value, or exit early, or into some other type that the `RawVal` represents in the success case.

### Known limitations

[TODO or N/A]


leighmcculloch:
@jonjove convinced me to find other ways to write code that has to do this. Related conversation https://discord.com/channels/897514728459468821/994750038314070096/995011554175107083.

PR #197, opened 2022-07-12
Make Symbol usage in match statements

leighmcculloch:
### What

Make Symbol its own standalone type separate from TaggedVal<TagSymbol>.

### Why

Symbol can't be used in match patterns because the RawVal type is not `#[derive(Eq, PartialEq)]`, and types must derive those traits to be used in pattern matching. 

This has an unfortunate affect that mapping we need to do must occur within a set of `if` blocks, which Rust does not optimize as well.

For a small match block of only 3 conditions this causes a Symbol mapping to be an extra 37 bytes in compiled WASM. Given that we use Symbols as constants in enum encoding, it seems like a substantial benefit to make it its own type.

An alternative would be to simply derive the equality traits on RawVal.

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-common/src/symbol.rs
+++ b/stellar-contract-env-common/src/symbol.rs
@@ -95,7 +95,7 @@ impl wasmi::FromValue for Symbol {
 
 impl Hash for Symbol {
     fn hash<H: Hasher>(&self, state: &mut H) {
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Closing because this introduces overhead that doesn't match the same savings we get if RawVal was usable in match.

PR #203, opened 2022-07-13
Add ENV_INTERFACE_VERSION and a big comment

graydon:
This is a first attempt at the problem of declaring interface versions. It's _mostly_ just a comment explaining things, plus a constant. This relates to https://github.com/stellar/stellar-core/pull/3471 which is for implementation versioning.

### DIFF
--- a/stellar-contract-env-common/src/version.rs
+++ b/stellar-contract-env-common/src/version.rs
@@ -1,0 +1,36 @@
+/// There are two different notions of versioning in this system:
+///
+///   1. Interface versioning
+///   2. Implementation versioning
+///
+/// Interface versioning controls the willingness of a host to _try_ to run a
+/// _new_ transaction against some contract. For this, a contract embeds an
+/// interface version number _inside itself_ and the host inspects it before
+/// starting the contract. If the contract's interface version is outside the
+/// supported range for a host, the host will fail the transaction with a useful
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):

stellar-contract-env-common/src/version.rs
```diff
@@ -0,0 +1,36 @@
...
+/// of the host crate: strictly identical versions for new transactions entering
+/// the network, and observably-identical ones for replay of historical
+/// transactions (with a policy to allow expiring old versions that differ only
+/// in ways no transactions in the public network history actually observe).
+/// Implementation versioning will also be done by storing a version on-chain
+/// that changes by consensus, but the host implementation version number is
+/// _not_ compiled into a contract, since the same contract will be expected to
+/// run against multliple implementations over a long period of time.
+
+pub const ENV_INTERFACE_VERSION: u64 = 1;
```
leighmcculloch, line 36: The ENV_ prefix feels repetitive given we're in the env crate. In another way it also feels too specific, since the Env is an interface, but the conceptual interface that needs guarding is really broader than the host fns alone since it also includes things like the RawVal type.

leighmcculloch:
```suggestion
pub const INTERFACE_VERSION: u64 = 1;
```

PR #206, opened 2022-07-13
Register Test Contract

jonjove:
### What

Add machinery to register a test contract and call it.

### Why

Testing multiple contracts is basically impossible right now. This facilitates calling contracts by contract identifier.

### Known limitations

The interface isn't amazing.

### DIFF
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -177,6 +177,27 @@ struct RollbackPoint {
     objects: usize,
 }
 
+#[cfg(feature = "testutils")]
+pub trait ContractVTableFnTrait {
+    fn call(&self, args: &[RawVal]) -> RawVal;
+
+    fn duplicate(&self) -> ContractVTableFn;
+}
+
+#[cfg(feature = "testutils")]
+pub struct ContractVTableFn(pub Box<dyn ContractVTableFnTrait>);
+
+#[cfg(feature = "testutils")]
+impl Clone for ContractVTableFn {
+    fn clone(&self) -> Self {
+        self.0.duplicate()
+    }
+}
+
+#[cfg(feature = "testutils")]
+#[derive(Clone)]
+pub struct ContractVTable(pub std::collections::HashMap<Symbol, ContractVTableFn>);
+
 /// Holds contextual information about a single invocation, either
 /// a reference to a contract [`Vm`] or an enclosing [`HostFunction`]
 /// invocation.
@@ -203,6 +224,8 @@ pub(crate) struct HostImpl {
     storage: RefCell<Storage>,
     context: RefCell<Vec<Frame>>,
     budget: RefCell<Budget>,
+    #[cfg(feature = "testutils")]
+    vtables: RefCell<std::collections::HashMap<Hash, ContractVTable>>,
 }
 
 /// A guard struct that exists to call [`Host::pop_frame`] when it is dropped,
@@ -254,6 +277,8 @@ impl Host {
             storage: RefCell::new(storage),
             context: Default::default(),
             budget: Default::default(),
+            #[cfg(feature = "testutils")]
+            vtables: Default::default(),
         }))
     }
 
@@ -868,6 +893,28 @@ impl Host {
         self.visit_storage(|storage| match storage.get(&acc)?.data {
             LedgerEntryData::Account(ae) => Ok(ae),
             _ => Err(HostError::General("not account")),
+        })
+    }
+
+    #[cfg(feature = "testutils")]
+    pub fn register_test_contract(
+        &self,
+        contract_id: Object,
+        vtable: ContractVTable,
+    ) -> Result<(), HostError> {
+        self.visit_obj(contract_id, |bin: &Vec<u8>| {
+            let mut vtables = self.0.vtables.borrow_mut();
+            let hash = Hash(
+                bin.clone()
+                    .try_into()
+                    .map_err(|_| HostError::General("bad contract id"))?,
+            );
+            if !vtables.contains_key(&hash) {
+                vtables.insert(hash, vtable);
+                Ok(())
+            } else {
+                Err(HostError::General("vtable already exists"))
+            }
         })
     }
 }
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -179,7 +179,7 @@ struct RollbackPoint {
 
 #[cfg(feature = "testutils")]
 #[derive(Clone)]
-pub struct ContractVTable(
+pub struct ContractFunctionSet(
     pub std::collections::HashMap<Symbol, &'static dyn Fn(Host, &[RawVal]) -> RawVal>,
 );
 
@@ -210,7 +210,7 @@ pub(crate) struct HostImpl {
     context: RefCell<Vec<Frame>>,
     budget: RefCell<Budget>,
     #[cfg(feature = "testutils")]
-    vtables: RefCell<std::collections::HashMap<Hash, ContractVTable>>,
+    contracts: RefCell<std::collections::HashMap<Hash, ContractFunctionSet>>,
 }
 
 /// A guard struct that exists to call [`Host::pop_frame`] when it is dropped,
@@ -263,7 +263,7 @@ impl Host {
             context: Default::default(),
             budget: Default::default(),
             #[cfg(feature = "testutils")]
-            vtables: Default::default(),
+            contracts: Default::default(),
         }))
     }
 
@@ -805,7 +805,7 @@ impl Host {
         })?;
 
         #[cfg(feature = "testutils")]
-        if let Some(vtable) = self.0.vtables.borrow().get(&id) {
+        if let Some(vtable) = self.0.contracts.borrow().get(&id) {
             if let Some(f) = vtable.0.get(&func) {
                 let mut frame_guard = self.push_test_frame(id.clone());
                 let res = Ok(f(self.clone(), args));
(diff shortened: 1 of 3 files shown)


stellar-contract-env-host/src/host.rs
```diff
@@ -177,6 +177,27 @@ struct RollbackPoint {
     objects: usize,
 }
 
+#[cfg(feature = "testutils")]
+pub trait ContractVTableFnTrait {
```
leighmcculloch, line 181: I understand what a vtable, but I don't think it should be a concept exposed by the SDK. What does a `ContractVTableFnTrait` represent? Does it represent a function that can be called with a list of arguments?
```suggestion
pub trait ContractFn {
```

stellar-contract-env-host/src/host.rs
```diff
@@ -870,6 +927,28 @@ impl Host {
             _ => Err(HostError::General("not account")),
         })
     }
+
+    #[cfg(feature = "testutils")]
+    pub fn register_test_contract(
```
leighmcculloch, line 932:
```suggestion
    pub fn register_contract(
```

stellar-contract-env-host/src/host.rs
```diff
@@ -177,6 +177,27 @@ struct RollbackPoint {
     objects: usize,
 }
 
+#[cfg(feature = "testutils")]
+pub trait ContractVTableFnTrait {
+    fn call(&self, args: &[RawVal]) -> RawVal;
+
+    fn duplicate(&self) -> ContractVTableFn;
```
leighmcculloch, line 184: What is `duplicate`'s behavior, and how is it different to the behavior of `clone`?

stellar-contract-env-host/src/host.rs
```diff
@@ -751,18 +768,33 @@ impl Host {
...
+                res
+            } else {
+                Some(Err(HostError::General("function not in vtable")))
+            }
+        } else {
+            None
+        }
+
+        #[cfg(not(feature = "testutils"))]
+        None
```
leighmcculloch, line 792: What's the intended future behavior of this fn when not in test mode, it seems odd to always return None.

leighmcculloch (commented):
👏🏻 I definitely think we need something like this.

Could you provide an example of how this will be used? It isn't clear how all components of interface will be used, like the try_ and call_ fns, and I suspect we should at least rename some of these things.

Some comments inline.

stellar-contract-env-host/src/host.rs
leighmcculloch: Actually, why do you need to be able to clone fns? That seems surprising.

stellar-contract-env-host/src/host.rs
```diff
@@ -794,18 +826,43 @@ impl Host {
...
+                    String::from("invalid contract hash"),
+                    ScStatus::HostObjectError(ScHostObjErrorCode::ContractHashWrongLength),
+                )
+            })?;
+            Ok(xdr::Hash(arr))
+        })?;
+
+        self.try_call_vtable(&id, &func, args)
+            .or_else(|| self.try_call_vm(&id, &func, args))
+            .unwrap_or_else(|| unimplemented!())
```
jayz22, line 860: Nitpick: I don't feel the vm call should be "nested" inside the vtable. They are more like parallel relationship instead of one being the fallback of the other (which `or_else` implies). 
If they are nested, then if feels more appropriate the other way around (`try_call_vm.or_else(try_call_vtable)`)? Since VM is the more default/production focused branch.

leighmcculloch: I think once we have native contracts the relationship might exist as such where we execute a vtable loaded contract instead of a vm one if it is present. In tests we are really just presenting the test contract as a native contract.

jonjove: I wouldn't call this nested, it's sequential. That is the appropriate relationship because I can only call one function so there must exist an order in which they are considered. This `or_else` is just a very compact expression of this relationship. As @leighmcculloch points out, if native functions were here then the order would be "test, native, vm".

But I disagree that it should be the other way around. If I've registered a test contract, I expect it to get called so I can debug appropriately.

stellar-contract-env-host/src/host.rs
jonjove: This goes a couple of layers deep. The root cause is that `EnvBase` supports `deep_clone`, which facilitates step-forward/step-backward state checking. For `Host`, this is implemented in terms of `HostImpl::clone`. Therefore `HostImpl` is `Clone`, which means that all parts of `HostImpl` need to be `Clone` as well. Specifically, `RefCell<std::collections::HashMap<Hash, ContractVTable>>` needs to be `Clone`, which is the case only if `ContractVTable` is `Clone`, which is the case only if `ContractVTableFn` is `Clone`. In summary, if we want to change this, then we need to change how step-forward/step-backward state checking works.

So how is `duplicate` different from `clone`? `Box<dyn Clone>` is not [object safe](https://doc.rust-lang.org/reference/items/traits.html#object-safety) because `Clone` uses `Self` outside of the receiver. This `duplicate` machinery, in contrast, is object safe. It works in conjunction with `Box<...>` by constructing a new `Box<...>` by cloning the boxed object.

jonjove: It might be possible to provide a custom implementation of `Clone` for `std::collections::HashMap<Hash, ContractVTable>` using the newtype idiom. I don't know if that will be better, but I will investigate now.

jonjove: It took a bit of finagling, but I think I was able to get this cleaned up. Is it better? Maybe. Did it allow me to find a different design flaw and fix it? Yes.

We can do more work on this interface in the future, but I don't want to block this PR on getting the best possible interface right now. It's important to remember that my goal is for this machinery to end up behind a macro. I'm already partially doing that in rs-stellar-token-contract.

stellar-contract-env-host/src/host.rs
jonjove: I was able to replace with `Fn(&[RawVal]) -> RawVal` (note: _not_ `Fn(RawVal) -> RawVal` because it must accept a variable number of parameters), although it took some work. The trait no longer exists so no name to change.

stellar-contract-env-host/src/host.rs
```diff
@@ -177,6 +177,27 @@ struct RollbackPoint {
...
+#[cfg(feature = "testutils")]
+impl Clone for ContractVTableFn {
+    fn clone(&self) -> Self {
+        self.0.duplicate()
+    }
+}
+
+#[cfg(feature = "testutils")]
+#[derive(Clone)]
+pub struct ContractVTable(pub std::collections::HashMap<Symbol, ContractVTableFn>);
```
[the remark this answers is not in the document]
jonjove: What is a contract in this context, other than a vtable?

stellar-contract-env-host/src/host.rs
jonjove: I reworked this to avoid this quirk.

stellar-contract-env-host/src/host.rs
```diff
@@ -751,18 +768,33 @@ impl Host {
...
+            None
+        }
+
+        #[cfg(not(feature = "testutils"))]
+        None
+    }
+
     #[cfg(feature = "vm")]
-    fn call_n(&self, contract: Object, func: Symbol, args: &[RawVal]) -> Result<RawVal, HostError> {
+    fn call_vm(&self, id: &Hash, func: &Symbol, args: &[RawVal]) -> Result<RawVal, HostError> {
```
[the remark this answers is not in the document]
jonjove: I renamed this `call_wasm_fn` to make it more obvious what it actually does. It's just a helper function for `call_n` to avoid it being ridiculously long.

stellar-contract-env-host/src/host.rs
jonjove: For the record, I removed this code but the new behavior is semantically very similar. The only difference is that in the final case it returns a host error instead of panic with `unimplemented`.

stellar-contract-env-host/src/host.rs
jonjove: I don't want to make this change. This machinery is explicitly for test contracts. Native contracts cannot work this way.

jonjove:
> Could you provide an example of how this will be used?

I have a complete example using this machinery in rs-stellar-token-contract. Once this gets merged, I'll open the small SDK PR. Once that gets merged I'll open the token example.

stellar-contract-env-host/src/host.rs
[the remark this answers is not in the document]
leighmcculloch: I don't think we know the answer to that yet.

If you prefer to keep the narrower focus I suggest `ContractFunctionSet`. I suggest that because if a developer asks me what the ContractVTable is, not knowing what a vtable is, I would tell them it is the functions of the contract.

I mostly want to remove the term vtable from the public API of the SDK. So even if it was only a vtable, I don't think we need to surface the term vtable. This is part of the API that contract developers will use. The term vtable is only common in certain technology ecosystems and it is not meaningful to probably a large variety of developers. I hope we can wrap this inside a macro, but even then I think the term vtable is unnecessary. It's simple to say that this is a `Contract`, or that this is a `ContractFunctionSet`.

stellar-contract-env-host/src/host.rs
leighmcculloch: Is a commit missed from being pushed? This code is still present in the PR.

jonjove: I no longer see this code. Did you refresh?

stellar-contract-env-host/src/host.rs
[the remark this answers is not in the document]
jonjove: Done.

stellar-contract-env-host/src/host.rs
leighmcculloch: 👍🏻 Let's circle back on how we can group and present test only functions in the SDK API.

leighmcculloch (approved):

PR #207, opened 2022-07-13
Add some infrastructure for event logging, including debug events.

graydon:
This is just some preliminary work to have an event buffer and allow writing both protocol-level (TBD) and diagnostic / debug-log (currently useful!) entries to it.

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -64,11 +64,15 @@ macro_rules! call_macro_with_all_host_functions {
 
             mod context "x" {
                 {"_", fn log_value(v:RawVal) -> RawVal }
+
                 /// Get the binary contractID of the contract which invoked the
                 /// running contract. Traps if the running contract was not
                 /// invoked by a contract.
                 {"0", fn get_invoking_contract() -> Object }
                 {"1", fn obj_cmp(a:RawVal, b:RawVal) -> i64 }
+
+                {"2", fn contract_event(v:RawVal) -> RawVal }
+                {"3", fn system_event(v:RawVal) -> RawVal }
             }
 
             mod u64 "u" {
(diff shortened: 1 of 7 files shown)


graydon:
(self-approving since I want it in place for the meeting I'm having with @jayz22 in 5 minutes and also I am going to keep iterating on it this afternoon so will address comments as provided here)

stellar-contract-env-common/src/env.rs
```diff
@@ -64,11 +64,15 @@ macro_rules! call_macro_with_all_host_functions {
...
                 {"_", fn log_value(v:RawVal) -> RawVal }
+
                 /// Get the binary contractID of the contract which invoked the
                 /// running contract. Traps if the running contract was not
                 /// invoked by a contract.
                 {"0", fn get_invoking_contract() -> Object }
                 {"1", fn obj_cmp(a:RawVal, b:RawVal) -> i64 }
+
+                {"2", fn contract_event(v:RawVal) -> RawVal }
+                {"3", fn system_event(v:RawVal) -> RawVal }
```
leighmcculloch, line 75: My understanding from the CAP was that system_event should not be callable, and so it shouldn't be a host fn. It was described as an "internal" fn only.
```
/// Internal host function. Same as $system_event but will use ContractEventType::SYSTEM instead
func $system_event(param $val i64) (result i64)
```
Ref: https://github.com/stellar/stellar-protocol/blob/15d3738523bd577abd711aca2d1582f625d815fc/core/cap-0056.md#host-functions

leighmcculloch (commented):
fyi I believe the Horizon team have asked for these to be called logs, not events. They were called logs, then renamed to events, and I think folks want it to be called logs. Ref: https://discord.com/channels/897514728459468821/989220928384806932 cc @sisuresh

jonjove:
I'm a bit confused. Are we actually going in this direction? How does this relate to the conversation in https://github.com/stellar/rs-stellar-contract-env/issues/205#issuecomment-1183553911?

PR #210, opened 2022-07-13
Generate meta xdr containing interface version

leighmcculloch:
### What

Generate meta containing the interface version.

### Why

So that the SDK can embed this information into WASM files.

Unfortunately generating the XDR in the SDK is somewhat problematic, because we need the literal value to be present at the time the macro runs, and the constant from another crate is not particularly useful unless we have the SDK macro crate depend on the env crate. If we do that, we run into the risk that the SDK macro crate will depend on a different version of the env crate as the SDK itself which would cause the WASM to lie about which env interface it was built with.

Given all that complexity risk, it seems simpler to build the meta within the env crate. After all this meta being built is the meta for the environment only.

The env meta XDR and the macro can be expanded to include other values if we need to add them.

Close https://github.com/stellar/rs-stellar-contract-sdk/issues/142

### Known limitations

[TODO or N/A]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -551,7 +551,7 @@ version = "0.0.0"
 dependencies = [
  "static_assertions",
  "stellar-contract-env-macros",
- "stellar-xdr 0.0.0 (git+https://github.com/stellar/rs-stellar-xdr?rev=0edf9326)",
+ "stellar-xdr 0.0.0 (git+https://github.com/stellar/rs-stellar-xdr?rev=540ed24)",
  "wasmi",
 ]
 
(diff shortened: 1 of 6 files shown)


leighmcculloch:
I don't love making this a proc-macro but there's a series of small things that make this too difficult to do any other way:

- To generate xdr we need std because rs-stellar-xdr doesn't support encoding without it, but std can't be in the main dependency list because common is used in wasm builds.

- Even if rs-stellar-xdr supported encoding without std the encoding would bloat the wasm builds because the encoding would not be executable in a const context at compile time. It would be very difficult to make rs-stellar-xdr encoding within a const context possible as there are so many constraints on what you can do in a const context.

If you look in the git history you'll see I had a simpler proc-macro initially working, but @graydon pointed out that it required a hardcoded byte length which was a hack.

I also considered just hardcoding the xdr value, because the xdr types for this are pretty basic and simple, but alas that also felt like a hack.

If you're wondering why the interface version has to be specified within the macro call itself and can't be referenced from without it, it's because the proc-macro can't see anything outside of it.

Thanks for understanding the imperfection in this PRs complexity.

PR #213, opened 2022-07-14
Change vec_new to take in an optional capacity parameter

jayz22:
### What

Resolves https://github.com/stellar/rs-stellar-contract-env/issues/196

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-common/src/tuple.rs
+++ b/stellar-contract-env-common/src/tuple.rs
@@ -39,7 +39,7 @@ macro_rules! impl_for_tuple {
         {
             fn into_env_val(self, env: &E) -> EnvVal<E, RawVal> {
                 let env = env.clone();
-                let vec = env.vec_new();
+                let vec = env.vec_new(RawVal::from_void());
                 $(let vec = env.vec_push(vec, self.$idx.into_val(&env));)*
                 EnvVal { env, val: vec.to_raw() }
             }
--- a/stellar-contract-env-common/src/tuple.rs
+++ b/stellar-contract-env-common/src/tuple.rs
@@ -39,7 +39,7 @@ macro_rules! impl_for_tuple {
         {
             fn into_env_val(self, env: &E) -> EnvVal<E, RawVal> {
                 let env = env.clone();
-                let vec = env.vec_new(RawVal::from_void());
+                let vec = env.vec_new($count.into());
                 $(let vec = env.vec_push(vec, self.$idx.into_val(&env));)*
                 EnvVal { env, val: vec.to_raw() }
             }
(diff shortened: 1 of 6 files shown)


stellar-contract-env-host/src/test.rs
```diff
@@ -976,6 +998,7 @@ fn invoke_cross_contract() -> Result<(), HostError> {
 }
 
 #[cfg(feature = "vm")]
+#[ignore]
```
jayz22, line 1001: Ignoring this test for now since I need to update the test contract to use the new `vec_new` signature.

stellar-contract-env-host/src/host.rs
```diff
@@ -1122,7 +1122,7 @@ impl CheckedEnv for Host {
 
     fn map_keys(&self, m: Object) -> Result<Object, HostError> {
         self.visit_obj(m, |hm: &HostMap| {
-            let mut vec = self.vec_new()?;
+            let mut vec = self.vec_new(RawVal::from_void())?;
```
leighmcculloch, line 1125: The host knows the size of the map and should use that.

stellar-contract-env-host/src/host.rs
```diff
@@ -1132,15 +1132,25 @@ impl CheckedEnv for Host {
 
     fn map_values(&self, m: Object) -> Result<Object, HostError> {
         self.visit_obj(m, |hm: &HostMap| {
-            let mut vec = self.vec_new()?;
+            let mut vec = self.vec_new(RawVal::from_void())?;
```
leighmcculloch, line 1135: The host knows the size of the map and should use that.

stellar-contract-env-common/src/tuple.rs
```diff
@@ -39,7 +39,7 @@ macro_rules! impl_for_tuple {
         {
             fn into_env_val(self, env: &E) -> EnvVal<E, RawVal> {
                 let env = env.clone();
-                let vec = env.vec_new();
+                let vec = env.vec_new(RawVal::from_void());
```
leighmcculloch, line 42: We should pass in `u32` `$count` here, as it is known. The syntax change I suggest is probably not quite right, because Rust will assume a number without a type will be an i32, but we need we actually need to build a u32 suffixed literal. The easy way to make it a u32 would be to add a u32 suffix to lines 50-61 to the first parameter.
```suggestion
                let vec = env.vec_new($count.into());
```

leighmcculloch (approved):
I left some requests for some small changes inline, but I think it is also fine if these are done as follow up, so approving. Let me know what you want to do.

jayz22:
Cool, I will fix those suggestions, thanks

stellar-contract-env-host/src/host.rs
jayz22: Fixed.

stellar-contract-env-host/src/host.rs
jayz22: Fixed.

stellar-contract-env-host/src/test.rs
leighmcculloch: Do you want me to hold merging this, or good to merge and this is happening in a follow up?

leighmcculloch (approved):

stellar-contract-env-host/src/test.rs
jayz22: Let's merge this. I will fix it as a follow up. The original contract code that generated this test isn't in the sdk repo yet.

PR #216, opened 2022-07-14
Fix map prev and next key to do strict comparison

jayz22:
### What

Resolves https://github.com/stellar/rs-stellar-contract-env/issues/180 

### Why

Previously the `map_prev_key` , `map_next_key` replicates the im_rc's behavior: if the given key exists, returns itself, which is not useful for implementing iterators. 

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-host/src/budget.rs
+++ b/stellar-contract-env-host/src/budget.rs
@@ -15,7 +15,7 @@ pub enum CostType {
 
 // TODO: add XDR support for iterating over all the elements of an enum
 impl CostType {
-    fn variants() -> std::slice::Iter<'static, CostType> {
+    pub fn variants() -> std::slice::Iter<'static, CostType> {
         static VARIANTS: &'static [CostType] = &[
             CostType::HostMapAllocMap,
             CostType::HostMapAllocCell,
@@ -23,6 +23,7 @@ impl CostType {
             CostType::HostVecAllocCell,
             CostType::WasmInsnExec,
             CostType::WasmMemAlloc,
+            CostType::HostEventDebug,
         ];
         VARIANTS.iter()
     }
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -1090,6 +1090,12 @@ impl CheckedEnv for Host {
     fn map_prev_key(&self, m: Object, k: RawVal) -> Result<RawVal, HostError> {
         let k = self.associate_raw_val(k);
         self.visit_obj(m, |hm: &HostMap| {
+            // OrdMap's `get_prev`/`get_next` return the previous/next key if the input key is not found. 
+            // Otherwise it returns the input key. Therefore, if `get_prev`/`get_next` returns the same
+            // key, we will clone the map, delete the input key, and call `get_prev`/`get_next` again. 
+            // Note on performance: OrdMap does "lazy cloning", which only happens if data is modified, 
+            // and we are only modifying one entry. The cloned object will be thrown away in the end
+            // so there is no cost associated with host object creation and allocation. 
             if let Some((pk, pv)) = hm.get_prev(&k) {
                 if *pk != k {
                     Ok(pk.to_raw())
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -1090,12 +1090,12 @@ impl CheckedEnv for Host {
     fn map_prev_key(&self, m: Object, k: RawVal) -> Result<RawVal, HostError> {
         let k = self.associate_raw_val(k);
         self.visit_obj(m, |hm: &HostMap| {
-            // OrdMap's `get_prev`/`get_next` return the previous/next key if the input key is not found. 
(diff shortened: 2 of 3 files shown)


stellar-contract-env-host/src/host.rs
```diff
@@ -1092,17 +1092,51 @@ impl CheckedEnv for Host {
...
+            if let Some((pk, pv)) = hm.get_next(&k) {
+                if *pk != k {
+                    Ok(pk.to_raw())
+                } else {
+                    if let Some((pk2, pv2)) = hm
+                        .clone()
+                        .extract(pk)
+                        .ok_or_else(|| HostError::General("key not exist"))?
+                        .1
+                        .get_next(pk)
```
leighmcculloch, line 1130: How do these lines of code work? I'm struggling to see what is being extracted. This block of code might benefit from a comment I think.

stellar-contract-env-host/src/test.rs
```diff
@@ -124,6 +124,82 @@ fn map_put_has_and_get() -> Result<(), HostError> {
...
 }
 
+/// map test
+#[test]
+fn map_prev_and_next() -> Result<(), HostError> {
+    let host = Host::default();
+    let scmap: ScMap = vec![
+        ScMapEntry {
+            key: ScVal::U32(1),
+            val: ScVal::U32(2),
```
leighmcculloch, line 134: Will this work for more complex key types, such as a key which is a Vec or a Map?

stellar-contract-env-host/src/host.rs
jayz22: I added a comment. Let me know if it still needs more explanation.

stellar-contract-env-host/src/test.rs
jayz22: Yes, these work under the `cmp` rules for `RawVal`. 
I've added a new more complex example in the test.

stellar-contract-env-host/src/host.rs
```diff
@@ -1089,17 +1089,51 @@ impl CheckedEnv for Host {
...
-            Some((pk, pv)) => Ok(pk.to_raw()),
-            None => Ok(UNKNOWN_ERROR.to_raw()), //FIXME: replace with the actual status code
+        self.visit_obj(m, |hm: &HostMap| {
+            if let Some((pk, pv)) = hm.get_next(&k) {
+                if *pk != k {
+                    Ok(pk.to_raw())
+                } else {
+                    if let Some((pk2, pv2)) = hm
+                        .clone()
+                        .extract(pk) // removes (pk, pv) and returns an Option<(pv, updated_map)>
```
leighmcculloch, line 1124: Oh interesting, so we're removing the key from the OrdMap, so that we can find the next key? That will result in a copy of the OrdMap is that right? 🤔 The main reason we're fixing these functions, besides the fact their broken, is so we can use them in an SDK to avoid making a copy of the map, because at the moment the SDK just copies the host object and removes the key itself. So it feels like we're shifting the existing fix upstream, which is better because maybe we can find a better fix someday host side, but is there not another way we can get the next without a copy?

jayz22: Unfortunately I couldn't find a different way to do it without cloning the OrdMap. One way I was trying to do it was to get the iterator on the input key in the map then call `next` (or `next_back`) on it. However, there are very few methods that return iterator to map entry (which seems very different from C++ containers and iterators). 
Anyways, I believe the performance should not be a big concern. The im-rc does "lazy cloning" which only happens when the data is modified, and our worse case here is deleting one entry, which means cloning is very limited (that node plus the parent nodes). The big difference from doing it on the SDK side is no new host object will be created. The overall performance should still on the order of `log(n)` and should be very similar to (if not slightly better than) a `map_del` (since no new host object is created).
I'm planning to wire up metering on the host functions soon, so hopefully we will have a more concrete answer to the performance soon. 
(I've also added a much longer comment in the code explaining some of the rationale)

stellar-contract-env-host/src/budget.rs
```diff
@@ -15,14 +15,15 @@ pub enum CostType {
...
-    fn variants() -> std::slice::Iter<'static, CostType> {
+    pub fn variants() -> std::slice::Iter<'static, CostType> {
         static VARIANTS: &'static [CostType] = &[
             CostType::HostMapAllocMap,
             CostType::HostMapAllocCell,
             CostType::HostVecAllocVec,
             CostType::HostVecAllocCell,
             CostType::WasmInsnExec,
             CostType::WasmMemAlloc,
+            CostType::HostEventDebug,
```
leighmcculloch, line 26: It's so odd that this part of the diff is showing up in the PR. GitHub must have a cache issue or a bug. The diff doesn't show this when viewed in git.

stellar-contract-env-host/src/host.rs
leighmcculloch: Makes sense. Also, even if there was no immediate to doing this in the host, if we ever find a better way to solve this in the future at least having it in the host means we can change it seamlessly. :tada:

leighmcculloch (approved):

PR #217, opened 2022-07-14
Add an iterator for the CostType enum

leighmcculloch:
### What

Add an iterator for the CostType enum.

### Why

So that tooling can iterate over all cost types and output details about them, wihtout having to hardcode knowledge of a fixed set of cost types.

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-host/src/budget.rs
+++ b/stellar-contract-env-host/src/budget.rs
@@ -1,4 +1,4 @@
-use strum::EnumIter;
+use strum::{EnumIter, IntoEnumIterator};
 
 use crate::HostError;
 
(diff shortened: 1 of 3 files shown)


stellar-contract-env-host/src/budget.rs
```diff
@@ -13,21 +15,6 @@ pub enum CostType {
     HostEventDebug = 6,
 }
 
-// TODO: add XDR support for iterating over all the elements of an enum
```
leighmcculloch, line 16: Ah, yes, I can add this. https://github.com/stellar/xdrgen/issues/120

leighmcculloch:
Will close and fix the unique case, and then assume this will be moved to XDR instead.

PR #220, opened 2022-07-15
Change HostError to just a Status, move diagnostic info to DebugEvent.

graydon:
This is a fairly large (not terribly deep, but important and systematic) change that implements my best attempt at addressing the conclusions of our conversation in https://github.com/stellar/rs-stellar-contract-env/issues/205

If it's not _totally_ objectionable, I'd like to ask for a fast review and landing on: it'll bitrot almost instantly if anyone lands anything else, and it touches almost every function in host and test.

What it does:

  - Splits the diagnostic-and-error-handling functionality in HostError into two:
    - `HostError` retains a pure `Status` value (not `ScStatus`, though they're mostly interchangeable) and is used for _error handling_ only
    - Diagnostic functionality is transferred to `DebugEvent` which can contain a small vec of `RawVal`s (possibly including a `Status`) as well as a `&'static str`. Small and cheap, so we can accumulate them in the `Events` buffer without a lot of cost.
    - Cases where we want to both record diagnostic information _and_ produce a handleable error are done by constructing a `DebugError`, which just wraps a `DebugEvent` and `Status` temporarily (at the error-origination site) while reporting, and then decomposes them back to the associated pair.
    - Adds a few `From<foo> for DebugError` impls to illustrate that you can log finer-grained _diagnostic_ information from the substructure of a type and then return a coarser-grained Status.
  - Adds a few helpers on `Host` for creating, recording, and converting these types to `HostError`.
  - Modifies nearly every site that produces errors to use these types.
  - Adds some new `From` impls to `Status` so that it can be created from any of the `ScStatus` _codes_ (each protocol-level code uniquely determines its type, so there's no real need for the user to write out both).
  - Adds a few helpers to the testsuite to make it less chatty (eg. creation of new vectors and binaries)
  - Rewrites all the testsuite uses of `assert_matches` since that's no longer a good way to do a match on `HostErrors`. Gotta project a code into `Status` and then do a match on a `Result`.

Why is this good?
  - Simplifies decisions around errors / avoids an inherent tradeoff. No longer any tension around whether to enrich `HostError`'s structure (for diagnostic sake) or reduce (for ease of handling / conversion to `Status`). Can pursue both separately.
  - Simplifies maintenance since diagnostic messages can _never_ influence error handling, and the structured type definition for any handleable error is defined in exactly one place: the XDR for ScStatus and its guts.
  - Makes the code that generates errors a little less chatty.
  - Gives a feasible approach for capturing much richer diagnostic info like "which value was not convertable" (which is something @jayz22 is blocked on for #181) -- can call something like `self.err(DebugError::new(ScHostFnErrorCode::InputArgsWrongType).msg("argument foo").arg(bad_raw_val))`

Why is it bad?
  - A bunch of tests were matching on `::General(...)` string values before. This is gone since diagnostic strings are literally not present in errors coming out of the system. These tests are therefore a bit unspecific now and ought to be revised with new error codes at some reasonable level of specificity (they're mostly bignum tests, so maybe some number of new bignum error codes need to be added to Status, I'm not sure)
  - Doesn't do the plausibly harder work of actually reorganizing the `Status` types yet. It just clears the way to make that task easier to think about for @jayz22 and anyone else who wants to dig into "how to categorize errors for handling cases at a reasonable granularity".

### DIFF
--- a/stellar-contract-env-common/src/raw_val.rs
+++ b/stellar-contract-env-common/src/raw_val.rs
@@ -45,6 +45,12 @@ pub enum Tag {
 #[derive(Copy, Clone)]
 pub struct RawVal(u64);
 
+impl Default for RawVal {
+    fn default() -> Self {
+        Self::from_void()
+    }
+}
+
 impl AsRef<RawVal> for RawVal {
     fn as_ref(&self) -> &RawVal {
         self
--- a/stellar-contract-env-host/src/events.rs
+++ b/stellar-contract-env-host/src/events.rs
@@ -1,39 +1,227 @@
-#![allow(dead_code)]
+use log::debug;
+use stellar_contract_env_common::{
+    xdr::{self},
+    RawVal, Status,
+};
 
-use arrayvec::ArrayVec;
-use stellar_contract_env_common::RawVal;
+use tinyvec::TinyVec;
+
+#[cfg(feature = "vm")]
+use crate::HostError;
+#[cfg(feature = "vm")]
+use stellar_contract_env_common::xdr::{ScUnknownErrorCode, ScVmErrorCode};
 
 // TODO: update this when ContractEvent shows up in the XDR defns.
 // TODO: optimize storage on this to use pools / bumpalo / etc.
 #[derive(Clone, Debug)]
-pub(crate) enum HostEvent {
+pub enum HostEvent {
+    #[allow(dead_code)]
     Contract(/*ContractEvent*/),
-    // Same size as LargeDebug but doesn't chew up a heap allocation.
-    SmallDebug {
-        msg: &'static str,
-        // ArrayVec<RawVal, 2> is 3 words: RawVal, RawVal, len.
-        args: ArrayVec<RawVal, 2>,
-    },
-    // Arbitrary set of k/v arguments.
-    LargeDebug {
-        msg: &'static str,
-        // Vec is 3 words: ptr, len, capacity.
-        args: Vec<RawVal>,
-    },
+    Debug(DebugEvent),
 }
 
 #[derive(Clone, Debug, Default)]
 pub struct Events(Vec<HostEvent>);
 
 impl Events {
-    pub(crate) fn add_debug_event(&mut self, msg: &'static str, in_args: &[RawVal]) {
-        let debug = if in_args.len() <= 2 {
-            let args: ArrayVec<RawVal, 2> = ArrayVec::from_iter(in_args.iter().cloned());
-            HostEvent::SmallDebug { msg, args }
-        } else {
-            let args = in_args.into();
-            HostEvent::LargeDebug { msg, args }
+    // Records the smallest variant of a debug HostEvent it can, returning the size of the
+    // in_args slice (for charging to a budget).
+    pub fn record_debug_event(&mut self, de: DebugEvent) -> u64 {
+        let len = de.args.len();
+        self.0.push(HostEvent::Debug(de));
+        len as u64
+    }
+
+    pub fn dump_to_debug_log(&self) {
+        for e in self.0.iter() {
+            match e {
+                HostEvent::Contract() => debug!("Contract event: <TBD>"),
+                HostEvent::Debug(e) => debug!("Debug event: {:?}", e),
+            }
+        }
+    }
+}
+
+/// A cheap record type to store in the events buffer for diagnostic reporting
+/// when something goes wrong. Should cost very little even when enabled. See
+/// [host::debug_event] for normal use.
+#[derive(Clone, Debug)]
+pub struct DebugEvent {
+    pub msg: &'static str,
+    pub args: TinyVec<[RawVal; 2]>,
+}
+
+impl DebugEvent {
+    pub fn new() -> Self {
+        Self {
+            msg: "",
+            args: Default::default(),
+        }
+    }
+
+    pub fn msg(mut self, msg: &'static str) -> Self {
+        self.msg = msg;
+        self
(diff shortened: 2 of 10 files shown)


stellar-contract-env-common/src/raw_val.rs
```diff
@@ -42,7 +42,7 @@ pub enum Tag {
 }
 
 #[repr(transparent)]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default)]
 pub struct RawVal(u64);
```
leighmcculloch, line 46: A default value for RawVal seems suspicious. The default value will be a zero u63. It might be advantageous to make it a void. Could we provide a Default impl that makes the default void?

leighmcculloch (commented):
Nice. This is pretty large, so I doubt I'm digesting it all, but I think I get the gist of it from the Host side, and from the Guest side for errors generated by the Host, and it looks good. 

The area I don't feel like I have a good grasp on is what this means for errors on the Guest side for Guest for errors generated by code running on the Guest side. Two questions to help me understand that:

- What will a conversion on the Guest side look like? What will the error be?

- How do contract developers use this to define their own errors?

stellar-contract-env-common/src/raw_val.rs
graydon: Sure I guess. This is only to support `TinyVec` but .. yeah I can set it to void.

stellar-contract-env-host/src/events.rs
```diff
@@ -1,39 +1,227 @@
...
+}
+
+impl DebugError {
+    pub fn new<T>(status: T) -> Self
+    where
+        Status: From<T>,
+    {
+        let status: Status = status.into();
+        Self {
+            event: DebugEvent::new().msg("status").arg::<Status>(status.into()),
```
jayz22, line 93: This is cool, that you are chaining these calls, each of which returns a `self`. 
Just curious, is there any advantage of doing it this way versus constructing the object by passing in fields directly? 
`event: DebugEvent {msg: "status", args: [status]}`?

stellar-contract-env-host/src/events.rs
```diff
@@ -1,39 +1,227 @@
...
+    pub fn msg(mut self, msg: &'static str) -> Self {
+        self.msg = msg;
+        self
+    }
+
+    pub fn arg<T>(mut self, arg: T) -> Self
+    where
+        RawVal: From<T>,
+    {
+        self.args.push(arg.into());
```
jayz22, line 71: This call may panic if overflow happens, should we prevent that from happening by doing the condition check?

stellar-contract-env-host/src/events.rs
graydon: Just brevity and flexibility at the call site (i.e. you can omit msg or args, or only pass 1 arg, with less noise). This pattern ("builders") is fairly common in Rust because Rust does not support keyword arguments, optional/default arguments, or argument-list overloading of constructors. It's the only real way to ship "flexible construction".

stellar-contract-env-host/src/events.rs
graydon: TinyVec is not ArrayVec -- it automatically overflows to the heap; the panic here would only happen if we literally run out of memory, which is a ubiquitous error we don't generally guard against.

stellar-contract-env-common/src/raw_val.rs
graydon: (fixed)

stellar-contract-env-host/src/events.rs
leighmcculloch: I think I'd rather the simpler `DebugEvent { msg: "...", args: tiny_vec!(1, 2, 3) }` as well. It actually appears to be more brief than the longer syntax here.

I'm not opposed to the builder pattern but it is usually beneficial when a type's constraints can't be enforced naturally by the type system, which isn't the case here.

leighmcculloch (approved):
I think this is good and we should merge it.

>@leighmcculloch Re your questions: when a guest generates an error in their own code, it never touches the host and thus never has any representation in the host's error repertoire. A guest can generate a status to return to their caller if they want -- and we can reserve a part of the type-space for them to do so -- and a guest can also call the log function to write a diagnostic message. But in general their errors are just data bouncing around inside their VM, not related to this any more than their own datatypes are.

When I was referring to the guest I meant the guest in tests, where we do want to surface back traces and get all that rich diagnostic information. I think we still need a story for how that will surface, and it would be great if the same mechanism can be used.

jayz22 (approved):
I agree we should merge this. I've created a separate [PR](https://github.com/stellar/rs-stellar-contract-env/pull/224/files) onto this branch with some minor fixes. But don't want to hold this one back.

stellar-contract-env-host/src/events.rs
```diff
@@ -1,39 +1,227 @@
...
-            HostEvent::LargeDebug { msg, args }
+    // Records the smallest variant of a debug HostEvent it can, returning the size of the
+    // in_args slice (for charging to a budget).
+    pub fn record_debug_event(&mut self, de: DebugEvent) -> u64 {
+        let len = de.args.len();
+        self.0.push(HostEvent::Debug(de));
+        len as u64
+    }
+
+    pub fn dump_to_debug_log(&self) {
```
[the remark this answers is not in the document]
graydon: By the embedding environment and/or unit tests. We'll probably want several different ways to get the debug events back out of the system, this is just a trivial one to call from (eg.) the command line.

Issue #215, opened 2022-07-14
Error if contract does not include env meta interface version


leighmcculloch:

graydon:
@leighmcculloch I think this is done now as of #228

PR #228, opened 2022-07-16
Check interface version + test cleanup

graydon:
I wanted this to only be one thing, but it wound up having to be 2, and then 2 led to .. a few more things:

  - Primary purpose: check WASM meta section has the right interface number (i.e. implement #215)
    - Secondary task: improve HostError reporting a lot to make it clear to user when interface number is wrong/missing.
      - Along the way: reorganize HostError into a submodule of host.
    - Secondary task: update WASM blobs in tests to actually have interface numbers.
      - Along the way: reorganize tests into a set of separate submodules.
      - Along the way: also move WASM blobs to files bin-copied in rather than using textual arrays.
      - Along the way: fixed errors in the memcpy test, made a few changes to the SDK (will file those shortly).

Phew! I hope that is all acceptable. I left a couple of the wasm-blob-using tests disabled since @jayz22 has not committed their sources anywhere so they can't be fixed up / regenerated / moved to external files yet. But we can pick that up next week.


stellar-contract-env-host/src/test/examples.rs
```diff
@@ -0,0 +1 @@
+
```
leighmcculloch, line 1: 👀 What's this do?

stellar-contract-env-host/src/test/invocation.rs
```diff
@@ -0,0 +1,254 @@
+use super::wasm_examples::ADD_I32;
```
leighmcculloch, line 1: 👀 Nice!

leighmcculloch (approved):
If you wanted I think you could copy the examples into this repo. Realistically those examples are probably going to go away in the SDK and be replaced by some simple examples. I mean add will probably stay, but the others will probably go.

stellar-contract-env-host/src/test/examples.rs
graydon: Oh, that's an edit artifact left over from when I was trying to decide what to name the submodule. Snuck in, will remove.

graydon:
> If you wanted I think you could copy the examples into this repo. Realistically those examples are probably going to go away in the SDK and be replaced by some simple examples. I mean add will probably stay, but the others will probably go.

Yeah this is probably not the last word on how these things get organized, I just wanted to take a step in the right direction.

jayz22:
> I left a couple of the wasm-blob-using tests disabled since @jayz22 has not committed their sources anywhere so they can't be fixed up / regenerated / moved to external files yet. But we can pick that up next week.

Here is the fix: https://github.com/stellar/rs-stellar-contract-env/pull/230

PR #235, opened 2022-07-19
Add a little more infrastructure for building ConversionErrors

graydon:
A little promised code from meeting earlier today, sorry for delay @jayz22 !

### DIFF
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -234,6 +234,17 @@ impl Host {
         self.err(DebugError::new(status).msg(msg))
     }
 
+    pub(crate) fn err_convert<T>(&self, v: RawVal) -> HostError {
+        if let Err(e) = self.debug_event(DebugEvent::new().msg("type conversion failed")) {
+            return e;
+        }
+        if let Err(e) = self.debug_event(DebugEvent::new().msg(core::any::type_name::<T>()).arg(v))
+        {
+            return e;
+        }
+        ConversionError {}.into()
+    }
+
     /// Given a result carrying some error type that can be converted to a
     /// DebugStatus, calls self.err with it when there's an error. Returns a
     /// result over HostError.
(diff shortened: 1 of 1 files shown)


stellar-contract-env-host/src/host.rs
```diff
@@ -234,6 +234,17 @@ impl Host {
         self.err(DebugError::new(status).msg(msg))
     }
 
+    pub(crate) fn err_convert<T>(&self, v: RawVal) -> HostError {
+        if let Err(e) = self.debug_event(DebugEvent::new().msg("type conversion failed")) {
+            return e;
+        }
+        if let Err(e) = self.debug_event(DebugEvent::new().msg(core::any::type_name::<T>()).arg(v))
```
leighmcculloch, line 241: I think we can combine these two debug events.

Can we format a msg to include the typename? The typename should be known at compile time, and this fn is generic and bound to the type at compile time, so it should be possible to compose the `type conversion into T failed` string at compile time.

(We'd need to use something like the [const_format](https://crates.io/crates/const_format) crate.)

If we can't do that for some reason, can we truncate and convert the T typename into a Symbol and include it as an arg?

jayz22: The [core](https://doc.rust-lang.org/beta/core/index.html)::[any](https://doc.rust-lang.org/beta/core/any/index.html)::[type_name](https://doc.rust-lang.org/beta/core/any/fn.type_name.html#) returns the typename as a string slice. 
I can combine these two events and incorporate any other suggestions you may have.

jayz22 (approved):

stellar-contract-env-host/src/host.rs
jayz22: This PR is to just check in Graydon's changes during today's meeting so that I can pick it up and continue on.

graydon: Oh neat, I hadn't seen const_format. Maybe! I will also be fleshing out support for multiple static strings (host and guest) as args beyond the initial msg, so this is very much a stopgap, not the final version.

jayz22: Hmm, I couldn't figure out how to do it with one debug_event. 
`const_format` (e.g. [concatcp!](https://docs.rs/const_format/latest/const_format/macro.concatcp.html) ) doesn't help here because the inputs must be const. 
Is there any way to get the compile-time type names?

leighmcculloch: Ah, so the `core::any::typename` is a `const fn`, but only in nightly behind the `const_type_name` feature. Doh!

leighmcculloch (approved):

jayz22:
Closing this as I've incorporated the changes into my follow-up work.

PR #238, opened 2022-07-20
Update test files

jayz22:
### What

Updating wasm files for test. 

### Why

Testing contract code was out of sync again. https://github.com/stellar/rs-stellar-contract-sdk/pull/133

### Known limitations

[TODO or N/A]


jonjove:
@jayz22 why did `call` get so much bigger? Did you, for example, `wasm-strip` the binaries?

jayz22:
> @jayz22 why did `call` get so much bigger? Did you, for example, `wasm-strip` the binaries?

The only difference is I switched my version of `call` with yours in the sdk ([diff](https://github.com/stellar/rs-stellar-contract-sdk/pull/133/commits/79046e2e6f84ec488300bfacb61d85d3c2e0e24f)), which requires each arg to be an `EnvVal` and return type to be spelled out.

PR #241, opened 2022-07-21
Fix frame length check in get_invoking_contract

jayz22:
### What

Frame length should be `>=2`. Added a comment. 

### Why

### Known limitations

[TODO or N/A]


jonjove:
Thanks for catching this!

Issue #229, opened 2022-07-16
Log / diagnostics from guest memory static strings


graydon:
Given that any `&'static str` in guest linear memory _has static address_, we can write a variant of the `log` host function that is typed as `&'static str` on the SDK side and passes a `(u32,u32)` pair denoting a linear memory span to the host. The host can then store that pair plus a refcount to the wasm module / vm in the debug log, and materialize the string when dumping the log just by fishing it out of the wasm module. This should let users log arbitrary static strings, including such delights as file names, module names, function names. Should be able to build a useful contract_trace!() macro out of it.

graydon:
Postponing for now, we got decent-enough support for format-string-based debug logging in local-testing mode in the functions added to `EnvBase`in #243 and the idea with that is that users _probably_ don't want their deployed contracts carrying static strings at all, so the `Guest` impl of `EnvBase` implements those functions as no-ops, which means there's no need to plumb through guest linear memory pairs referring to static strings.

Can revisit if users wind up asking why their format-string-based debug log messages don't show up in deployed WASM contracts, only local testing mode.

graydon:
Experimented with infrastructure for this in #584 and determined that it's .. really quite a lot of effort for something we have no real signal that anyone will want yet. It anticipates "high volume debug logging in production contracts" and .. there's not a lot of reason to assume that will actually happen, given the likelihood of size-optimization and such. Sticking with local-testing-only mode as mentioned above. Reopen if anyone complains in the future, but I'm betting nobody ever will.

Issue #205, opened 2022-07-13
Revise status to have more structure, pack bits differently.


graydon:
Currently the `Status` type reuses a major/minor structure that we also use for object index/type pairs: a 28 bit status-type and a 32 bit status-code.

This is plausible, but it's also not necessarily the _best_ representation of Status values, and we've been feeling a lot of pressure to pack more information into Status values in at least a few places:

  - Conversion errors want to pack at least a from-type and to-type into a status value.
  - Debug / trace logging would benefit from packing some or all of a file name, line number and user-provided static string into a status.
  - User-specified abort Statuses (as in #146) could also benefit from these things.

I'm thinking it might be nice to decompose the 60 bits available to Status values a little differently. For example: assuming WASM binaries in the current design are capped at 64KB, we can assume any linear-memory address (such as those for string constants) fits in 16 bits. Even if we are being more future-proof-y and saying 1MB / 20 bits, we could easily do a 4-field split: 4 bit type, 16 bit code-or-line-number, and 2 20-bit string constants (for filename and user message). Or keep the type space a bit bigger, stick with 64k for addresses, and go 12-16-16-16. Or go with 256KB / 18 bit addresses, and do 8-16-18-18.

We might even be able to figure out a way to build a 1-level indirection table for string constants -- we really only care about the start positions, not all the sub-positions within them -- so we can denote the start positions by table index rather than resolved address, and drop to (say) 10 bit indexes -- not likely anyone will have a contract with more than 1024 files.

(We could also do something even more elaborate and add a status _object_ type, maybe that we overflow to, but I don't know if that'd be any better.)

graydon:
(It is also possible all this is a bit overkill, and we could get away with something like what I did in the earlier prototype: say, a function that pushes a filename / line / user message _into the debug log buffer_ and then fails more-simply with today's 28/32-split `Status` value. The interested user can turn on logging to see the nature of the failure; the less-interested user can just accept the less-informative `Status`)

graydon:
cc @jayz22, @sisuresh and @jonjove

graydon:
@jonjove asked for a more "goals"-oriented approach, so let's go through those (this comment is being composed in a meeting with multiple parties):

## Questions

  1. Do we want users to handle and recover from errors, or just panic?
      - Usually panic but occasionally handle.
  2. In which circumstances do we want to trap in host vs. return status?
      - Return status on failure by default. This is _different_ from what we currently do.
[...]

Issue #179, opened 2022-07-07
Properly handle native (feature != "VM") case for linear memory hostfns


jayz22:
https://github.com/stellar/rs-stellar-contract-env/pull/147/#discussion_r916089138

jayz22:
Completed by https://github.com/stellar/rs-stellar-contract-env/pull/243. 
Also see https://github.com/stellar/rs-stellar-contract-env/pull/256#issuecomment-1195887904 on why `unimplemented!()` is the correct behavior for `binary_{copy_to, copy_from, new_from}_linear_memory` functions in the native mode.

PR #243, opened 2022-07-21
Implement contract memory access and debug formatting

graydon:
This addresses two sets of related problems we've been discussing recently:

  - How to get contract-memory slices into binaries when `Env=Host` (call this "native" or "test contract", however you like): #179
  - How to get richer debug logging from contracts _at least_ (perhaps only) when building `Env=Host` (multiple args, format strings, static string args): #229, #181, #205 

The common problem / issue here (and what took so long, I tried a lot of different unsatisfying approaches) is "what's the safest and least-terrible way of receiving pointers-into-contract-memory on the host when `Env=Host`". We know what to do when `Env=Guest` -- a contract pointer is a linear memory address, a u32 -- but when `Env=Host` it means we're sharing a process with the host and we might find it ugly, unsafe or even unsound to try to transmit pointers through the narrow vocabulary of `Env`, which only knows about `u64` and `RawVal`.

The answer I came up with eventually is "provide the pointer-passing interfaces in `EnvBase`, have them take lifetime-qualified Rust slices, and _decay those to `Env` calls_ only in the `Env=Guest` case, and even then optionally (eg. the logging interface decays them to no-ops because we want `Env=Guest` builds to be mimimum-sized)". Then make the pointer-receiving paths on `Env` fail when called from a non-VM frame (i.e. native/test -- "definitely not `Guest`"), as they do today.

I think this sorts out how to move forwards with such questions, and more concretely it gives a pattern (as well as a few initial functions) for debug logging functions that take multiple args from _either_ guest or host side, formatting them lazily on print-out. I _think_ these should handle the error-reporting cases people are stuck on (eg. the rich type-conversion errors and such).

### DIFF
--- a/stellar-contract-env-common/src/env.rs
+++ b/stellar-contract-env-common/src/env.rs
@@ -10,6 +10,49 @@ pub trait EnvBase: Sized + Clone {
 
     // Used to clone an environment deeply, not just a handle to it.
     fn deep_clone(&self) -> Self;
+
+    // Helpers for methods that wish to pass Rust lifetime-qualified _slices_
+    // into the environment. These are _not_ done via Env trait methods to avoid
+    // the need to convert, and thus trust (or validate) "raw numbers" coming
+    // through that interface as "potentially pointers in the same address space
+    // as the host". This is a bit of a defense-in-depth approach as we _could_
+    // just accept "numbers as pointers in our address space" on a codepath that
+    // is sure its input is coming from a "trusted" contract, and arrange enough
+    // other static safety checks elsewhere in the calling path (eg. in the SDK)
+    // to ensure that "all callers are trusted" .. but we want to minimize the
+    // chance of future maintainers accidentally violating such an invariant,
+    // since getting it wrong would let guest code violate memory safety. So the
+    // _only_ interface to passing contract pointers to the host is going to be
+    // in EnvBase, not Env, and as a bonus we get lifetime checking for free.
+    fn binary_copy_from_slice(&self, b: Object, b_pos: RawVal, mem: &[u8]) -> Object;
+    fn binary_copy_to_slice(&self, b: Object, b_pos: RawVal, mem: &mut [u8]);
+    fn binary_new_from_slice(&self, mem: &[u8]) -> Object;
+
+    // As with the binary functions above, these take _slices_ with definite
+    // lifetimes. The first slice is interpreted as a (very restricted)
+    // format-string -- containing literal text interspersed with some number of
+    // `{}` markers which must match the number of other args passed -- with
+    // actual formatting delayed until someone asks to see the event (which may
+    // never happen). Other args may be static strings, [RawVal]s, or a mix.
+    //
+    // When the SDK is built with Env = Host, both the format string slice and
+    // all static string slice args (and any [RawVal] args) will be passed
+    // through into the debug-event subsystem of the host and _stored_
+    // unformatted in the debug buffer, until/unless someone dumps some portion
+    // of that buffer out. They are therefore quite cheap -- just pushing static
+    // pointers and numbers into the debug buffer -- and can be called fairly
+    // ubiquitously to provide details on any interesting diagnostic events
+    // and/or errors that occur in either SDK or contract code.
+    //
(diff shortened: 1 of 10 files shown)


graydon:
I'll add a few more debug-message-logging functions later on, just wanted to post this for an initial glance now.

stellar-contract-env-common/src/env.rs
```diff
@@ -63,6 +106,8 @@ macro_rules! call_macro_with_all_host_functions {
             // argument list of comma-separated arg:type pairs
 
             mod context "x" {
+                // This one variant of logging does not take a format string and
+                // is live in both Env=Guest and Env=Host configurations.
                 {"_", fn log_value(v:RawVal) -> RawVal }
```
leighmcculloch, line 111: Should we get rid of this one?

stellar-contract-env-guest/src/guest.rs
```diff
@@ -4,11 +4,60 @@
...
 
 // In guest code the environment is global/implicit, so is represented as a unit struct.
 #[derive(Copy, Clone, Default)]
 pub struct Guest;
 
+// The Guest struct is only meaningful when compiling for the WASM target. All
+// these fns should not be called at all because the SDK's choice of Env should be
+// Host for a non-WASM build.
+#[cfg(not(target_family = "wasm"))]
+impl EnvBase for Guest {
```
leighmcculloch, line 18: ❗ We only build the guest crate as a whole for wasm. When will this crate be used in non-wasm?

stellar-contract-env-host/src/host.rs
```diff
@@ -885,6 +921,81 @@ impl EnvBase for Host {
...
+        // This is only called from native contracts, either when testing or
+        // when the contract is otherwise linked into the same address space as
+        // us. We therefore access the memory we were passed directly.
+        //
+        // This is also why we _panic_ on errors in here, rather than attempting
+        // to return a recoverable error code: native contracts that call this
+        // function do so through APIs that _should_ never pass bad data.
+        //
+        // TODO: we may revisit this choice of panicing in the future, depending
+        // on how we choose to try to contain panics-caused-by-native-contracts.
```
leighmcculloch, line 935: +1 I think we should revisit this.

leighmcculloch (commented):
Looks good to me. I agree with the TODO you added that we should revisit the panicking. We should at least discuss it in greater depth when more people are available to do so. I have one concern marked with the ❗, and I think my questions in that comment probably indicate I have a poor understanding of how some of this works.

stellar-contract-env-guest/src/guest.rs
graydon: Should never be. This is a just-in-case. If you like I can remove it.

graydon: (Note there are static asserts that we're on a 32-bit platform in the other case, so like the crate won't build for most hosts anyways, I don't know if maybe we want it to build-but-be-inert in some other case?)

leighmcculloch: I think it's fine to remove this. The assumption is the entire crate is wasm32 only. Even in the SDK we never import it for non-wasm builds.

leighmcculloch: Oh, although maybe the crate builds for non-wasm in local builds of this workspace. I don't mind if it's included.

leighmcculloch (approved):

jayz22 (approved):
This is great, thanks for doing! I'll need more time to digest the nuances but approving first.

stellar-contract-env-host/src/events.rs
```diff
@@ -32,56 +34,85 @@ impl Events {
...
+            None => {
+                for arg in self.args.iter() {
+                    write!(f, "{}", arg)?;
                 }
-                first = false;
-                write!(f, "{:?}", rv)?;
+                Ok(())
+            }
+            Some(fmt) => {
+                let args = dyn_fmt::Arguments::new(fmt, self.args.as_slice());
```
jayz22, line 95: This is really cool, didn't know it!

PR #248, opened 2022-07-22
Impl std::error::Error for HostError and add some helpers

graydon:
This adds an `impl Error for HostError` (which was accidentally lost when removing the use of `thiserror`) as well as a couple helper functions that factor out error-generating paths in host function.

### DIFF
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -257,7 +257,7 @@ impl Host {
         self.err(DebugError::general().msg(msg))
     }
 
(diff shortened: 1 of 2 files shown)


stellar-contract-env-host/src/host.rs
```diff
@@ -257,6 +257,15 @@ impl Host {
         self.err(DebugError::general().msg(msg))
     }
 
+    pub(crate) fn err_conversion_rawval_to<T>(&self, rv: RawVal) -> HostError {
```
leighmcculloch, line 260:
```suggestion
    pub(crate) fn err_conversion_raw_val_to<T>(&self, rv: RawVal) -> HostError {
```

leighmcculloch (approved):

PR #252, opened 2022-07-23
Impl TryIntoEnvVal for all EnvVal<>: TryInto

leighmcculloch:
### What
Impl `TryIntoEnvVal` for all `EnvVal<>: TryInto`.
### Why
To surface the useful `.try_into_env_val()` and `.try_into_val()` functions for types that implement `TryFrom<EnvVal<>>` or `TryInto<> for EnvVal<>`.

This new blanket impl will be the first time for some types that multiple implementations are possible for this trait, and so there will be some places in the SDK that will have ambiguous uses of these functions that will require some type annotations. I've already run through them and there are only a half dozen that are easy to fill in.

### DIFF
--- a/stellar-contract-env-common/src/env_val.rs
+++ b/stellar-contract-env-common/src/env_val.rs
@@ -105,26 +105,36 @@ pub trait IntoEnvVal<E: Env, V>: Sized {
     fn into_env_val(self, env: &E) -> EnvVal<E, V>;
 }
 
-pub trait IntoVal<E: Env, V>: IntoEnvVal<E, V> {
+pub trait IntoVal<E: Env, V> {
+    fn into_val(self, env: &E) -> V;
+}
+
+impl<E: Env, V, T> IntoVal<E, V> for T where
+T: IntoEnvVal<E, V> {
     fn into_val(self, env: &E) -> V {
         Self::into_env_val(self, env).val
     }
 }
-
-impl<E: Env, V, T> IntoVal<E, V> for T where T: IntoEnvVal<E, V> {}
 
 pub trait TryIntoEnvVal<E: Env, V>: Sized {
     type Error;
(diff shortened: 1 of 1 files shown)


stellar-contract-env-common/src/env_val.rs
```diff
@@ -123,6 +123,25 @@ pub trait TryIntoEnvVal<E: Env, V>: Sized {
...
+    fn try_into_env_val(self, env: &E) -> Result<EnvVal<E, T>, Self::Error> {
+        let ab: T = EnvVal {
+            env: env.clone(),
+            val: self,
+        }
+        .try_into()?;
+        Ok(EnvVal {
+            env: env.clone(),
+            val: ab,
+        })
```
leighmcculloch, line 141: All these EnvVal constructions look a bit intimidating from a performance point-of-view, especially since the latter one will be discarded in most places when used with `TryFromVal::try_from_val`, but it doesn't really matter because the for the wasm builds the `EnvVal` struct will have a zero sized `env`.

stellar-contract-env-common/src/env_val.rs
```diff
@@ -123,6 +123,25 @@ pub trait TryIntoEnvVal<E: Env, V>: Sized {
...
+    fn try_into_env_val(self, env: &E) -> Result<EnvVal<E, T>, Self::Error> {
+        let ev: T = EnvVal {
+            env: env.clone(),
+            val: self,
+        }
+        .try_into()?;
+        Ok(EnvVal {
+            env: env.clone(),
+            val: ev,
+        })
```
leighmcculloch, line 141: All these EnvVal constructions look a bit intimidating from a performance point-of-view, especially since the latter one will be discarded in most places when used with `TryFromVal::try_from_val`, but it doesn't really matter because the for the wasm builds the `EnvVal` struct will have a zero sized `env`.

PR #253, opened 2022-07-25
license everything

graydon:
Add license blocks to all files and license files to the root.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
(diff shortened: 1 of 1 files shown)


stellar-contract-env-common/src/convert.rs
```diff
@@ -1,3 +1,7 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
```
leighmcculloch, line 4: Is this necessary? We don't do this in most of our repos.

LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2016 Stellar Development Foundation
```
leighmcculloch, line 190:
```suggestion
   Copyright 2022 Stellar Development Foundation
```

COPYING
```diff
@@ -0,0 +1,15 @@
...
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+As well as the file LICENSE adjacent to this file.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
```
leighmcculloch, line 15: I don't think we need to follow the GNU pattern of including a COPYING file. The LICENSE file is sufficient. Can we drop this? Or do we need it for some specific reason. Fwiw we don't include a COPYING in most repos.

stellar-contract-env-common/src/convert.rs
graydon: Opinions vary but not a lot -- eg. the apache project recommends putting it in every file, as do most lawyers I've ever asked. It clarifies 2 questions that can get lost otherwise: who has copyright and what license it came under.

COPYING
graydon: You have one in the go repo, we have one in core .. I'm not sure what you mean by "most"

graydon: (The LICENSE file does not say who owns the copyright or connect that ownership to an act of licensing -- it's a different statement)

leighmcculloch: According to GitHub we don't include it in most of our repos, but you're right, does look like it is in go too. Sounds fine to me, just seems unnecessary.

leighmcculloch: https://github.com/stellar/rs-stellar-contract-env/blob/eafb7d5a8b17886844b059724897af3803403b72/LICENSE#L190-L196

graydon:
Modified to _just_ have LICENSE file.

PR #254, opened 2022-07-25
Add new workflow for checking interface version changes

graydon:
Check when someone touches env.rs that they also touch the meta.rs file, bump interface version, regenerate example wasms.

### DIFF
--- a/.github/workflows/check-bump-interface-version.yml
+++ b/.github/workflows/check-bump-interface-version.yml
@@ -16,7 +16,7 @@ jobs:
       uses: pablo-statsig/gh-pr-content-checker@c5b547fff05b0617977daa9e622f1a9e0c7be89a
       with:
         github-token: ${{github.token}}
-        diffContainsRegex: "interface_version: \d+,"
+        diffContainsRegex: "interface_version: \\d+,"
     - name: "check stellar-contract-env-host/wasm-examples/*.wasm were modified"
       uses: brettcannon/check-for-changed-files@36aaa76474fc979db4e19e44c05e8e978759ac01
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/check-bump-interface-version.yml
```diff
@@ -0,0 +1,25 @@
+name: check-bump-interface-version
+
+on:
+  push:
+    paths: 'stellar-contract-env-common/src/env.rs'
```
leighmcculloch: You can probably omit this as having this check run on the main branch is probably defeating the purpose a bit and given our current setup I don't think we'll get a notification of it. We need to catch this in PRs and we can because we require PRs to be up-to-date anyway.
```suggestion
```

.github/workflows/check-bump-interface-version.yml
```diff
@@ -0,0 +1,25 @@
...
+  pull_request:
+    paths: 'stellar-contract-env-common/src/env.rs'
+
+jobs:
+  check_pr:
+    runs-on: 'ubuntu-20.04'
+    steps:
+    - name: "check stellar-contract-env-common/src/meta.rs was modified"
+      uses: brettcannon/check-for-changed-files@v1
```
leighmcculloch, line 15: It's a security hardening best practice to reference third-party actions with a commit sha so that third-parties can't inject code into our builds at their discretion. Ref: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
```suggestion
      uses: brettcannon/check-for-changed-files@ 551263a821fde32615499b71f9283a7e98fcc657 # v1.0.1
```

Can you update the others too?

graydon:
Something is still not working. Not triggering. I will try deleting and reopening.

PR #256, opened 2022-07-25
Impl binary linear memory fns for the native mode

jayz22:
### What

Resolves https://github.com/stellar/rs-stellar-contract-env/issues/179.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/stellar-contract-env-host/src/host.rs
+++ b/stellar-contract-env-host/src/host.rs
@@ -40,6 +40,8 @@ use crate::{
     ConversionError, EnvBase, IntoEnvVal, Object, RawVal, RawValConvertible, Status, Symbol, Val,
     UNKNOWN_ERROR,
 };
+#[cfg(not(feature = "vm"))]
+use std::ptr::{slice_from_raw_parts, slice_from_raw_parts_mut};
 
 mod error;
 pub use error::HostError;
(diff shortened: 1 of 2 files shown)


jayz22:
Spoken to @graydon about this. This is no longer needed. With https://github.com/stellar/rs-stellar-contract-env/commit/d638ac66888c377ac37861411bd635283ed5a987 , the correct way for a memcpy in the native mode is to call `EnvBase::binary_copy_from_slice` which does not involve any raw pointers. 
In the guest mode (in VM), the guest will call the same guest interface `guest::binary_copy_from_slice` that decays a slice into two `RawVal::u32` for the host which then invokes the `binary_copy_from_linear_memory` hostfn. 
In the scenario which native contract developer calls `binary_copy_from_linear_memory` by mistake, if the VM features are not included (`cfg(not(feature="VM"))`), this should remain `unimplemented!()`.

leighmcculloch:
> the correct way for a memcpy in the native mode is to call `EnvBase::binary_copy_from_slice` which does not involve any raw pointers

Fyi I've hidden these fns for the moment, because they need to be moved into the `Binary` type before we expose them in the SDK API. This work is captured by https://github.com/stellar/rs-stellar-contract-sdk/issues/247.

PR #259, opened 2022-07-26
Add conversions Symbol <-> ScVal

jonjove:
### What

Added TryFrom Symbol/ScVal.

### Why

Make https://github.com/stellar/rs-stellar-contract-sdk/pull/316 work.

### Known limitations

This is eating conversion errors, in the interest of getting something working as quickly as possible.

### DIFF
--- a/stellar-contract-env-common/src/symbol.rs
+++ b/stellar-contract-env-common/src/symbol.rs
@@ -282,6 +282,7 @@ impl FromIterator<char> for Symbol {
     }
 }
 
+#[cfg(feature = "std")]
 use crate::xdr::ScVal;
 
 #[cfg(feature = "std")]
--- a/stellar-contract-env-common/src/symbol.rs
+++ b/stellar-contract-env-common/src/symbol.rs
@@ -282,6 +282,46 @@ impl FromIterator<char> for Symbol {
     }
 }
 
+use crate::xdr::ScVal;
+
+#[cfg(feature = "std")]
+impl TryFrom<ScVal> for Symbol {
+    type Error = ();
+    fn try_from(v: ScVal) -> Result<Self, Self::Error> {
+        (&v).try_into()
+    }
+}
+#[cfg(feature = "std")]
+impl TryFrom<&ScVal> for Symbol {
+    type Error = ();
+    fn try_from(v: &ScVal) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


jonjove:
@leighmcculloch @jayz22 I'm going to merge this as soon as possible so I can finish the macros work. We can clean up anything we need regarding errors in a separate PR.

stellar-contract-env-common/src/symbol.rs
```diff
@@ -282,6 +282,46 @@ impl FromIterator<char> for Symbol {
     }
 }
 
+use crate::xdr::ScVal;
+
+#[cfg(feature = "std")]
+impl TryFrom<ScVal> for Symbol {
+    type Error = ();
```
leighmcculloch, line 289: ConversionError?

stellar-contract-env-common/src/symbol.rs
```diff
@@ -282,6 +282,46 @@ impl FromIterator<char> for Symbol {
...
+        ))
+    }
+}
+
+#[cfg(feature = "std")]
+impl TryFrom<&Symbol> for ScVal {
+    type Error = ();
+    fn try_from(s: &Symbol) -> Result<Self, Self::Error> {
+        s.clone().try_into()
+    }
```
leighmcculloch, line 322: I usually try to do these in reverse, implement for the ref version, and then have the other one cast to it. You're collecting in the non-ref version anyway, so then the ref version becomes 2x clones.

stellar-contract-env-common/src/symbol.rs
jonjove: We'll clean it up in follow-up work. I'm not up to date on the best practices regarding error propagation in the host crate.

leighmcculloch (approved):
Could you fix the error to be ConversionError? Otherwise LGTM.

I note an optimization, but nbd.

stellar-contract-env-common/src/symbol.rs
jonjove: Does the reverse work in this case? I'm calling `into_iter` which takes ownership. Is there a non-into option here?

stellar-contract-env-common/src/symbol.rs
jonjove: Opened #260

PR #271, opened 2022-07-26
More struct wrapping

graydon:
This is a followup to https://github.com/stellar/rs-stellar-contract-env/commit/85b352afa70530d4561b273abd0462fe3440dfc5 .. I'm not 100% sure we want to continue down this path, but I think we _might_.

I actually _think_ this is basically 1/3-1/2 of the work necessary to back out the use of `TaggedVal` as an optimistic-but-wrong adventure in attempting to generalize-and-simplify multiple separate wrapper types into one more-generic one. I think this has plausibly proven to be a mistake -- it is at least a lot of cognitive load and maintenance burden -- and I suggest that it might make sense to try to complete the adventure by backing it all the way out, and implementing concrete wrapper types via macros alone and/or manual replication of code, not generics.

The benefit to users are that there's much less type-system noise especially in error messages, and less trait-inference to fall prey to ambiguities and the need for "trait features that don't actually exist in stable rust or possibly ever" like specialization. Downside is we either write lots of code by hand or have more macros in the implementation. But I think I'm willing to give it a try, @leighmcculloch WDYT?

### DIFF
--- a/stellar-contract-env-common/src/bitset.rs
+++ b/stellar-contract-env-common/src/bitset.rs
@@ -95,6 +95,21 @@ impl RawValConvertible for BitSet {
     }
 }
 
+#[cfg(feature = "vm")]
+impl wasmi::FromValue for BitSet {
+    fn from_value(val: wasmi::RuntimeValue) -> Option<Self> {
+        let maybe: Option<TaggedVal<TagBitSet>> = val.try_into();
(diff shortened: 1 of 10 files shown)


leighmcculloch:
+1 I think this is good. Can hold off merging this until #269 merged?

graydon:
continued in #273

PR #273, opened 2022-07-27
Kill tagged val

graydon:
This follows on #271 and finishes pulling the bandaid off, removing `TaggedVal` entirely, replacing it with a shared boilerplate macro that generates the necessary impls. This went remarkably smoothly and I have really nothing bad to say about it except maybe there is more deleting to do from the middle of the macro and maybe other places if this follows / is combined with #269 as well.

Overall I think this is a significant improvement in comprehensibility, though it means a bunch of stuff is macro-generated.


stellar-contract-env-common/src/val_wrapper.rs
```diff
@@ -0,0 +1,151 @@
...
+                    env: ev.env,
+                    val: ev.val.to_raw(),
+                }
+            }
+        }
+        impl<E: Env> From<EnvVal<E, $tagname>> for RawVal {
+            fn from(ev: EnvVal<E, $tagname>) -> Self {
+                ev.val.0
+            }
+        }
```
leighmcculloch, line 59: I think we need some more conversions:
- `impl From<$tagname> for RawVal`
- `impl<E: Env> TryIntoVal<E, RawVal> for $tagname`

leighmcculloch: Oh, I see the first one further up, so no worries.

leighmcculloch: Which also means the second one should be blanket impl.

leighmcculloch (approved):
Would be good to see what the SDK changes look like before merging this, otherwise this is 💯.

stellar-contract-env-common/src/val_wrapper.rs
graydon: I was about to add these but ..

`From<$tagname> for RawVal` is already implemented up on line 20.

and I _think_ `TryIntoVal<E, RawVal>` comes for free from the fact that raw_val.rs line 121 has a `impl<E: Env> TryFrom<EnvVal<E, RawVal>> for $T` and then you blanket-impl `TryIntoVal` on env_val.rs line 94.

IOW I think they're already covered. If you find others you need, by all means.

stellar-contract-env-common/src/val_wrapper.rs
```diff
@@ -0,0 +1,151 @@
...
+                        }
+                    }
+                    None => None,
+                }
+            }
+        }
+        #[cfg(feature = "vm")]
+        impl From<$tagname> for wasmi::RuntimeValue {
+            fn from(v: $tagname) -> Self {
+                wasmi::RuntimeValue::I64(v.as_raw().get_payload() as i64)
```
jayz22, line 97: What's the use case for these conversions to/from `wasmi::RuntimeValue`? So that we can specify them directly as the host function argument types instead of `RawVal`? 
Also, why is the `from_value` expecting an `u64` while the `from` converts the payload to an `i64`?

PR #284, opened 2022-07-28
Sketch of release automation

graydon:
The idea here is that we have two new automation triggers:

  - Publishing a github release will do a CI build and then run `cargo publish`
  - Pushing to `release-dry-run*` will do the same but run `cargo publish --dry-run`
 
Several details I'm a bit hazy on (the exact nature of secret-redaction github does, the env-var overriding cargo supports, the expression language used for needs-skipping, the events and ref-names, etc.) so suggestions welcome. I will avoid opening more copies of this on the other repos until we've decided on how to do this one.


.github/workflows/rust.yml
```diff
@@ -54,3 +56,29 @@ jobs:
...
+    - if: contains(needs.*.result, 'failure')
+      run: exit 1
+    - uses: actions/checkout@v3
+    - run: rustup update
+    - run: cargo publish
+    env:
+    - CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  cargo-publish-dry-run:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/heads/release-dry-run')
```
leighmcculloch, line 74:
```suggestion
    if: github.event_name == 'push' && startsWith(github.ref_name, 'release-dry-run')
```

.github/workflows/rust.yml
```diff
@@ -54,3 +56,29 @@ jobs:
...
+    needs: [complete]
+    runs-on: ubuntu-latest
+    steps:
+    - if: contains(needs.*.result, 'failure')
+      run: exit 1
+    - uses: actions/checkout@v3
+    - run: rustup update
+    - run: cargo publish
+    env:
+    - CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
leighmcculloch, line 71: It would be ideal if we only expose the secret to the publish command, and not everything else.
```suggestion
      env:
      - CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```

.github/workflows/rust.yml
```diff
@@ -54,3 +56,29 @@ jobs:
...
     - if: matrix.sys.test
       run: cargo hack --feature-powerset test --target ${{ matrix.sys.target }}
+
+  cargo-publish:
+    if: github.event_name == 'release'
+    needs: [complete]
+    runs-on: ubuntu-latest
+    steps:
+    - if: contains(needs.*.result, 'failure')
+      run: exit 1
```
leighmcculloch, line 66: This shouldn't be necessary. Jobs will only unconditionally run if they have `if: always()`. Are you finding this is necessary for some reason?

leighmcculloch (commented):
I think this looks decent. Some comments inline.

.github/workflows/rust.yml
```diff
@@ -54,3 +56,29 @@ jobs:
...
+    - uses: actions/checkout@v3
+    - run: rustup update
+    - run: cargo publish
```
leighmcculloch, line 69: Can you add the --frozen option? It should make sure the build process only uses what is in the lock file, so we know for sure the lock file is valid.
```suggestion
    - run: cargo publish --locked
```
The lock file will be discarded by crates.io, but it will jsut be a safeguard that the lock file we have at this moment is legit. We should probably add --locked to all the other jobs in here too.

.github/workflows/rust.yml
graydon: I thought jobs were unconditionally run (or rather the success/failure of their `needs` jobs are ignored) if they have _any_ `if:` condition

.github/workflows/rust.yml
graydon: you prefer `--frozen` or `--locked`?

.github/workflows/rust.yml
graydon: Github docs say "If a job fails, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue" which suggests to me that the `if:` I have there _would_ allow the job to continue unless I inspect its `needs` otherwise.

.github/workflows/rust.yml
graydon: (I _think_ `--frozen` won't work since that says "don't touch the network", so then .. `--locked`?)

leighmcculloch: I think frozen just won't touch the network to check, but it might still download, I'm not sure. Locked is probably sufficient.

graydon: Ok pushed locked everywhere.

.github/workflows/rust.yml
leighmcculloch: >You can use the following status check functions as expressions in if conditionals. A default status check of success() is applied unless you include one of these functions.

Ref: https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

Unless the `if:` references one of the four status functions GitHub automatically adds `success()` as a condition. Unless I'm missing something, I don’t think you need the manual exit 1 failure in these jobs.

.github/workflows/rust.yml
```diff
@@ -51,6 +53,32 @@ jobs:
...
-      run: cargo hack --feature-powerset test --target ${{ matrix.sys.target }}
+      run: cargo hack --feature-powerset test --locked --target ${{ matrix.sys.target }}
+
+  cargo-publish:
+    if: github.event_name == 'release'
+    needs: [complete]
+    runs-on: ubuntu-latest
+    steps:
+    - if: contains(needs.*.result, 'failure')
+      run: exit 1
```
graydon, line 66:
```suggestion
```

.github/workflows/rust.yml
```diff
@@ -51,6 +53,32 @@ jobs:
...
+      env:
+      - CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  cargo-publish-dry-run:
+    if: github.event_name == 'push' && startsWith(github.ref_name, 'release-dry-run')
+    needs: [complete]
+    runs-on: ubuntu-latest
+    steps:
+    - if: contains(needs.*.result, 'failure')
+      run: exit 1
```
graydon, line 79:
```suggestion
```

.github/workflows/rust.yml
graydon: fascinating. removed!

.github/workflows/rust.yml
```diff
@@ -2,8 +2,10 @@ name: Rust
 
 on:
   push:
-    branches: [main]
+    branches: [main, "release-dry-run*"]
```
leighmcculloch, line 5: I think we should rename this to `release/**`, and plan to use it for a release branches in general, not just as dry runs.
```suggestion
    branches: [main, "release/**"]
```

.github/workflows/rust.yml
```diff
@@ -51,6 +53,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: rustup update
+    - run: cargo publish --locked
+      env:
+      - CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  cargo-publish-dry-run:
+    if: github.event_name == 'push' && startsWith(github.ref_name, 'release-dry-run')
```
leighmcculloch, line 72:
```suggestion
    if: github.event_name == 'push' && startsWith(github.ref_name, 'release/')
```

.github/workflows/rust.yml
```diff
@@ -51,6 +53,28 @@ jobs:
...
+    needs: [complete]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: rustup update
+    - run: cargo publish --locked --dry-run
+      env:
+      - CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
leighmcculloch, line 80: Does the dry run need the token? It might be worth trying without it just to reduce the places that get the token to the absolute minimal. But if it's needed lgtm.

leighmcculloch, line 80: Maybe we should dry run with multiple targets, like x86 linux and wasm32, as well as run the publish dry run with cargo hack. I think the only check that the versions in the dependencies (and not the git URLs) are correct, will be done by the publish, which means we should probably check all the features too.

leighmcculloch (commented):
This looks mostly good. Couple small things, and then the only big concern is the dry run, and we should probably have the dry run do a run through all features using cargo hack. See comment inline.

PR #287, opened 2022-07-29
Add some host API docs

graydon:
More docs!

### DIFF
--- a/soroban-env-common/src/checked_env.rs
+++ b/soroban-env-common/src/checked_env.rs
@@ -69,7 +69,7 @@ macro_rules! generate_checked_env_trait {
     {
         // This macro expands to a single item: the CheckedEnv trait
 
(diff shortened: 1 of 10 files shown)


soroban-env-common/src/checked_env.rs
```diff
@@ -69,7 +69,7 @@ macro_rules! generate_checked_env_trait {
     {
         // This macro expands to a single item: the CheckedEnv trait
 
-        /// This trait is a variant of the [crate::Env] trait used to define the
+        /// This trait is a variant of the [Env](crate::Env) trait used to define the
```
leighmcculloch, line 72: Oh nice. I didn't know you could do this. I've been doing the below, but it's somewhat annoying to have `use` statements out in teh code for docs. I like this way you're doing, I'll start doing same.
```rust
#[cfg(doc)]
use crate::Env;
```

PR #297, opened 2022-08-01
Set default budget limits to unlimited

jayz22:
### What

So that we don't block testing and running examples. 

### Why

https://github.com/stellar/rs-soroban-sdk/issues/382

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/budget.rs
+++ b/soroban-env-host/src/budget.rs
@@ -258,7 +258,7 @@ impl Default for Budget {
         // using the calibration above. Very roughly!
         // b.mem_bytes.limit = 0xa_0000;
         // b.cpu_insns.limit = 400_000;
-        // TODO: come up with proper limits once all the machinary is in place and models are calibrated.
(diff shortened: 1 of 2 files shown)


soroban-env-host/src/budget.rs
```diff
@@ -256,8 +256,11 @@ impl Default for Budget {
         // 4bn instructions / sec. So about 4000 instructions per usec, or 400k
         // instructions in a 100usec time budget, or about 5479 wasm instructions
         // using the calibration above. Very roughly!
-        b.mem_bytes.limit = 0xa_0000;
-        b.cpu_insns.limit = 400_000;
+        // b.mem_bytes.limit = 0xa_0000;
+        // b.cpu_insns.limit = 400_000;
+        // TODO: come up with proper limits once all the machinary is in place and models are calibrated.
```
leighmcculloch, line 261:
```suggestion
        // TODO: Set proper limits once all the machinary (including in SDK tests) is in place and models are calibrated.
```

leighmcculloch (approved):
👏🏻 Thanks!

PR #299, opened 2022-08-02
Clean up host storage/data related code

jayz22:
### What

Split out some storage access, data massage code into a new data helper file. Some code goes into the conversion helper file.  

### Why

Shorten the `host`, make it clearer in order to prepare for some cost budget/metering work. 

### Known limitations

[TODO or N/A]


graydon (approved):
Looks good, thanks!

PR #308, opened 2022-08-03
Merge github.com/stellar/rs-soroban-sdk

leighmcculloch:
### What
Merge github.com/stellar/rs-soroban-sdk

### Why
Much of the sdk lives in env-common, and we seem to be constantly developing both in conjunction. It feels artificial that they live in separate repos.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,7 +35,7 @@ jobs:
         - os: ubuntu-latest
           target: x86_64-unknown-linux-gnu
           test: true
-        # TODO: Re-enable if we see value in doing so.
+        # TODO: Re-enable these builds if we see value in doing so.
         # - os: macos-latest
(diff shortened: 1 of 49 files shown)


leighmcculloch:
I want to merge this with a merge commit, but annoying GitHub won't let me do that.

leighmcculloch:
This PR appears to not preserve history of the original tree as well as I had hoped, I'll have to have another crack at it.

leighmcculloch:
After talking with @tomerweller and @ire-and-curses, some things said has changed my mind about chasing this. I think we do need some separation of process between the env and sdk repos. And while a decent amount of the SDK lives in the env-common crate, it is more static, less likely to change, and we can use the `CODEOWNERS` file in this repo to have GitHub require the Core team to review changes to the host, while letting others make changes to common. So I think this is good as is.

PR #313, opened 2022-08-03
Replace FrameGuard drop trait with guarded closure; add metering

jayz22:
### What

- Remove the `drop` trait from `FrameGuard` which automatically cleans up after frame goes out of scope. 
- Add a `roll_back` function that does the clean up.
- Wrap all push_frame calls in `with_frame_guard`, which automatically calls `commit` on success and `roll_back` on err. 
- Add metering to frame pushing and popping.

### Why

Preparing for the metering change. 
The frame guard functions needs to be metered (it needs to copy data), which may fail due to reaching budget limit, therefore they need to return `Result`. `drop` function does not allow that.
As discussed with @graydon , this is the alternative "cleanup-upon-going-out-of-score" pattern to `drop`. 

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/budget.rs
+++ b/soroban-env-host/src/budget.rs
@@ -13,6 +13,8 @@ pub enum CostType {
     HostEventDebug = 6,
     HostFunction = 7,
     VisitObject = 8,
+    PushFrame = 9,
+    PopFrame = 10,
 }
 
 // TODO: add XDR support for iterating over all the elements of an enum
@@ -28,6 +30,8 @@ impl CostType {
             CostType::HostEventDebug,
             CostType::HostFunction,
             CostType::VisitObject,
+            CostType::PushFrame,
+            CostType::PopFrame,
         ];
         VARIANTS.iter()
     }
@@ -246,6 +250,7 @@ impl Default for Budget {
         b.cpu_insns
             .get_cost_model_mut(CostType::WasmInsnExec)
             .lin_param = 73;
+        // TODO: set cpu, mem model parameters for other CostTypes
 
         // Some "reasonable defaults": 640k of RAM and 100usec.
         //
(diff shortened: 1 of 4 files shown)


graydon (changes requested):
This is close but a little off from what I was hoping for. Could you make a couple changes?

1. I understand keeping the multiple "push" functions separate from the single "with" function, but I think there is no longer any need for the `FrameGuard` object itself. All it does is wrap a `RollbackPoint`. So I think the "push" functions should just return a `Result<RollbackPoint, HostError>` and the "with" function can take the `RollbackPoint` as an argument, which it then passes through to `pop_frame` on error.
2. I thought part of the point of this was to enable charging for the capture and/or restoration of the `RollbackPoint`; is that something you're still intending to do, maybe in a later PR? Or did you want to put some of it in here also?

jayz22:
> I think there is no longer any need for the FrameGuard object itself. All it does is wrap a RollbackPoint.

I see. Yes you are right. Since we are explicitly handling the drop logic when the closure returns error, that means we don't need the FrameGuard any more. Does that deviate from your expectation? Is there any disadvantages for doing things this way? I'm not seeing any (I actually think this way is much cleaner) and I will make the change. Just wondering.

> enable charging for the capture and/or restoration of the RollbackPoint; is that something you're still intending to do, maybe in a later PR? 

Yes that's something I intended to do, but in a later PR as that will touch the `CostType` which I need to hash things out (e.g. with this [PR](https://github.com/stellar/rs-soroban-env/pull/315)) . Trying to keep this PR clean.

jayz22:
@graydon Fixed both your requests. 
I decided to add the new `CostType`s and hooked the `charge_budget` functions in which revealed even more cleanups! 
I also try to add comments whenever I can, since we are now replacing FrameGuard with a less obvious pattern, which is the only con I see in this approach. But otherwise I think this looks much cleaner than before (minus 50 some lines in the host :).

graydon (changes requested):
This is almost right, but there are two separate parts to the task:

1. Any frame that's pushed should always be popped, so that we maintain the correct context.
2. A pop that happens after an _error_ should also roll back the storage and object state.

I've rearranged the code and pushed a change that shows what I mean. If you are ok with it then we can merge.

jayz22:
> I've rearranged the code and pushed a change that shows what I mean. If you are ok with it then we can merge.

@graydon This makes sense, thank you!

PR #315, opened 2022-08-04
Add metering to Val conversions, host function invocation

jayz22:
### What

- Adds a new budget type that covers for simple val conversion. E.g `RawVal` to a non-object `ScVal`, or to a `ScObject::U64`. With a test.
- Add `HostFunction` metering to the dispatcher functions. With a test.
- Add several testing utils functions to make it easy to test budget and metering.
- Refactored the invocation tests code to use `utils` functions.

### Why

Part of the bigger picture of budget metering.

### Known limitations

[TODO or N/A]


soroban-env-host/src/budget.rs
```diff
@@ -153,13 +167,13 @@ impl BudgetDimension {
     }
 
     pub fn is_over_budget(&self) -> bool {
+        if self.limit == u64::MAX {
+            return false;
+        }
```
graydon, line 172:
```suggestion
```

graydon: This is redundant with the existing condition below.

graydon (changes requested):
Looks fine except for the one suggested block to remove.

soroban-env-host/src/budget.rs
jayz22: Thanks, fixed.

PR #319, opened 2022-08-09
Native token

jonjove:
### What

Adds machinery (still incomplete) for native contracts and specifically implements a token contract.

### Why

Native contracts are one of our goals, and this is an experiment in making it work.

### Known limitations

The work is incomplete.


soroban-env-macros/src/lib.rs
```diff
@@ -66,3 +74,54 @@ pub fn generate_env_meta_consts(input: TokenStream) -> TokenStream {
...
+
+fn get_methods(imp: &ItemImpl) -> impl Iterator<Item = &ImplItemMethod> {
+    imp.items.iter().filter_map(|i| match i {
+        ImplItem::Method(m) => Some(m),
+        _ => None,
+    })
+}
+
+#[proc_macro_attribute]
+pub fn contractimpl(_metadata: TokenStream, input: TokenStream) -> TokenStream {
```
leighmcculloch, line 113: This PR appears to duplicate a decent amount of logic from the SDK. Could you share a sketch of the plan and vision you're working towards and how these will co-exist?

jonjove: This is an experiment in doing two things:

1. Rewrite the token contract so it never panics (unless the host functions themselves panic…) using `CheckedEnv`
2. Duplicate and modify as necessary the smallest possible set of SDK functionality to support (1)

For comparison in relation to (2): the SDK is 6101 lines (5492 excluding tests) whereas the part of the SDK that was duplicated and modified in this PR is 561 lines. Basically, we only need 10% of the SDK to actually write the native contracts that are currently under consideration.

In terms of how these will co-exist, I don’t know the answer yet. My goal was to prove that it would be possible to deliver native contracts in a not-completely-terrible way, not to find the best possible way to do it. In that sense my priority was to achieve (1) by whatever means necessary, and this PR—excluding a few mistakes that will be fixed soon—achieves that. In the process, what I implemented for (2) tried to track the SDK as much as possible but deviated where necessary to support the requirement to never-be-anywhere-close-to-panic. A large portion of the deviation is related to using `CheckedEnv` instead of `Env`.

My plan in the short term is to bring this code to a mergeable state, then merge it so we can build the remainder of the token functionality. From there, I would hope that we can collaborate on improving this model and making it easier to maintain (perhaps by finding places where the SDK and native semantics can be unified). Does this sound reasonable to you? I'm definitely open to suggestions here, but I also want to make sure we stay cognizant of the importance to meet our goals and deliver in a timely manner.

cc @MonsieurNicolas because we've discussed native contracts quite a bit

leighmcculloch: Does this approach mean that the token contract will not be buildable as a wasm contract?

The `soroban-sdk` and others use and exposes types from the env crates. So as to clearly scope the native SDK and to not make it confusing whether they should or shouldn't be used in other contexts, could all its types and macros be placed into crates that are separate from the existing env crates? E.g.:
- `soroban-native-sdk`
- `soroban-native-sdk-macros`

jonjove: > Does this approach mean that the token contract will not be buildable as a wasm contract?

Right now, it's not buildable as a WASM contract because (a) it depends on `CheckedEnv` which isn't available as guest and (b) it emits `HostError`. These are related, because `CheckedEnv` directly exposes `HostError`. It would be possible to work around these issues, but I'm not sure it's worthwhile. Despite the fact that it can't be built as WASM, the code is actually quite similar (the main differences are function return types and question marks everywhere). I agree that it would be interesting to explore exposing the `CheckedEnv` interface and corresponding "native SDK" machinery to contracts. This would probably be a pretty big chunk of work, and would require some host machinery to determine which environment interface you get when the contract runs--this could perhaps be a different kind of contract code entry `WASM_CHECKED` vs `WASM_UNCHECKED`.

Right now I put the "native SDK" stuff into its own module inside of `Host`, and all the native contracts are inside of that module. This means the native SDK types don't leak out of that module, so they shouldn't be exported anywhere. I'm happy to move this to another crate if you prefer, just let me know. As for the macros, it probably would be best to put them into their own crate so I will definitely do that.

leighmcculloch: +1 I think the native contract types should move out of host into a native-sdk crate that has its own macro crate, following the same pattern we use in the non-native-sdk. I also think the token contract needs moving out to its own crate and it shouldn't be bundled inside the host as it is very much outside the scope of responsibility of the host I think.

jonjove: When we discussed this in the past, you and @graydon were opposed to creating so many separate crates because of maintenance and release burden. Are we happy with creating 3 new crates (all inside the env repo) to get the above separation? Just making sure that we aren't creating unnecessary work for ourselves.

jonjove: Just for clarity, I'm going to start separating into 3 new crates now (it won't be hard given how the code was initially partitioned into modules). But let's finish this conversation and if we decide on a different scheme then I'll update the separation.

leighmcculloch: We will automate the release process and so I'm less concerned about the release burden now. 

What's the motivation for moving the token contract into this repo vs having it live in the soroban-token-contract repo?

jonjove: Would you want to leave the current token contract in soroban-token-contract as well (make it a workspace with two crates)? Or would you want to move the current token contract into the examples repo? Or do something else entirely?

Personally, I don't see an issue with the token contract living inside the host crate. It is really part of the host, by virtue of being a native contract. But now that I think about it, I'm not even sure how the token contract could be moved into a separate crate. The token contract depends on the host (because it depends on things like `Host` and `HostError`), and the host depends on the token contract (because it needs to be able to run the token contract). Maybe we could try to break the dependency of token on host if we wanted... but it seems like a lot of work for not a lot of benefit right now. A different option would be to dynamically load the token contract, but that seems really overkill at this point. Thoughts?

jonjove: This same problem actually impacts the native SDK too. Native SDK depends on host (for `Host` and `HostError`), but host depends on SDK (for the token implementation). We should break these cycles. I think we might be able to by replacing `Host` and `HostError` with generics that `impl CheckedEnv`. I'd rather not do this right now because it's going to make everything looks much less like the normal SDK and normal token contract. If you're okay with this, that means only the macros will move into their own crate but I'll open an issue to improve the dependencies as future work.

leighmcculloch: 👍🏻 

Regarding what to do with the existing token contract: I think putting it in the example repo as you suggest is an appropriate thing to do if it won't be the standard and audited official token impl.

leighmcculloch: You may like to put the trait for the token contract in its own crate so that others can import it and the types? I don't feel strongly, just thinking.

PR #320, opened 2022-08-11
Support calibration benches on apple silicon macs

graydon:
This adds support for running the budget calibration benchmark code on M1 / apple silicon macs.

(It might also work on intel macs, I will give it a try tomorrow morning, but this should be enough to unblock @jayz22 on calibration)

### DIFF
--- a/soroban-env-host/Cargo.toml
+++ b/soroban-env-host/Cargo.toml
@@ -32,6 +32,9 @@ assert_matches = "1.5.0"
 optimization = "0.2.0"
 env_logger = "0.9.0"
 log = "0.4.17"
+tracking-allocator = "0.4.0"
+tabwriter = "1.2.1"
+thousands = "0.2.0"
 
 [features]
 vm = ["wasmi", "parity-wasm", "soroban-env-common/vm"]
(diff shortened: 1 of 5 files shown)


soroban-env-host/Cargo.toml
```diff
@@ -40,9 +43,10 @@ testutils = []
 
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 perf-event = "0.4.7"
-tracking-allocator = "0.4.0"
-tabwriter = "1.2.1"
-thousands = "0.2.0"
+
+[target.'cfg(all(target_os = "macos", target_arch = "aarch64"))'.dev-dependencies]
+macos-perf = "0.1.1"
+sudo = "0.6.0"
```
leighmcculloch, line 49: The description of the `sudo` crate is terrifying. From the docs.rs:
>Detect if you are running as root, restart self with sudo if needed or setup uid zero when running with the SUID flag set.

As far as I can tell we're only using the detection mechanism, and not the auto-restart as sudo. Right?

leighmcculloch (approved):
One question, otherwise this is exciting 👏🏻.

jayz22:
Thank you so much @graydon !

soroban-env-host/Cargo.toml
graydon: Yeah. I think I will actually be able to roll this change back and use getrusage_v4 instead, but as it stands this is "use for detection only" and "only in a benchmark" -- no need to be alarmed.

PR #321, opened 2022-08-11
Add ci workflow to set min rust-version

leighmcculloch:
### What
Add ci workflow to set min rust-version.

### Why
To set rust-version so that people see a meaningful error if they try to use an older version of rust-version.


graydon (approved):
LGTM

Issue #312, opened 2022-08-03
Add missing host functions related to ledger header information


MonsieurNicolas:
Gaps:
* needed to write any kind of conditional based on "time"
    * ledger number
    * ledger close time
* needed to deal with core protocol ("host version") upgrades
    * protocol version

leighmcculloch:
For the SDK side of this, if it is added at the same time, I think we should add a `Ledger` type that has these fns impl, where that ledger value is retrievable via the Env. 

```rust
impl Env {
    pub fn current_ledger(&self) -> Ledger
```

```rust
impl Ledger {
    pub fn number(&self) -> u32;
    pub fn close_time(&self) -> ?;
```

This pattern is the same pattern we use for `ContractData`.

graydon:
On the host side I think we could do this either with some reserved magic-value storage keys or with something even lighter-weight like a constant symbol=>val map that a given host is populated with when it's constructed, representing its current (fixed-during-its-lifetime) execution context.

I think we should largely avoid providing host functions (or even SDK functions) with stellar-specific terminology; it would be hard to walk back / make-generic if we specialize too much, and relatively easy to keep generic / non-specialized.

leighmcculloch:
I think we should avoid magically named contract data elements. They won't be naturally documented. They won't have types. I think SDK functions are appropriate, people will use this information in contracts and we should make it accessible.

In Solidity you get access to a `block` which has a `timestamp`, a `blockhash`, and a `blocknumber`, and so I don't think it is bespoke to provide similar information.

We could avoid Stellar terminology and use the term `timestamp` instead of `close_time`.

We could go further and use the term `block` instead of `ledger`, however `ledger` is particularly Stellar specific anyway, so it probably has little value in changing.

jonjove:
We should also be realistic about the fact that some of the things we are building have stellar-specific meaning. To me that is both expected and important for getting a good level of integration. We have the host functions that get account thresholds and signers, for example. The token contract contains the notion of accounts in it. There will be host functions to move assets back and forth across the classic/smart boundary (which will only be usable by the token contract).

We should have a feature that we use to enable/disable the stellar-specific functionality. We could even move it into its own module to make it easy to include/exclude it.

graydon:
Posted a version of this as described. Nothing fancy but enough to build on.

leighmcculloch:
> needed to deal with core protocol ("host version") upgrades
> * protocol version

[...]

PR #325, opened 2022-08-12
Add ledger info host functions

graydon:
Basic implementation of #312 -- some host functions to access new constant data from the invocation context. It's optional, and if a user tries to access it when it hasn't been set it will trap. We'll set it in core and, I guess, the CLI will have to set it also to some kind of synthetic values.


soroban-env-common/src/env.rs
```diff
@@ -152,6 +152,15 @@ macro_rules! call_macro_with_all_host_functions {
                 /// running contract. Traps if the running contract was not
                 /// invoked by a contract.
                 {"4", fn get_current_contract() -> Object }
+
+                /// Return the protocol version of the current ledger as a u32.
+                {"5", fn get_ledger_version() -> RawVal }
```
leighmcculloch, line 157: Is `ledger_version` the term we typically use for this? It makes it sound like a single ledger could exist in multiple states, like this is ledger number 10 version 1, then this is it in version 2. Should we just call it protocol version?

Maybe we should leave it out. From best I can tell Ethereum doesn't expose a version to contracts, at least not noted here: https://docs.soliditylang.org/en/latest/units-and-global-variables.html

@jonjove @tamirms @paulbellamy  Do you know if Ethereum exposes a version number to contracts so they can change their behavior depending on version?

soroban-env-common/src/env.rs
```diff
@@ -152,6 +152,15 @@ macro_rules! call_macro_with_all_host_functions {
...
+                /// Return the sequence number of the current ledger as a u32.
+                {"6", fn get_ledger_sequence() -> RawVal }
+                /// Return the timestamp number of the current ledger as a u64.
+                {"7", fn get_ledger_timestamp() -> RawVal }
+                /// Return the network identity of the current ledger as bytes.
+                {"8", fn get_ledger_network_id() -> RawVal }
```
leighmcculloch, line 163: Is the network ID the same thing as the network passphrase? Sgtm if so, I'm just not sure.

leighmcculloch (commented):
Couple questions, but lgtm, thanks for adding.

soroban-env-common/src/env.rs
```diff
@@ -152,6 +152,15 @@ macro_rules! call_macro_with_all_host_functions {
                 /// running contract. Traps if the running contract was not
                 /// invoked by a contract.
                 {"4", fn get_current_contract() -> Object }
+
+                /// Return the protocol version of the current ledger as a u32.
+                {"5", fn get_ledger_version() -> RawVal }
+                /// Return the sequence number of the current ledger as a u32.
+                {"6", fn get_ledger_sequence() -> RawVal }
+                /// Return the timestamp number of the current ledger as a u64.
+                {"7", fn get_ledger_timestamp() -> RawVal }
```
jayz22, line 161:
```suggestion
                {"7", fn get_ledger_timestamp() -> Object }
```

soroban-env-common/src/env.rs
jayz22, line 163: Both "7" and "8" should return `Object` instead of `RawVal`? 
The convention so far has been if an argument is an `Object` val type we explicitly denote such. We use `RawVal` only if it's one of the other 7 val types. 
E.g.:
`fn obj_from_u64(&self, u: u64) -> Result<Object, HostError> `

jayz22, line 163:
```suggestion
                {"8", fn get_ledger_network_id() -> Object }
```

leighmcculloch:
Superseded by #330.

PR #326, opened 2022-08-12
Switch mac calibration to older and more portable rusage_info_v4

### LINKED ISSUE #324: Switch to getrusage_v4 for mac calibration
In https://github.com/stellar/rs-soroban-env/pull/320 I switched on calibration for apple silicon. There's actually a different (older) interface in macos -- an extended variant of getrusage -- that reports per-process instruction counts as well. If this also still works on apple silicon we should switch to it since it does (or did at one point) work on intel macs as well.

(Generally it might be worth releasing a small utility crate that just abstracts over this functionality -- "get current instruction count" -- across platforms)

graydon:
This just switches our instruction-counting calibration machinery from the fancy M1-specific instruction counter I found yesterday to a more general (but still deeply fancy) super-extended rusage-based process-instruction-count facility that's existed on macOS since 10.13 / High Sierra / 2017 / XNU 4570.1.46, which I think is fairly close to the oldest still-supported macOS in existence (and surely older than any of our devs are using).

More importantly: this isn't aarch64-specific, it's an architecture-neutral interface so this works on the machines used by those of us still on archaic intel macs.

(Also gets rid of "having to run as root", which was awkward)

Fixes #324


soroban-env-host/benches/common/measure.rs
```diff
@@ -174,19 +174,26 @@ mod cpu {
     }
 }
 
-#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+#[cfg(target_os = "macos")]
+#[path = "."]
```
leighmcculloch, line 178: What does the path attribute do here? I've read about it, but unclear how it works on a mod that has a body.

leighmcculloch (commented):
🎉 This is great.

soroban-env-host/benches/common/measure.rs
graydon: it just overrides the default mapping of mod-with-submods-means-directory (by default rust would look for `cpu/rusagev4.rs`)

PR #332, opened 2022-08-13
Add Host::register_test_contract_wasm

leighmcculloch:
### What
Add Host::register_test_contract_wasm, which accepts a WASM byte slice and a contract ID to register it against.

### Why
When writing tests, developers will want to register not only their own contract, as is doable via `Host::register_test_contract`, but also other contracts for which they have only the WASM for so that they can do broader integration testing.

Some examples in the soroban-examples repo have achieved this by creating the other contract with the `Host::create_contract_from_contract` function. The problem with this approach is that the contract being created is created as if it was created by the developers contract. This is conceptually not aligned with how a developer thinks about deploying the other contract, because a contract that a developer writes won't always be deploying the contract it is interacting with.

Therefore it is helpful if we allow developers to simply register contracts without going through all the steps to deploy that contracts from within their own.

This is only for tests infrastructure and is gated behind the testutils feature.

### DIFF
--- a/soroban-env-host/Cargo.toml
+++ b/soroban-env-host/Cargo.toml
@@ -39,6 +39,7 @@ tabwriter = "1.2.1"
 thousands = "0.2.0"
 
 [features]
+default = ["vm"]
 vm = ["wasmi", "parity-wasm", "soroban-env-common/vm"]
 serde = ["soroban-env-common/serde"]
 testutils = []
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -546,18 +546,27 @@ impl Host {
         Ok(EnvVal { env, val: v })
     }
 
-    pub fn create_contract_helper(
+    pub fn create_contract_with_id(
+        &self,
+        contract: ScContractCode,
+        contract_id: Object,
+    ) -> Result<(), HostError> {
+        let contract_id = self.hash_from_obj_input("contract_id", contract_id)?;
+        let storage_key = self.contract_code_ledger_key(contract_id.clone());
+        if self.0.storage.borrow_mut().has(&storage_key)? {
+            return Err(self.err_general("Contract already exists"));
+        }
+        self.store_contract_code(contract, contract_id, &storage_key)?;
+        Ok(())
+    }
+
+    pub fn create_contract_with_id_preimage(
         &self,
         contract: ScContractCode,
         id_preimage: Vec<u8>,
     ) -> Result<Object, HostError> {
         let id_obj = self.compute_hash_sha256(self.add_host_object(id_preimage)?.into())?;
-        let new_contract_id = self.hash_from_obj_input("id_obj", id_obj)?;
-        let storage_key = self.contract_code_ledger_key(new_contract_id.clone());
-        if self.0.storage.borrow_mut().has(&storage_key)? {
-            return Err(self.err_general("Contract already exists"));
-        }
-        self.store_contract_code(contract, new_contract_id, &storage_key)?;
+        self.create_contract_with_id(contract, id_obj)?;
         Ok(id_obj)
     }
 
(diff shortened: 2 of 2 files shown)


leighmcculloch:
See https://github.com/stellar/rs-soroban-sdk/pull/438 for an example of how this will be used.

jonjove:
I'm not sure I understand the motivation for this change. I wasn't able to find any examples that actually do what's described as

> Some examples in the soroban-examples repo have achieved this by creating the other contract with the Host::create_contract_from_contract function. The problem with this approach is that the contract being created is created as if it was created by the developers contract. This is conceptually not aligned with how a developer thinks about deploying the other contract, because a contract that a developer writes won't always be deploying the contract it is interacting with.

but I feel like I'm looking for the wrong thing because I'm not understanding the motivation yet. Can you point me to one? There are some contracts (eg. liquidity pool) that actually do need to deploy another contract, and there are some tests (eg. for liquidity_pool and single_offer) which deploy other contracts as test contracts but not using WASM.

There are other ways to deploy WASM contracts that don't depend on doing it from in a contract. The `create_contract_from_ed25519` host function exists, but isn't exposed in the SDK right now because it hasn't been needed. The main difference I see between what's in this PR and `create_contract_from_ed25519` is the ability to choose the exact ID that the deployed contract receives. I suppose this could be an important feature if you want a hard-coded address in your contract (whether this is a good idea or not, I'm not really sure). Is the ability to deploy at a specific address the main motivation for this change?

leighmcculloch:
> I wasn't able to find any examples that actually do what's described as
> 
> > Some examples in the soroban-examples repo have achieved this by creating the other contract with the `Host::create_contract_from_contract` function

I'm referring to this code where the liquidity pool contract deploys the token contract: https://github.com/stellar/soroban-examples/blob/04781cfadd9d6fc42dd80867e6fb2a8e5a06adda/liquidity_pool/src/token_contract.rs. I think this approach makes sense for the liquidity pool contract because it actually does as part of normal operation deploy contracts. Another contract I'm writing just invokes another contract and has no responsibility for deploying it.

> The main difference I see between what's in this PR and `create_contract_from_ed25519` is the ability to choose the exact ID that the deployed contract receives.

+1 another option would be to plumb through `Host::create_contract_from_ed25519`, but it seems odd to me for this to show up in the SDK given contracts can't use it, and it requires more code to use since a user needs an ed25519 key to be generated.

Since we already provide `Host::register_contract`, it feels more natural and simpler for us to also allow registering a wasm contract.

If we end up plumbing through `create_contract_from_ed25519` we could consider removing `register_contract_wasm`.

Does the above address your concerns?

jonjove:
> but it seems odd to me for this to show up in the SDK given contracts can't use it

A contract can create a contract on behalf of an ed25519 public key, it would just need to receive the signature as input.

> and it requires more code to use since a user needs an ed25519 key to be generated

This is definitely true, and as such it's not really convenient to deploy a contract this way. I'm fine with merging this, especially for the purpose of making it possible to choose the ID.

jonjove:
> I'm not too concerned about using a specific ID, unless of course a developer is hardcoding an ID in their code.

If this part doesn't seem relevant, it probably would have been better to add a wrapper method to the SDK that generates an ed25519 keypair and uses `create_contract_from_ed25519` to actually do the work. This would mean that functionality is syntactic sugar from the host's perspective instead of having specific host functionality to support it.

leighmcculloch:
I think it's relevant, just not how I ended up here. A developing hardcoding an ID would need this for testing, and have them jump through hoops relating to ed25519 feels irrelevant to most tests. Using randomly generated values inside a test can at times by inconvenient and get in the way for testing. I know @paulbellamy is also wrestling a little right now of how to build contract deployment into SDKs because the terms aren't intuitive. Even if we plumb through `create_contract_from_ed25519` it will be a testutils only function, and it will just serve to expose impl details. I think this is okay as is, but we can revisit it if we see it create problems?

Issue #322, opened 2022-08-11
Finish the contract_event / system_event host functions


graydon:
These host functions currently don't do anything -- they're `todo!()` -- and there's an empty stub in `events.rs` for the `ContractEvent` type. This all needs to be filled in so that the CLI and host embedding can report events to their callers.

graydon:
cc @jayz22 as you might want to take this also? I'll do it if you do not want it or are too busy.

jayz22:
Sure I'm happy to take it. Although, I probably won't make concrete progress on it until later next week as I'm currently still occupied by the budget metering.

PR #342, opened 2022-08-18
Support contract events

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/322

cc @paulbellamy 

### Why

### Known limitations

~Lacks testing.~ Added a native test.
Need to update the sdk (https://github.com/stellar/rs-soroban-sdk/issues/451) ~and write a test contract which records events.~


soroban-env-common/src/env.rs
```diff
@@ -146,8 +146,14 @@ macro_rules! call_macro_with_all_host_functions {
...
-                {"2", fn contract_event(v:RawVal) -> RawVal }
-                {"3", fn system_event(v:RawVal) -> RawVal }
+                /// Records a contract event. `topics` is expected to be a `SCVec` with
+                /// length <= 4 that cannot contain Vecs, Maps, or Binaries > 32 bytes
+                /// On succes, returns an `SCStatus::Ok`.
+                {"2", fn contract_event(topics:Object, data:RawVal) -> RawVal }
+                /// Records a system event. `topics` is expected to be a `SCVec` with
+                /// length <= 4 that cannot contain Vecs, Maps, or Binaries > 32 bytes
+                /// On succes, returns an `SCStatus::Ok`.
+                {"3", fn system_event(topics:Object, data:RawVal) -> RawVal }
```
leighmcculloch, line 156: ❓ I thought system_event was an internal function, not a function that we'd let a contract call. Is that right?

soroban-env-host/src/host.rs
```diff
@@ -238,6 +260,12 @@ impl Host {
             .map_err(Host)
     }
 
+    pub fn recover_events(self) -> Result<Events, Self> {
```
leighmcculloch, line 263: 💡 What's the significance of the term `recover` here? Is the idea that this takes the events out of the host that have been queued? I normally see the term recover used in situations where failure has occurred. Is this for a failure situation?

The term `collect` may carry less other meanings than recover.

soroban-env-host/src/host/conversion.rs
```diff
@@ -150,4 +155,87 @@ impl Host {
...
+                                Some(ho) => match ho {
+                                    // TODO: use more event-specific error codes than `UnexpectedType`
+                                    HostObject::Vec(_) => {
+                                        Err(self.err_status(ScHostObjErrorCode::UnexpectedType))
+                                    }
+                                    HostObject::Map(_) => {
+                                        Err(self.err_status(ScHostObjErrorCode::UnexpectedType))
+                                    }
+                                    HostObject::Bin(b) => {
+                                        if b.len() > 32 {
```
leighmcculloch, line 194: 💡 I think it might be worth giving a name to this 32, by moving it into a constant. It's a rather magic number.

soroban-env-host/src/host/conversion.rs
```diff
@@ -150,4 +155,87 @@ impl Host {
...
+                                            return Err(
+                                                self.err_status(ScHostObjErrorCode::UnexpectedType)
```
leighmcculloch, line 196: ❗ This error doesn't indicate that the size is wrong. The unexpected type error may be misleading since the caller is passing the correct type.

soroban-env-host/src/host/conversion.rs
```diff
@@ -150,4 +155,87 @@ impl Host {
...
+
+    pub(crate) fn event_topics_from_host_obj(&self, topics: Object) -> Result<ScVec, HostError> {
+        unsafe {
+            self.unchecked_visit_val_obj(topics.into(), |ob| {
+                self.charge_budget(CostType::ValXdrConv, 1)?;
+                match ob {
+                    None => Err(self.err_status(ScHostObjErrorCode::UnknownReference)),
+                    Some(ho) => match ho {
+                        HostObject::Vec(vv) => {
+                            if vv.len() > 4 {
```
leighmcculloch, line 225: 💡 Same here, this value would benefit from being in a constant.

leighmcculloch (commented):
Couple of concerns (❓❗), otherwise LGTM.

I also left some suggestions (💡), but defer to you on them.

soroban-env-host/src/host/conversion.rs
jayz22: Fixed. Thank you.

soroban-env-host/src/host/conversion.rs
jayz22: Good point. I added a TODO comment above it to replace this with a more descriptive error code.

soroban-env-host/src/host/conversion.rs
jayz22: Fixed.

soroban-env-host/src/host.rs
jayz22: Yeah currently the way it's being used (e.g. in [core](https://github.com/stellar/stellar-core/blob/master/src/rust/src/contract.rs#L184)) fits `collect` or `retrieve` better. This name follows the same as `recover_storage` above. Maybe @graydon has a better perspective on how these two functions are intended to be used (e.g. if there is a failure recover part involved).

soroban-env-common/src/env.rs
jayz22: You are right, I missed it from reading the CAP. 
I've fixed it by moving the `system_event` function internal. However, this will fail if called within the `HostFunction` context, i.e. invoked from stellar-core. Added a new issue: https://github.com/stellar/rs-soroban-env/issues/345.

graydon (commented):
I pushed a change with a revised interface for finalizing a host and getting its events, budget and storage back -- there can only be one such call since if it succeeds it will destroy the underlying HostImpl.

Also think this needs a refresh to handle the optional contract ID issue. Will coordinate with @sisuresh and update this PR to match.

soroban-env-host/src/host.rs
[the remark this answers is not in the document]
jayz22: Thank you, I completely missed that! Makes total sense now.

graydon: replaced with `try_finish`. The idea here is to have a function that deconstructs a host and returns ownership of its contained objects, efficiently (without cloning them). It's a fairly normal pattern in Rust code when you're done with a thing, and we use it to convey information back to the contract-invoker in core.

graydon (approved):

PR #354, opened 2022-08-20
Move WASM tests and inputs to env repo, fix #329

### LINKED ISSUE #329: Copy/move host test vector contracts from SDK repo to this repo
It appears that the stellar-env-host crate has tests that depend on the wasm outputs of the SDK test contracts. 

This is not ideal because this has caused at times for us to have test contracts in the SDK repo that we're bespoke for a specific host test and otherwise made little sense to SDK developers or people who work only in the SDK repo.

We should keep test vectors for the host near the host. I think we should move or copy the test contracts that the host relies on from the SDK repo to the host repo.  

Advantages of doing this:
- host test vectors are changed on a cadence that makes sense for the core team.
- the host can maintain test vectors that use different versions of the SDK.

cc @graydon @jayz22 @jonjove @sisuresh

graydon:
This moves the WASM test inputs and machinery for regenerating them into a new (non-published) crate in the env repo. We can also hopefully reference-and-use these constants from core. It's not perfect and uses a bit more Makefile and a bit less build.rs than maybe an ideal implementation, but it works ok for now and gets us out of the business of hacking up and copying stuff out of the SDK repo.

Fixes #329


soroban-test-wasms/wasm-workspace/Cargo.toml
```diff
@@ -0,0 +1,64 @@
+# This file defines a workspace that is _periodically_ recompiled as part of
```
leighmcculloch, line 1: I don't have a concern about this, but I'm asking to learn:

What's the motivation for having these live in a separate workspace?

The SDK has its test vectors in the same workspace, even the test vectors that SDK tests rely on for .wasm files. In the SDK to make sure .wasm files always exist for tests, we just build them for wasm32 before running tests. For example: https://github.com/stellar/rs-soroban-sdk/blob/aa61c62a01cd7a8ba4297889945a7ddd68b46aaf/Makefile#L12-L18

graydon: A few reasons, each of which may be surmountable -- patches welcome!

  - I wanted them to inherit wasm release-build profile settings, and the outer workspace has different settings
  - I wanted it to be a separate build step (not triggered by an outer-workspace cargo build) and commit the binaries as versioned inputs, so the outer `soroban-test-wasms` crate could be usable from other crates & repos (especially: core) via cargo, without having to run cross-repo make / rebuild them on the fly
  - I wasn't sure what would happen if I patched `soroban-env-common` to point to local _within_ the workspace defining it as local (and I need to patch in order to include the SDK as a dev-dep of the WASMs). Figured keeping the two resolution runs separate would make it easier to be sure about the meaning of things.

I'm happy to iterate if you think there's a cleaner way / think none of these are real issues.

leighmcculloch: I don't feel strongly towards one way or the other. I don't know what the short comings of how I've set it up in the SDK yet. I suspect there are some, maybe lots 😬. Let's see how they play out.

soroban-test-wasms/wasm-workspace/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+# You should actually be able to set this to be any existing version of wasmi because we are
+# not building the SDK in host mode, so the "vm" feature is off.
+wasmi = { package = "soroban-wasmi", git = "https://github.com/stellar/wasmi", rev = "7b1f2355" }
+
+# The following SDK patch lines must denote an SDK version that is _compatible_
+# with the adjacent version of soroban-env-guest, selected above. If you are
+# making a change to soroban-env-common and/or soroban-env-guest, and are lucky,
+# you can modify soroban-env-common and/or soroban-env-guest _without_ having to
+# bump this rev, and can just recompile the WASMs here and embed them in your
+# change to the env repo.
```
jayz22, line 41: A few clarification / questions: 
1. Any change to the `soroban-env-host` alone does not require recompilation of the wasm files, only changes touching the `-common` and `-guest` does?  Thus "change to the env repo" here does not include change to `soroban-env-host`?
2. How does this relate to the `interface_version` in `meta.rs`? The `interface_version` specifies version of `soroban_env_common::env`, thus is a narrower restriction? I.e. Change to the `Env` requires bumping the `interface_version` which requires recompiling wasm files, but changes in the  `-common` and `-guest` without touching the `Env` may not need to recompile? 
3. The `sdk` also specifies the commit hash of the env repo: https://github.com/stellar/rs-soroban-sdk/blob/main/Cargo.toml#L45-L49. When recompiling with local sdk, do I need to change these lines to point to the local env repo?

jayz22 (commented):
Thank you very much for these changes! Just a few clarification / questions. I also think this benefits a little more explanation of when the wasm files need to ("have to" vs. "it's a good practice to but not required") be recompiled, since their regeneration is on-demand/not fully automated.

soroban-test-wasms/wasm-workspace/Cargo.toml
graydon: 1. Correct, these tests are built into WASM, which does not see the host crate at all; it only knows about the guest crate.
2. The interface version number should be changed anytime _anything_ that invalidates the interface between guest and host happens; changes to the Env trait are a minimal example of that, but (as a different example) suppose the bit-packing of RawVal changed, or the enum values of SCVal -- these would represent a "break to the interface" just as surely, and would therefore need to bump the interface version number. We need to be conservative about the interface version number, and bump it _whenever_ we have reason to believe there might be breakage. In terms of how this affects recompiling tests: a bump to the interface version necessarily means you have to recompile the tests; they won't run if the interface version compiled-in to them differs from the version supported on the host. The bigger question is whether the tests actually still compile at all, or whether they need to be updated (and/or the SDK needs to be updated). There is a sequence of things to ensure here, but the tests should either fail to compile, or fail to run, or both if there is a step missed.
3. No, the whole point of this change is that you do not need to use a local SDK to recompile at all. We're referring to a _remote_ SDK (by published-on-github git hash) and hoping it's still compatible with the local, updated env-common / env-guest (which we're injecting into its dependency tree, overriding the SDK's own `rev=` spec for the env repo).

soroban-test-wasms/wasm-workspace/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+#
+# The SDK typically refers to things in the env-guest crate by name and the
+# names of things don't change that often, so you will usually be lucky.
+#
+# If you are unlucky and your changes to env-common and/or env-guest require
+# changes to the SDK as well, not just a recompile, then you need to redirect
+# the following patch lines to point to a local checkout of the SDK that you've
+# modified to be compatible with the env-common and/or env-guest changes you're
+# making, then rebuild, embed the rebuilt wasms in your PR against the env repo,
+# merge it, and then open a PR against the SDK repo with the SDK changes.
```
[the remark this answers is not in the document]
graydon: No, this is only necessary in the "you got unlucky" case. The point here is that the SDK _shouldn't_ need to be updated for lots of changes; it can change later, when it needs to use the new functionality or otherwise catch up with guest changes, but not immediately. At least when you are lucky. When you're lucky, if the SDK compiles against the guest before your change, and your change is additive / doesn't radically break the SDK's assumptions, then the _same SDK version_ should compile against the guest with your change.

soroban-test-wasms/wasm-workspace/Cargo.toml
graydon: (A thing to understand here is that this comment and the crate references the file is defining are within a `[patch.crates-io]` section of this workspace: a section that declares _overrides_ of all the transitive crate references of everything in the workspace; this is a feature that exists specifically for this sort of "build X but with my own override for its dependency on Y" scenario)

graydon:
@jayz22 I added more commentary on this in https://github.com/stellar/rs-soroban-env/pull/361 -- let me know if it's still unclear

jayz22:
> @jayz22 I added more commentary on this in https://github.com/stellar/rs-soroban-env/pull/361 -- let me know if it's still unclear

I recently made a few compatibility breaking changes and took this new process to a test drive. 
Overall I think it is quite useful in catching compatibility breaks in both directions and instruction here are well explained and easy to follow. So thank you again @graydon!
The one thing I found slightly tricky is the `stellar-xdr` dependency. The outer workspace and the sdk both depends on it and might be using different versions. [Here](https://github.com/stellar/rs-soroban-env/blob/main/soroban-test-wasms/wasm-workspace/Cargo.toml#L35-L36) says "update this rev to the XDR rev used by the outer workspace" but this may not work if the sdk depends on a xdr version that is lower but incompatible (one example is the outer xdr version has new field to a enum which the sdk has not yet incorporated, so it would not build). So should this instead say something in the line of "the highest compatible version between the outer workspace and the sdk"?

leighmcculloch:
> The one thing I found slightly tricky is the `stellar-xdr` dependency. The outer workspace and the sdk both depends on it and might be using different versions.

It's worse than this. The CLI, and proc-macros all reference the stellar-xdr as well. It's becoming tricky to make sure they're all on the same version. We might be nearing the limit of how well we can manage this complex web of dependencies outside of a mono repo. We can do it, but it is taking a toll.

jayz22:
To also mention, when the sdk does break the host wasm files (which is expected right now), it was not very straightforward to debug (besides looking at which contract got broken and going though the sdk's commit history). It is fine this time since the breakage was small, but for bigger ones in the future (e.g. if they go out of sync for longer period of time) it might be more troublesome. cc @MonsieurNicolas since we just chatted about this.

PR #357, opened 2022-08-22
Replace TryFrom<EnvVal<>> with TryFromVal<>

leighmcculloch:
### What

Replace TryFrom<EnvVal<>> with TryFromVal<>, and disconnect the link between TryFrom and TryFromVal.

### Why

Disconnecting TryFrom and TryFromVal is necessary because we otherwise can't provide an implementation for all Option<T> for conversion in. This is because the core lib has an impl of T to Option<T> for all T.

TryFromVal is more of a serialization interface than it is a type conversion interface, even though we present it as the latter. Given that our interface has unique and specific behavior for all types, even builtin types like Option<T>, I think we need to have it standalone and not utilize the builtin From/TryFrom traits.

There is an alternative, we could manually or with macros implement every Option<T> interface we need, but it feels like philosophically we shouldn't be saying TryFromVal should always be bound to TryFrom. And after making this change I gotta say it simplifies things to take away some of these blanket implementations which is really all this change is doing.

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-common/src/env_val.rs
+++ b/soroban-env-common/src/env_val.rs
@@ -83,15 +83,6 @@ pub trait IntoVal<E: Env, V>: Sized {
     }
 }
 
-impl<E: Env, F, T> IntoVal<E, T> for F
-where
-    F: Into<T>,
-{
-    fn into_val(self, _: &E) -> T {
-        self.into()
-    }
-}
-
 pub trait TryIntoVal<E: Env, V>: Sized {
     type Error;
     fn try_into_val(self, env: &E) -> Result<V, Self::Error>;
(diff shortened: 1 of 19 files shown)


soroban-env-common/src/env_val.rs
```diff
@@ -104,62 +95,36 @@ pub trait TryIntoVal<E: Env, V>: Sized {
...
-{
-    type Error = T::Error;
-    fn try_from_val(env: &E, v: V) -> Result<Self, Self::Error> {
-        Self::try_from(EnvVal {
-            env: env.clone(),
-            val: v,
-        })
-    }
-}
-
```
leighmcculloch, line 135: This is the crux of this change.

leighmcculloch:
This PR is ready, but I'm making the same changes to the SDK so I can land both at the same time and validate that this is actually an okay change from the SDKs POV.

PR #358, opened 2022-08-22
Clarify comment about meta version, fix #328

### LINKED ISSUE #328: Meta interface comment should match code

The comment in this file describes functionality that doesn't exist.

The comment and functionality should match.

N/A

graydon:
This attempts to clarify the relationship between the meta version declaration and the code that checks it, as well as making it an explicit 1-point inclusive range with a "contains" check, and describing what will happen in the future in more detail.

Fixes #328


jonjove (approved):
This really helps clarify what a developer needs to do in this file, thank you!

PR #363, opened 2022-08-23
Add to object conversion for tuple.

jayz22:
### What

Tuples are already implemented as `Vec` and converts `RawVal`. By definition, this should also convert to `Object` (since `Vec` is an object). 

### Why

Needed for https://github.com/stellar/rs-soroban-sdk/pull/469 

### Known limitations

[TODO or N/A]


graydon (commented):
Looks ok to me but I think @leighmcculloch is more likely to have strong feelings about how conversions should be set up. I know he's been working on this space recently.

soroban-env-common/src/tuple.rs
```diff
@@ -32,15 +32,25 @@ macro_rules! impl_for_tuple {
             }
         }
 
-        impl<E: Env, $($typ),*> IntoVal<E, RawVal> for ($($typ,)*)
+        impl<E: Env, $($typ),*> IntoVal<E, Object> for ($($typ,)*)
```
leighmcculloch, line 35: Can you also implement `TryFromVal<E, Object>`? The RawVal version is just above this function, and these implementations accompany each other.

soroban-env-common/src/tuple.rs
```diff
@@ -57,6 +67,19 @@ macro_rules! impl_for_tuple {
...
+            fn into_val(self, env: &E) -> EnvVal<E, Object> {
+                let rv: Object = self.into_val(env);
+                EnvVal{
+                    env: env.clone(),
+                    val: rv,
+                }
+            }
+        }
+
         impl<E: Env, $($typ),*> TryIntoVal<E, ($($typ,)*)> for RawVal
```
leighmcculloch, line 83: Can you also add an implementation of this `for Object`? TryIntoVal for the RawVal type accompanies the TryFromVal at the top of the file (I guess these should be side-by-side).

leighmcculloch (commented):
👍🏻 Looks good, and sounds fine. We don't have Object conversions everywhere, but if it is convenient I think it's fine to add.

There are two trait impls missing for going the other way, could you add them for symmetry?

jayz22:
Turns out this is not needed with the solution adopted in https://github.com/stellar/rs-soroban-sdk/pull/469#discussion_r953242174

PR #367, opened 2022-08-25
Undo cyclic dev-dependency that is confusing rust-analyzer

graydon:
c533b3f9b160e33b18adc460275ab6e3f459625d introduced a cyclic dev-dependency so common could do some tests via host, which has been making rust-analyzer very sad -- this is something cargo apparently supports but only in not-very-reliable ways that are often confusing and apparently confuse r-a.

Easy fix: move the tests a layer down to host.


jayz22:
:tada::tada::tada: Thank you @graydon, you are a true life saver :sob:

leighmcculloch:
Was this causing rust-analyzer to break or was it just a warning. Whilst I saw the warning too it didn't seem to impact rust-analyzer for me, but I might not have noticed the impact. 

Related issues on rust-analyzer appear to be:
- https://github.com/rust-lang/rust-analyzer/issues/2414
- https://github.com/rust-lang/rust-analyzer/issues/8330

The inconvenience of this change is that tests in the host crate will with the CODEOWNERS definition require review from the core team, so now all changes that are tested in common will require review from the core team. 

Maybe we should instead have these in a `soroban-env-common-tests` crate that isn't published?

leighmcculloch:
> Maybe we should instead have these in a `soroban-env-common-tests` crate that isn't published?

I don't think we should hastily spend time on this, but it seems like our crate setup is a bit artificial because:
- some things in common can only be tested in SDK 
- some things in common can only be tested in host

graydon:
@leighmcculloch it was causing r-a to emit hard errors, not warnings, and fail to resolve / typecheck / allow navigation of a substantial fraction of imported types (all definitions imported from common). It was fairly unusable for editing the host crate.

Concerning the permission-to-edit-tests-in-the-host issue: I have no problem with any of the following solutions:
- making the soroban-env-host/tests directory more widely co-owned (CODEOWNERS allows overrides on subdirs)
- adding a special co-owned subdir of tests
- adding a new unpublished testing crate
- moving these tests to the SDK so they're under your purview

leighmcculloch:
Got it. All good ideas I think. Let's not move these tests to the SDK though, that just means this code can break here and be found to be broken only when the SDK is updated.

I'm leaning towards embracing the integration test mindset and having a special co-owned subdir of tests, but have them in a separate crate which is like the "here lives the integration tests" so that tests in the host crate are clearly more focused on unit testing the host itself.

leighmcculloch:
Thanks for fixing this.

PR #368, opened 2022-08-25
Flesh out support for ContractCode, Hash and PublicKey.

graydon:
While we're still discussing what to do with Hash, PublicKey and ContractCode, we've actually started _using_ them in some code paths -- the recent native contract stuff in particular -- and therefore we need to be able to interconvert for basic tests to continue to function (eg. the contract_data unit test in core, which forms a ContractCode LE).


jonjove:
We've actually only really started using `ContractCode` in native contracts. I stand by the belief that the other ones shouldn't exist.

leighmcculloch:
At the moment the SDK isn't using Hash, PublicKey, ContractCode, Status, or BitSet.

The one in this list we probably need to start using is Status, the others less clear.

PR #371, opened 2022-08-25
Check if rust-analyzer can parse project

leighmcculloch:
### What
Add ci job that checks if rust-analyzer can parse project.

### Why
A couple times now we have done things that are technically allowed in Rust but are not supported by rust-analyzer. Everytime we've walked the change back because rust-analyzer compatibility is important for developer productivity. Let's prevent these things from happening by checking if rust-analyzer can parse the project.


leighmcculloch:
If someone introduces a cyclic dependency that rust-analyzer cannot resolve, we'll see this:
Screen Shot 2022-08-25 at 9 44 03 AM

Screen Shot 2022-08-25 at 9 43 32 AM

jayz22:
🎉 Thank you for making this change @leighmcculloch !

graydon:
Yeah this is rad, thanks!

PR #375, opened 2022-08-26
Add base64 feature to xdr import

paulbellamy:
### What

Add `base64` feature to xdr import.

### Why

So we can save a few characters when encoding/decoding to base64.

### DIFF
--- a/soroban-env-common/Cargo.toml
+++ b/soroban-env-common/Cargo.toml
@@ -12,12 +12,12 @@ rust-version = "1.63"
 
 [dependencies]
 soroban-env-macros = { version = "0.0.3" }
-stellar-xdr = { version = "0.0.1", default-features = false, features = [ "next", "base64" ] }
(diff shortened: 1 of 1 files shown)


soroban-env-common/Cargo.toml
```diff
@@ -12,7 +12,7 @@ rust-version = "1.63"
 
 [dependencies]
 soroban-env-macros = { version = "0.0.3" }
-stellar-xdr = { version = "0.0.1", default-features = false, features = [ "next" ] }
+stellar-xdr = { version = "0.0.1", default-features = false, features = [ "next", "base64" ] }
```
leighmcculloch, line 15: Can you make it so the base64 feature of only activated when the std feature is enabled? Change a few lines down to this:

```toml
std = ["stellar-xdr/std", "stellar-xdr/base64"]
```

PR #380, opened 2022-08-26
Add hostfns to construct BigInt from bytes/array of digits

jayz22:
### What

Host side support for https://github.com/stellar/rs-soroban-sdk/issues/493

### Why

### Known limitations

[TODO or N/A]


leighmcculloch (approved):

soroban-env-host/src/host.rs
```diff
@@ -1588,6 +1588,35 @@ impl CheckedEnv for Host {
...
+    fn bigint_from_radix_be(
+        &self,
+        sign: RawVal,
+        buf: Object,
+        radix: RawVal,
+    ) -> Result<Object, Self::Error> {
+        let s = self.bigint_sign_from_rawval(sign)?;
+        let r = self.bigint_radix_from_rawval(radix)?;
+        let res = self.visit_obj(buf, |b: &Vec<u8>| {
+            if r != 256 && b.iter().any(|&b| b >= r as u8) {
```
leighmcculloch, line 1608: What's the significant of r != 256? Does this mean the radix has to be 256?

leighmcculloch: Oh I see, this check only occurs when the radix is < 256. There's no check here that `r` isn't > 256.

jayz22: This is just to skip the comparison for slight performance improvement. Since in `bigint_radix_from_rawval` we made sure that `r` is <= 256.

leighmcculloch: Oh, I'm wrong, the check happens inside `bigint_radix_from_rawval`.

soroban-env-host/src/host.rs
```diff
@@ -1588,6 +1588,35 @@ impl CheckedEnv for Host {
...
+    }
+
+    fn bigint_from_radix_be(
+        &self,
+        sign: RawVal,
+        buf: Object,
+        radix: RawVal,
+    ) -> Result<Object, Self::Error> {
+        let s = self.bigint_sign_from_rawval(sign)?;
+        let r = self.bigint_radix_from_rawval(radix)?;
```
leighmcculloch, line 1606: I like the way this parsing happens with functions dedicated to the type of value being accepted. 🎉

Issue #317, opened 2022-08-08
Type: Vec: Add fns first_index_of, last_index_of, binary_search


leighmcculloch:
Add a `vec_first_index_of` host function that accepts a `Vec`, and a `RawVal`, and returns a u32 of the first index indicating if the `Vec` contains the value.

Add a `vec_last_index_of` host function that accepts a `Vec`, and a `RawVal`, and returns a u32 of the last index indicating if the `Vec` contains the value.

Add a `vec_binary_search` host function that accepts a `Vec` assumed to be sorted, and a `RawVal`, and returns a u32 index of the value in the `Vec`. If the item cannot be found, returns a u32 of the index of where the value could be inserted. Unclear how we should distinguish these two u32 values, maybe they should be high/low of a u64. (cc @graydon)

cc @jayz22

jayz22:
> Unclear how we should distinguish these two u32 values, maybe they should be high/low of a u64. (cc @graydon)

I like this idea, however, we can't disambiguate the `u64::ZERO` (`0u32|0u32` ) case with this approach. 
I'm going to go with a tuple of  `(RawVal, RawVal)` as the return type, where one of them will be a `Static::Void` and the other one contains the index. I.e.:

```
match res => {
(idx, Status::Void) => !todo() // "found value at idx"
(Status::Void, idx) => !todo() // "not found, correct place to insert is idx"
}
```

leighmcculloch:
The tuple will require a Vec and a lot more instructgions on the wasm side.

How about we make the return value positive if found, negative if not found?

leighmcculloch:
Oh right, as you say, can't disambiguate zero.

leighmcculloch:
Wait, how about we use high low in a different way?

The low can always be the index. The high can be 0x0 if found, and 0x1 if not.

jayz22:
Ah yes, that can work!

PR #387, opened 2022-08-29
Add hostfns to find an element in a vec

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/317.
- `vec_first_index_of`
- `vec_last_index_of`
- `vec_binary_search`

### Why

### Known limitations

[TODO or N/A]


leighmcculloch (approved):

soroban-env-host/src/host/conversion.rs
```diff
@@ -294,4 +294,27 @@ impl Host {
...
         }
     }
+
+    /// Converts a binary search result into a u64. `res` is `Some(index)`
+    /// if the value was found at `index`, or `Err(index)` if the value was not found
+    /// and would've needed to be inserted at `index`.
+    /// Returns a Some(res_u64) where :
+    /// - the high-32 bits is 0x0001 if element existed or 0x0000 if it didn't
+    /// - the low-32 bits contains the u32 representation of the `index`
+    /// Err(_) if the `index` fails to be converted to an u32.
```
jayz22, line 304: I hope the documentation of the u64 return value here is good enough. It does feel a bit ad-hoc compared to the standard `Val` encoding we currently have.

leighmcculloch: I agree, but there's nothing here we can't change. @graydon will probably have some good ideas for how we can revisit it.

Issue #341, opened 2022-08-18
Should vec_push_front / vec_pop_front host functions exist?


jonjove:
These functions don't exist right now, but are actually as efficient as `vec_push_back` / `vec_pop_back` in `im_rc`. For context, see [`im_rc::Vector::push_front`](https://docs.rs/im-rc/latest/im_rc/struct.Vector.html#method.push_front).

leighmcculloch:
If they're efficient yup. +1

jayz22:
+1. I can add these.

PR #388, opened 2022-08-29
Add vec_push_front and vec_push_back

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/341.
Also renamed `vec_push` --> `vec_push_back` and `vec_pop` --> `vec_pop_back` and re-indexed to group the "push" and "pop" methods together. I know we discourage renaming/reindexing host functions in general as they are likely go through more iterations. But this change seems like an obvious one that improves clarity. Plus the recent work by @graydon on wasm test file restructuring makes it easy to recompile and update the `.wasm` files so they do not go stale. 
I will make the follow up change to update the SDK side once this is merged.

### Why

### Known limitations

[TODO or N/A]


leighmcculloch (approved):
Changing the host function values might be problematic ordinarily, but since right now we require a matching env version, it should be fine and have no consequence. Thanks! I'll plumb these into the SDK.

PR #394, opened 2022-08-30
Make token auth match soroban-sdk-auth

jonjove:
### What

Make token auth match soroban-sdk-auth.

### Why

It was completely impossible to use the native token without this because the necessary types weren't exported anywhere.

### Known limitations

Lightly tested.

### DIFF
--- a/soroban-env-host/src/native_contract/token/public_types.rs
+++ b/soroban-env-host/src/native_contract/token/public_types.rs
@@ -1,72 +1,62 @@
 use crate::host::Host;
-use crate::native_contract::base_types::{BigInt, BytesN, Map, Vec};
+use crate::native_contract::base_types::{Bytes, BytesN, Map, Vec};
 use crate::native_contract::token::error::Error;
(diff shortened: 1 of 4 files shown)


soroban-env-host/src/native_contract/token/public_types.rs
```diff
@@ -1,72 +1,62 @@
...
 
-pub type U256 = BytesN<32>;
-pub type U512 = BytesN<64>;
-
 #[derive(Clone)]
-#[contracttype]
-pub struct KeyedEd25519Signature {
-    pub public_key: U256,
-    pub signature: U512,
+#[contracttype(lib = "soroban_sdk_auth")]
```
leighmcculloch, line 8: Does the native sdk use the lib field?

jonjove: Nope, that was a copy-paste error when moving from soroban-sdk-auth. I'll fix this shortly.

PR #402, opened 2022-09-02
Minor fixes

graydon:
This just fixes a recent break and exports some types that need to be exported for the host to be used from core.

### DIFF
--- a/soroban-env-host/src/lib.rs
+++ b/soroban-env-host/src/lib.rs
@@ -46,8 +46,6 @@ mod test;
 
 #[cfg(feature = "testutils")]
 pub use host::ContractFunctionSet;
-#[cfg(feature = "testutils")]
-pub use host::LedgerInfo;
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):

soroban-env-host/src/lib.rs
```diff
@@ -46,8 +46,6 @@ mod test;
 
 #[cfg(feature = "testutils")]
 pub use host::ContractFunctionSet;
-#[cfg(feature = "testutils")]
-pub use host::LedgerInfo;
-pub use host::{Host, HostError};
+pub use host::{Host, HostError, LedgerInfo, metered_map::MeteredOrdMap};
```
leighmcculloch, line 49: LedgerInfo needs to be in testutils I think?

graydon: As far as I know it needs to be always available, since it will be populated on production networks.

PR #422, opened 2022-09-09
Migrate to wasmi_v1 API

graydon:
This upgrades the env crates to use the new wasmi_v1 API that arrives with [the upgrade to wasmi 0.16.0](https://github.com/stellar/wasmi/pull/2).

Unfortunately it's quite invasive: every host function signature changes inside the host. The fallout _should_ be limited to the host (and people who embed and invoke it). Most of the changes have to do with the fact that wasmi now requires us to thread a `&mut Store` through to the host function if we want to use anything stored _in_ its store (such as linear memory). We can't dynamically acquire mutable access to it elsewhere, as it's already mutably borrowed during execution.

Everything else is mostly just "same task, different API" sort of stuff. Module parsing and dispatch functions actually got a little simpler.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,6 +277,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "indexmap-nostd"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


leighmcculloch (approved):
Nice!

PR #425, opened 2022-09-12
Add fail_with_status host function, fix #146

### LINKED ISSUE #146: Add host function for aborting current subtransaction with details / a status code
~~We need a way for a contract to exit the subtransaction / call that it's currently running immediately, without popping / returning all the way back out to its outermost frame. The moral equivalent of the `exit(int)` standard library function in C.~~

~~Pass void for success, or a status for failure -- or just a status if we're going to keep the "ok" / "success" status code (this might be the exception to the "never pass a status" rule).~~

We should allow aborting the current subtransaction with _some_ details -- eg. calling `env.abort(status)` -- rather than just calling panic!(). This might include revising the status type to include a variant that denotes its file name / line number / a user-provided message as in #205.

graydon:
Addresses #146 by providing a fail_with_status function that users can call to fail the current contract call (trapping its vm) with a user-provided status. The status has to be the newly-added `ScStatusType::ContractError` with an opaque user-provided u32 body.


graydon:
cc @jonjove and @leighmcculloch I am posting this to make discussion concrete -- we shouldn't merge this until we have consensus that it's the right thing to do.

soroban-env-common/src/env.rs
```diff
@@ -168,6 +168,10 @@ macro_rules! call_macro_with_all_host_functions {
                 /// vector contains the contract id as Hash, and a function as
                 /// a Symbol.
                 {"8", fn get_current_call_stack() -> Object }
+                /// Causes the currently executing contract to fail immediately
+                /// with a provided status code, which must be of error-type
+                /// `ScStatusType::ContractError`. Does not actually return.
+                {"9", fn fail_with_status(status:Status) -> RawVal }
```
leighmcculloch, line 174: I'm wondering if fail is the best term for us to use. Maybe we should use `abort_with_`, since what will happen is equivalent to aborting? This isn't particularly important and it is fine, and maybe preferable, as is.

leighmcculloch (approved):
👍🏻 Looks good to me.

Let's continue to discuss whether we should do this in https://github.com/stellar/stellar-xdr-next/issues/28.

soroban-env-host/src/host.rs
```diff
@@ -2400,4 +2400,19 @@ impl VmCallerCheckedEnv for Host {
...
         Ok(self.add_host_object(res)?.into())
     }
+
+    fn fail_with_status(
+        &self,
+        vmcaller: &mut VmCaller<Self::VmUserState>,
+        status: Status,
+    ) -> Result<RawVal, Self::Error> {
+        if status.is_type(ScStatusType::ContractError) {
+            Err(self.err_status_msg(status, "failing with contract-provided status code"))
```
leighmcculloch, line 2410: @graydon If the currently executing contract is being invoked by another contract, will that other contract get the status as the result of the invocation? Or will the entire stack of invocations be aborted?

graydon: If A calls B using call, and B fails, A will fail.

If A calls B using try_call, and B fails, A will get the status result.

This extrapolates to the call chain. All calls-of-calls will fail unless there's a try_call somewhere along the chain.

leighmcculloch: Makes sense. This made me realize we do not actually have try_call wired up in the SDK or in the generated clients. I'm going to go plumb that through.

soroban-env-host/src/host.rs
```diff
@@ -2516,4 +2516,19 @@ impl VmCallerCheckedEnv for Host {
...
```
leighmcculloch, line 2526: Lets use the term `contract error status code`, because the status type is contract error.
```suggestion
            Err(self.err_status_msg(status, "failing with contract error status code"))
```

leighmcculloch (approved):
One word change in error text, otherwise LGTM.

Also, one build is failing.

soroban-env-host/src/host.rs
graydon: integrated with rebase

PR #427, opened 2022-09-13
Add invocation error debug event to call like try_call

leighmcculloch:
### What
Add invocation error debug event to `Env::call` like `Env::try_call`.

### Why
A developer will choose call or try_call based on how they want to handle an error. In both cases diagnostics about the error are useful, and it is confusing that the diagnostic events are inconsistent. The fact the debug events are different implies something different has taken place inside the called contract, when the calls are the same.

### DIFF
--- a/soroban-env-host/src/test/invocation.rs
+++ b/soroban-env-host/src/test/invocation.rs
@@ -2,8 +2,9 @@ use soroban_env_common::{xdr::ScVmErrorCode, RawVal};
 
 use crate::{
     budget::Budget,
+    events::{DebugArg, HostEvent},
     vm::Vm,
-    xdr::{Hash, ScHostObjErrorCode, ScVal, ScVec},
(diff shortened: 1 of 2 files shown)


soroban-env-host/src/test/invocation.rs
```diff
@@ -70,10 +71,60 @@ fn invoke_cross_contract_with_err() -> Result<(), HostError> {
...
+                    )
+                );
+            } else {
+                panic!("got: {:?}, want debug arg containing Val", de);
+            }
+        }
+        _ => {
+            panic!("got: {:?}, want debug event", last_event);
+        }
+    };
```
leighmcculloch, line 97: @graydon FYI asserting on DebugEvents is pretty painful because there's a `RawVal` embedded in there, and we can't do equality checks on `RawVal`s.

PR #429, opened 2022-09-13
Update stellar-xdr

leighmcculloch:
### What
Update stellar-xdr.

### Why
To get the latest XDR which has updated.


soroban-env-common/src/status.rs
```diff
@@ -274,22 +275,20 @@ impl Status {
...
-            ScStatus::VmError(code) => code as i32,
-            ScStatus::UnknownError(code) => code as i32,
+            ScStatus::HostContextError(code) => code as i32 as u32,
+            ScStatus::HostValueError(code) => code as i32 as u32,
+            ScStatus::HostObjectError(code) => code as i32 as u32,
+            ScStatus::HostFunctionError(code) => code as i32 as u32,
+            ScStatus::HostStorageError(code) => code as i32 as u32,
+            ScStatus::VmError(code) => code as i32 as u32,
+            ScStatus::UnknownError(code) => code as i32 as u32,
+            ScStatus::ContractError(code) => code as u32,
```
leighmcculloch, line 289: @graydon I'm not sure if there's any value in noting the `as i32` still here. The codes are i32.

Issue #296, opened 2022-08-01
Cost budget and metering: introduce reasonable host budget parameters and limits to prevent DDOS.


jayz22:

smac711:
This is just the DOS prevention.

PR #443, opened 2022-09-16
Add comments to each `CostType`; add very crude, untested model parameters

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/296 by putting in some cost parameters and budget.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/budget.rs
+++ b/soroban-env-host/src/budget.rs
@@ -9,68 +9,128 @@ use crate::{xdr::ScVmErrorCode, HostError};
 #[repr(i32)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum CostType {
+    // Cost of running 1 wasm interpreter loop
     WasmInsnExec = 0,
     WasmMemAlloc = 1,
+    // Cost of forming a debug event and pushing it into the event storage
     HostEventDebug = 2,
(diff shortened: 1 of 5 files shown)


soroban-env-host/src/budget.rs
```diff
@@ -143,6 +203,134 @@ impl CostType {
     }
 }
 
+pub const PARAMS_CPU: &'static [[u64; 5]; 60] = &[
```
graydon, line 206: I think having this big table maintained in un-checked alignment with the CostType enum is a recipe for maintenance-induced errors. Can we change it so that you have a big function that matches on `CostType` and returns a `CostModel` in each arm of the match? Then you will get the match-expression exhaustiveness-checker checking that you cover all cases and that each case matches the thing you think it should (i.e. no need for all these comments to be here, much less be lined up correctly)

graydon (changes requested):

soroban-env-host/src/budget.rs
graydon: (returns a `CostModel` or modifies an existing already-retrieved `CostModel` or whatever .. just some pattern that uses the static enumeration of `CostType`s to our advantage)

graydon:
Given that you're away this week @jayz22 I just made the review edits I had in mind here myself and pushed them. Will merge now, thanks!

Issue #447, opened 2022-09-16
Support DebugEvents in WASM builds


leighmcculloch:
The SDK provides logging functionality that is only available in debug builds. That logging functionality creates DebugEvents by calling this function:

- `log_static_fmt_general `

That function isn't implemented for the Guest Env though, so when a contract is built using the dev profile the debug events are lost.

We should add support for debug events so that contract developers can see debug events / logs when using the CLI tool or when contract importing a dev profile built contract for testing.

leighmcculloch:
cc @graydon I'm not sure when we can schedule this. I had hoped to support this style of debug event logging for Meridian, but I failed to understand the current intended limitations with debug events. I understand if this is a post-Meridian thing.

graydon:
It's not so much a huge amount of work as something I want to make sure we do intentionally / carefully. This functionality is turned of for reasons, and turning it on / plumbing it through should address those reasons.

There are 3 considerations:

  - The lifetime of guest memory: We currently embed a full copy of the host's event buffer (at the time of error) in `HostError`, along with a stacktrace, when we generate a `HostError`, so that we can print it to the user when/if they get a `Debug` dump in a panic. `HostError` currently does not hold onto either a `Host` or a `Vm`, so any values in an event struct have no `Host` or `Vm` to "resolve" against. This is incidentally the same reason that a host-object reference in a debug error message, when you format it, does not resolve to a "full" object, but just an object-handle number wrapped in an object type code name. Because it's just a context-free `RawVal`, no idea what `Host` or `Vm` it should resolve against. They might even have been torn down by the time you're dumping the `Debug` view of the event. If we embed a guest linear-memory address in an event (as the address of a format string for a log message), we would just be making this worse. We can do it with host static strings only because they're constant over the life of the OS process -- they always outlive the host anyways. We could copy guest memory into a longer-lived value (eg. a dynamic String) but that would cost much more at event-logging time (doing a memcpy at least, if not a full string-format, of data we probably want to ignore/discard 99% of the time) and also be a new way for guests to fragment host memory. So we'd probably want to make this optional / only turned on in debug builds of the host or something.
  - Exposure to the string formatting library: we're using a _very_ simple dynamic string formatter that is probably safe to expose to user input, but it's still a new attack vector, and "format strings exposed to user input" is an old favorite channel for security vulnerabilities. We'd want to be very, very sure about it.
[...]

PR #464, opened 2022-09-22
Support logging debug events via the guest interface

leighmcculloch:
### What
Support logging debug events via the guest interface, gated behind the debug-assertions build flag.

### Why

Close https://github.com/stellar/rs-soroban-env/issues/447

~The reason I've used the debug-assertions build flag and not a feature is because the Rust reference describes the flag as something that can be used to gate extra debug code, and I think we'll use it in the SDK for debug logs since it is a pretty huge advantage to not have to ask every contract developer to layer a `debug` feature into their contract just to get logging. Having that consistency between SDK and env is convenient.~
~https://doc.rust-lang.org/stable/reference/conditional-compilation.html?highlight=debug_assertions#debug_assertions~

### Status

This is a WIP.


soroban-env-host/src/host.rs
```diff
@@ -1074,6 +1074,31 @@ impl VmCallerCheckedEnv for Host {
...
+        &self,
+        _vmcaller: &mut VmCaller<Host>,
+        fmt: Object,
+        args: Object,
+    ) -> Result<RawVal, HostError> {
+        if cfg!(debug_assertions) {
+            let fmt: String = self
+                .visit_obj(fmt, move |hv: &Vec<u8>| Ok(String::from_utf8(hv.clone())))?
+                // TODO: Remove unwrap.
+                .unwrap();
```
leighmcculloch, line 1087: This allocates a String, but only when debug-assertions are enabled. This code will not be compiled in releases used in stellar-core.

soroban-env-host/src/host.rs
```diff
@@ -1074,6 +1074,31 @@ impl VmCallerCheckedEnv for Host {
...
+            self.record_debug_event(
+                DebugEvent::new()
+                    .msg(fmt)
+                    .args(args.iter().map(|arg| arg.to_raw())),
+            )?;
+            Ok(RawVal::from_void())
+        } else {
+            Err(self.err_general(
+                "log_fmt_values unsupported outside of debug-assertion builds of the host",
+            ))
```
leighmcculloch, line 1098: We actively error if someone tries to execute a debug build on a host that does not have debug enabled. I think this is preferred over doing nothing as this ensures we will never have contracts deployed and in use that log. The downside of this is if a developer accidentally releases then finds out sometime later they left debug on and one of their functions they need fails.

I don't mind what we do here, maybe it would be saferto leave this error out and make it a no-op.

Wdyt @graydon?

soroban-env-host/src/events.rs
```diff
@@ -87,13 +87,13 @@ impl Display for DebugArg {
 /// [host::Host::debug_event](crate::host::Host::debug_event) for normal use.
 #[derive(Clone, Debug)]
 pub struct DebugEvent {
-    pub msg: Option<&'static str>,
+    pub msg: Option<Cow<'static, str>>,
```
leighmcculloch, line 90: I _think_ this is the right way to do the Cow so it can hold either the static str or the String. I'm not overly familiar with Cow though. @graydon Am I using this right?

leighmcculloch (commented):
This is ready for review.

Related PRs are here:
- https://github.com/stellar/rs-soroban-sdk/pull/654
- https://github.com/stellar/soroban-cli/pull/148

I'm not sure how to rebuild the wasm files. I had a go using the Makefile, but then nothing compiled after that.

leighmcculloch:
I updated the PR to use a feature instead of debug-assertions, because if we ever wanted to use debug-assertions to debug the host, we'd be enabling new functionality for contracts which could be dangerous.

soroban-env-host/src/host.rs
leighmcculloch: I changed it to a no-op out of concern it would create less problems for contracts that are accidentally deployed with a debug log.

leighmcculloch:
@graydon requesting a post-merge review on this one when you have a moment.

PR #466, opened 2022-09-22
Add initial "hostile contract" and unit test calling it

graydon:
This is a (preliminary, not very extensive, sketch) of a "hostile contract" that has functions that attempt to break sandbox limits (time, memory, stack depth, etc.) as well as a unit test in the host that tests that calling each of these functions causes a VM trap as expected.

It could use a lot more badness, and it's worth carefully inspecting the bytecode to check that the optimizer isn't getting rid of the badness you write. Cc @jayz22 when he's back -- this is something that needs lots more coverage.


jayz22:
@graydon Do you want to merge this one? I can pick it up and add more test cases going forward.

graydon:
@jayz22 yes please!

jayz22:
I've opened https://github.com/stellar/rs-soroban-env/pull/521 to merge this (can't commit to this branch directly).

Issue #494, opened 2022-09-29
Expose network id to contracts


leighmcculloch:
I think we should expose network id to contracts.

The network passphrase is the string:
```
Public Global Stellar Network ; September 2015
```

The network ID is the 32byte SHA256 hash of the network passphrase, in this case:
```
7ac33997544e3175d266bd022439b22cdb16508c01163f26e5cb2a3e1045a979
```

Early on we were exposing a field named network id to contracts and, if I'm remembering correctly, I asked for it to be named network passphrase. What I failed to remember in that moment that network id would actually be a hash of the passphrase, and therefore would be a fixed length value that we'd be able to use in things like as domain separators for signing in the same way that the `TransactionSignaturePayload` does.

A fixed length value can have an advantage in that it always takes up the same amount of space in a message, and therefore doesn't rely on prefixes or separators to ensure that the value disambiguates two values where one value is a prefix of another.

This doesn't impact anywhere we are using the network passphrase because the only places we use a network passphrase is inside structured values that contain length prefixes that would distinguish two values like described above. However if we expose network id we would probably use it instead of network passphrase.

In hindsight we probably could have exposed on network id and it was probably the wrong call on my part to expose network passphrase instead. There's no harm in having exposed network passphrase, however network id will be more useful.

cc @sisuresh @graydon @C0x41lch0x41

PR #496, opened 2022-09-29
Add get_ledger_network_id host fn

leighmcculloch:
### What
Add get_ledger_network_id host fn

### Why
Early on we were exposing a field named network id to contracts and, if I'm remembering correctly, I asked for it to be named network passphrase. What I failed to remember in that moment that network id would actually be a hash of the passphrase, and therefore would be a fixed length value that we'd be able to use in things like as domain separators for signing in the same way that the `TransactionSignaturePayload` does.

The network passphrase is the string:
```
Public Global Stellar Network ; September 2015
```

The network ID is the 32byte SHA256 hash of the network passphrase, in this case:
```
7ac33997544e3175d266bd022439b22cdb16508c01163f26e5cb2a3e1045a979
```

A fixed length value can have an advantage in that it always takes up the same amount of space in a message, and therefore doesn't rely on prefixes or separators to ensure that the value disambiguates two values where one value is a prefix of another.

This doesn't impact anywhere we are using the network passphrase because the only places we use a network passphrase is inside structured values that contain length prefixes that would distinguish two values like described above. However if we expose network id we would probably use it instead of network passphrase.

In hindsight we probably could have exposed on network id and it was probably the wrong call on my part to expose network passphrase instead. There's no harm in having exposed network passphrase, however network id will be more useful.

Close https://github.com/stellar/rs-soroban-env/issues/494


graydon:
You need to bump the env interface version. Also do you want to _remove_ the network passphrase? I don't really care, it and it seems like it might be a footgun?

leighmcculloch:
> Also do you want to _remove_ the network passphrase? I don't really care, it and it seems like it might be a footgun?

Maybe, unclear at this point, probably, but I'd rather defer that decision until some later point, and then just sunset the host function.

PR #497, opened 2022-09-29
Disable the create_contract_from_ed25519 host fn

leighmcculloch:
### What
Disable the create_contract_from_ed25519 host fn

### Why
We aren't quite ready to use it because we need to revise the domain separator. We want to disable it for a little while until we come back to spend some more time on it. Disabling it with a feature flag is a possibility, however feature flags come with some non-zero maintenance cost. We could comment the code but no doubt it would become stale and we may not return to it soon. It's very little code so deleting it doesn't introduce that much of a cost to recreate it.


sisuresh (approved):
LGTM. Just need to to keep in mind the stellar-core test cases will break when it updates env and this is why.

PR #500, opened 2022-09-29
Use from_xdr instead of read_xdr for slice decoding

leighmcculloch:
### What
Use from_xdr instead of read_xdr for slice decoding.

### Why
It will ensure that the bytes passed in are completely consumed by the read operation.


leighmcculloch:
For some reason GitHub auto closed this PR. Reopening in #501.

Issue #479, opened 2022-09-26
Consider if account ID needs adding to budget


leighmcculloch:
In https://github.com/stellar/rs-soroban-env/pull/473 we added the account ID type, but left it's budgeting as nothing like the contract code type. This is in the fn `charge_for_new_host_object`. It's unclear to me if we want a budget for this? I think so, but rather than add it prematurely we've left it out and created this issue.

cc @sisuresh

PR #505, opened 2022-09-29
Add metering for `AccountId` and `ContractCode`

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/479 for `AccountId`.
Plus `ContractCode`. 
Along with some minor fixes.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/host/metered_clone.rs
+++ b/soroban-env-host/src/host/metered_clone.rs
@@ -2,7 +2,7 @@ use crate::{
     budget::{Budget, CostType},
     host::Events,
     host::MeteredOrdMap,
-    xdr::{Hash, Uint256},
+    xdr::{AccountId, Hash, PublicKey, ScContractCode, Uint256},
     HostError,
 };
 use std::rc::Rc;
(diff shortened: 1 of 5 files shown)


soroban-env-host/src/host/metered_clone.rs
```diff
@@ -43,3 +43,21 @@ impl<K, V> MeteredClone for MeteredOrdMap<K, V> {
...
+            budget.charge(CostType::BytesClone, c.len() as u64)?;
+        }
+        Ok(self.clone())
+    }
+}
+
+impl MeteredClone for AccountId {
+    fn metered_clone(&self, budget: &Budget) -> Result<Self, HostError> {
+        match self.0 {
+            PublicKey::PublicKeyTypeEd25519(_) => budget.charge(CostType::BytesClone, 32 as u64)?,
```
leighmcculloch, line 59: Do we need to measure heap allocated vs not clones differently? At the moment we're bundling them into the same `BytesClone`, but it is likely to have some different cost.

The AccountId is a fixed size object that doesn't require any heap allocation, so the cost of using it is likely different to the cost of cloning the bytes of things like the wasm contract code.

cc @graydon 

(I say this with limited knowledge about how these things measure up in Rust. My experience with how stack allocated vs heap allocated data structures perform mostly comes from Go, which is likely to be very different. So I acknowledge I might be way off with this message and maybe they are the same cost.)

leighmcculloch (approved):
Looks straightforward to me. Thanks for fixing up the account id uses. When I added them (I think it was me) I didn't know if we should be measuring them or not.

I have one question, but I don't think it needs to block merging this change.

soroban-env-host/src/host/metered_clone.rs
jayz22: Hmm interesting point. I'm not sure about performance difference in heap vs non-heap clones in Rust, I'll defer that to Graydon. 
Regarding bundling them into the same type, it overestimates the cost by assuming the size is not fixed, which aligns with our goal so it should be ok.

PR #507, opened 2022-09-30
Add budget metering for the native token contract

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/476

This PR turns out to be mostly just some minor changes + a bunch of comments on why things are covered. 
- Most of the operations in the token library are just wrapping existing host functions/objects where metering is already accounted for. 
- The new types introduced in `base_types.rs` and public_types.rs` contains just an `Object` (u64), so cloning are basically free. 
- The `transfer_account_balance` and `transfer_trustline_balance` contain some arithmetics which are simple enough to skip metering for now but may be needed in the future (if they get complex)

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/host/data_helper.rs
+++ b/soroban-env-host/src/host/data_helper.rs
@@ -98,11 +98,11 @@ impl Host {
         Ok(buf)
     }
 
-    //TODO: metering!
+    // notes on metering: covers the byte cloning (write_xdr) for AlphaNum12 (worst case)
     pub fn id_preimage_from_asset(&self, asset: Asset) -> Result<Vec<u8>, HostError> {
         let pre_image = HashIdPreimage::ContractIdFromAsset(asset);
         let mut buf = Vec::new();
-        //self.charge_budget(CostType::BytesClone, 64)?; // key + salt
+        self.charge_budget(CostType::BytesClone, 44)?; // CreditAlphanum12: 12 bytes asset code + 32 bytes accountId
         pre_image
             .write_xdr(&mut buf)
             .map_err(|_| self.err_general("invalid preimage"))?;
--- a/soroban-env-host/src/host/data_helper.rs
+++ b/soroban-env-host/src/host/data_helper.rs
@@ -135,7 +135,7 @@ impl Host {
             account_id: self
                 .visit_obj(a, |id: &AccountId| Ok(id.metered_clone(&self.0.budget)?))?,
         });
-        self.visit_storage(|storage| match storage.get(&acc)?.data {
+        self.with_mut_storage(|storage| match storage.get(&acc)?.data {
             LedgerEntryData::Account(ae) => Ok(ae),
             _ => Err(self.err_general("not account")),
         })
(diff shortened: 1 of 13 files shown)


soroban-env-host/src/native_contract/base_types.rs
```diff
@@ -246,6 +248,13 @@ impl<const N: u32> BytesN<N> {
     }
 }
 
+impl<const N: u32> MeteredClone for BytesN<N> {
+    fn metered_clone(&self, budget: &Budget) -> Result<Self, HostError> {
+        budget.charge(CostType::BytesClone, N.into())?;
+        Ok(self.clone())
```
leighmcculloch, line 254: This `self.clone()` is not actually cloning anything host side, so I don't think we should be charging for it. If you see here this is the type, and it is just an `Object`, so when we clone we're just cloning a u64 value: https://github.com/stellar/rs-soroban-env/blob/0f2119839fb339e17bb592ea695317b3bf590127/soroban-env-host/src/native_contract/base_types.rs#L160-L161

leighmcculloch: I think this comment would apply to a lot of the types and clones in the native contract.

jayz22: Thanks @leighmcculloch, good catch! I updated it. 
In which case, there isn't really much changes left in this PR (which is great!) since most operations are trivial/already covered by the host.

soroban-env-host/src/host/data_helper.rs
```diff
@@ -98,11 +98,11 @@ impl Host {
         Ok(buf)
     }
 
-    //TODO: metering!
+    // notes on metering: covers the byte cloning (write_xdr) for AlphaNum12 (worst case)
     pub fn id_preimage_from_asset(&self, asset: Asset) -> Result<Vec<u8>, HostError> {
         let pre_image = HashIdPreimage::ContractIdFromAsset(asset);
         let mut buf = Vec::new();
-        //self.charge_budget(CostType::BytesClone, 64)?; // key + salt
+        self.charge_budget(CostType::BytesClone, 44)?; // CreditAlphanum12: 12 bytes asset code + 32 bytes accountId
```
leighmcculloch, line 105: Should we be using `size_of` to get the size of values? Ref: https://doc.rust-lang.org/std/mem/fn.size_of.html

soroban-env-host/src/native_contract/token/contract.rs
```diff
@@ -161,6 +164,7 @@ impl TokenTrait for Token {
             ClassicMetadata::AlphaNum12(asset) => {
                 //TODO: Better way to do this?
                 let mut code12 = [0u8; 12];
+                e.charge_budget(CostType::BytesClone, 12)?;
```
leighmcculloch, line 167: Is there a reason we can't use the len of the value we're measuring?
```suggestion
                e.charge_budget(CostType::BytesClone, code12.len())?;
```

leighmcculloch (approved):
Couple small questions, but they are more targeting the broader methods we're using to measure and not specific to this PR, so I'm approving and merging, but would be interested in your response @jayz22.

jayz22:
> Couple small questions, but they are more targeting the broader methods we're using to measure and not specific to this PR, so I'm approving and merging, but would be interested in your response @jayz22.

Thanks @leighmcculloch. I've realized some of these budget charges are repetitive (and easy to be done in a not consistent way) and addressed it in this [followup](https://github.com/stellar/rs-soroban-env/pull/510).

PR #510, opened 2022-09-30
Cleanup, consolidate metering for read/write_xdr

jayz22:
### What

Follow-up to https://github.com/stellar/rs-soroban-env/pull/507#pullrequestreview-1127202133

### Why

There are several `write_xdr` operations (and a couple `from_xdr`) which needs to be metered in the same way. It makes sense to create a shared pattern so they can be consolidated, and made sure being done consistently. 

### Known limitations

[TODO or N/A]


soroban-env-host/src/host/metered_xdr.rs
```diff
@@ -0,0 +1,28 @@
...
+        buf: &mut Vec<u8>,
+    ) -> Result<(), HostError> {
+        obj.write_xdr(buf)
+            .map_err(|_| self.err_general("failed to write xdr"))?;
+        // Notes on metering": "write first charge later" means we could potentially underestimate
+        // the cost by the largest sized host object. Since we are bounding the memory limit of a
+        // host object, it is probably fine.
+        // Ideally, `charge` should go before `write_xdr`, which would require us to either 1.
+        // make serialization an iterative / chunked operation. Or 2. have a XDR method to
+        // calculate the serialized size. Both would require non-trivial XDR changes.
```
leighmcculloch, line 20: It would nice to be able to do this, but realistically because XDR structures are nested, it would require mostly the same amount of work to calculate an accurate size as it would to just do the serialization. There a couple things we could do though, and one is very achievable today:

Put a `std::io::Write` implementation between the `buf` and the write_xdr function. The write_xdr function writes to a `std::io::Write`, which `Vec<u8>` just happens to implement. We can provide our own implementation that calls `charge_budget` as it goes along. You can do this today with just a few lines of code, and it would make the budget charging capability here much much safer.

jayz22: You mean reimplementing `std::io::write` for `Vec<u8>` to weave in budget charging? 
Or define a new type that contains `Vec<u8>` and implement `std::io::write` for it? 
The `HostError` in the return result might be an issue though.

leighmcculloch: I mean something like this:

```rust
struct MeteredWrite<'a, W>
where
    W: Write,
{
    pub host: &'a Host,
    pub w: &'a mut W,
}

impl<'a, W> Write for MeteredWrite<'a, W>
where
    W: Write,
{
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        self.host
            .charge_budget(CostType::ValSer, buf.len() as u64)?;
        self.w.write(buf)
    }

    fn flush(&mut self) -> std::io::Result<()> {
        self.w.flush()
    }
}
```

You're that passing the `HostError` from `charge_budget` back up through the reader might be a problem. What I would do here is write a conversion for `HostError: Into<std::io::Error>` that results in the `HostError` being materialized back as it was in `HostError: From<std::io::Error>`.

jayz22: Brilliant solution, thanks @leighmcculloch !

soroban-env-host/src/host/metered_xdr.rs
```diff
@@ -0,0 +1,45 @@
...
+        &self,
+        obj: &impl WriteXdr,
+        w: &mut Vec<u8>,
+    ) -> Result<(), HostError> {
+        let mut w = MeteredWrite { host: &self, w };
+        obj.write_xdr(&mut w)
+            .map_err(|_| self.err_general("failed to write xdr"))
+    }
+
+    pub(crate) fn metered_read_xdr<T: ReadXdr>(&self, bytes: &[u8]) -> Result<T, HostError> {
```
leighmcculloch, line 41: This is subtle but I'd name this `metered_from_xdr` so it matches the from_xdr fn name it is wrapping. Read_xdr and from_xdr has some different semantics in the XDR lib and it could be confusing if we mix the names here. I don't feel strongly though, it's a very minor detail.

soroban-env-host/src/host/error.rs
```diff
@@ -133,3 +133,9 @@ impl TryFrom<&HostError> for ScStatus {
         err.status.try_into()
     }
 }
+
+impl From<HostError> for std::io::Error {
+    fn from(e: HostError) -> Self {
+        std::io::Error::new(std::io::ErrorKind::Other, "host error")
```
leighmcculloch, line 139: Instead of passing a string in here you should be able to pass HostError as the second argument. Then in the reverse conversion we should make it pull the HostError out.

leighmcculloch (commented):
One thing to do with how the error is transmitted.

soroban-env-host/src/host/metered_xdr.rs
jayz22: Fixed.

soroban-env-host/src/host/error.rs
jayz22: I fixed passing in the HostError. However, I was having had a hard time pulling it out in `metered_from_xdr` (mostly Rust-related). For now I just used `err_general` and added a `TODO` comment there. I might spend a bit more time on it later. But this PR should be ok to merge without it.

jayz22: Fixed it.

soroban-env-host/src/host/metered_xdr.rs
```diff
@@ -0,0 +1,59 @@
...
+                        e3.clone()
+                    } else {
+                        self.err_general("failed to write xdr")
+                    }
+                } else {
+                    self.err_general("failed to write xdr")
+                }
+            }
+            _ => self.err_general("failed to write xdr"),
+        })
```
leighmcculloch, line 52: Oh this is nice. Fyi you can use the common `.source()` function on the Error rather than needing to match and get inner:
```suggestion
        obj.write_xdr(&mut w).map_err(|e| {
            if let Some(e2) = e.source() {
                if let Some(e3) = (*e2).downcast_ref::<HostError>() {
                    e3.clone()
                } else {
                    self.err_general("failed to write xdr")
                }
            } else {
                self.err_general("failed to write xdr")
            }
        })
```

leighmcculloch (approved):

Issue #336, opened 2022-08-16
Make invoke/call contract require explicit flag to enable reentry


leighmcculloch:
Contracts have access to calling other contracts but we haven't yet implemented things we discussed would help prevent unintentional re-entry. 

The idea is that calling a contract should require setting a flag for whether re-entry is allowed. And in this way the possibility for re-entry will be explicitly visible to the developer.

jayz22:
The host side has disabled reentry. We still need to expose a flag to set it in the host functions and plumb it on the sdk side.

jayz22:
Reopened to follow up on: https://github.com/stellar/rs-soroban-env/pull/513#issuecomment-1267220021
cc @leighmcculloch

esteblock:
Reentrancy is needed to offer liquidity for flashloans.
I like that reentrancy it's by default not allowed... and with a flag whether re-entry is wanted.

esteblock:
Hello @leighmcculloch @jayz22 @graydon 
Do you have any update on this?

It's better to leave the possibility wuth an explicit flag, than to not let do reentry at all in Soroban!

jayz22:
@esteblock We did not have chance to prioritize this. We may consider it in the next phase. This feature has just not been requested much and we felt it could create more footgun if used carelessly. 
I would like to understand first if there is any other ways to enable the flash loan feature without reentry. If you could provide more details (e.g. code snippet or WIP examples) that would be helpful!

PR #513, opened 2022-10-03
Forbid contract reentry

jayz22:
### What

Partially resolves https://github.com/stellar/rs-soroban-env/issues/336. 

~Add a `reentry: RawVal` argument to the `try_call` function. If set, it checks the host context for frames with the same contract id.~
Disallow contract reentry by checking the host context for frames with the same contract id. 
In a later follow up, we will add a `reentry: RawVal` flag to the `try_call` function and use that. 

### Why

- Why there is no checking function name? We assume if reentry is disabled, the same contract cannot be entered at all, even if the entry point is a different function than what had been called. 
- (In the future) Why only add it in `try_call`?  The `call` host function is the recommended (safe) path of invocation, where any failure will results in an host error.  The `try_call` is the more loose version that allows failure to be returned as a status. It makes sense to only enable re-entry in the `try_call` invocation path since that is a further loosening feature (@graydon has suggested it). 

### Known limitations

[TODO or N/A]


jayz22:
I haven't bumped the env version because I haven't figured out how to make the sdk changes (proc-macro) in order to regen the wasm tests.

leighmcculloch (approved):
This looks good to me, but I'm a little confused by the comment that has been added, since it is incorrect in the context of testing.

leighmcculloch:
> It makes sense to only enable re-entry in the try_call invocation path since that is a further loosening feature (@graydon has suggested it).

I don't think it makes sense to couple reentry as a capability with error handling, as there are disconnected reasons to choose one and the other.

In the SDK I hope to add support for "invocation options" on functions that call both call and try_call, and it will muddy the SDK API if these options are inconsistent between the two functions.

At the moment these options will only include the reentry flag, but I imagine them including other options in the future.

cc @graydon

graydon (approved):
LGTM! Thanks!

leighmcculloch:
@jayz22 I think we should add new call and try_call functions that take the additional option. I think we can simply call them `call_2` and `try_call_2`, and each time we add a new option we can add a new host function.

PR #515, opened 2022-10-03
Reduce the amount of roundtrips for built-in token contract

dmkozh:
### What

- `AccountId` is decoded from `RawVal` straight into XDR, as `AccountId` XDR is a host object type now
- `AccountId` is passed around in internal host fns instead of `Object`
- Removed `ClassicMetadata` as it became an internal-only structure. Instead, just pass the encoded asset to the contract asset initializer

### Why

This came up during merging the tests with the recently changed code; the recent changes made it pretty hard to understand and use the built-in token due to existence of several 'wrapper' types on the host side. This also addresses https://github.com/stellar/rs-soroban-env/issues/503 partially.

### Known limitations

N/A


sisuresh (approved):
LGTM. Thanks for cleaning this up!

soroban-env-host/src/host/conversion.rs
```diff
@@ -333,4 +335,12 @@ impl Host {
             }
         }
     }
+
+    pub(crate) fn deserialize_asset(&self, asset: Object) -> Result<Asset, HostError> {
+        self.visit_obj(asset, |hv: &Vec<u8>| {
+            self.charge_budget(CostType::ValDeser, hv.len() as u64)?;
+            Asset::from_xdr(&mut hv.as_slice())
```
jayz22, line 342: Can you use the `metered_from_xdr` function instead?
https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/metered_xdr.rs#L50

dmkozh: That's nice, I've also added an `Object` helper to avoid `visit_obj` boilerplate in callers.

jayz22 (approved):

PR #521, opened 2022-10-04
Add initial "hostile contract" and unit test calling it

jayz22:
### What

To merge https://github.com/stellar/rs-soroban-env/pull/466

### Why

### Known limitations

[TODO or N/A]


jayz22:
Can I get a merge on this?

soroban-env-common/src/meta.rs
```diff
@@ -56,5 +56,5 @@
 // implementations over a long period of time.
 
 soroban_env_macros::generate_env_meta_consts!(
-    interface_version: 23,
+    interface_version: 24,
```
graydon, line 59: I don't think this needs an interface bump. Nothing in env.rs is changing

graydon (changes requested):

soroban-env-common/src/meta.rs
jayz22: I undid the version bump.

jayz22:
@graydon I've undone the version bump. Also upgraded all the wasm dependencies of soroban from 0.0.4 to 0.0.6. Should be good to merge.

graydon (approved):

PR #529, opened 2022-10-06
Vm tuning

graydon:
Handful of changes to help with measuring and tuning the VM, including one significant perf boost which is building with LTO and codegen-units=1. Calibration is adjusted accordingly.

(I'm not merging this presently in case @leighmcculloch is mid-release, but once you're done with that I think it's worth merging)


leighmcculloch (approved):
This is probably fine to merge since the profile changes will have no impact on any importer of the crate. For these changes to materialize in stellar-core the profile config will need specifying there, not here.

Issue #442, opened 2022-09-15
Object type Debug implementations should render contents


leighmcculloch:
The Debug implementation for Vec's that gets triggered by DebugEvent formatting is a very simple `Object(Vec(3))` where the `3` is the length.

We should render something that shows the individual elements as well so that DebugEvents containing Vec's are more meaningful.

I assume the same needs doing for Maps.

---

>@graydon It is worth noting that the debug output of Vecs is pretty underwhelming in debug events ☝🏻. In the SDK Vec's have a very different output as their Debug implementation outputs every individual element. Maybe we should do something similar in the host?

_Originally posted by @leighmcculloch in https://github.com/stellar/rs-soroban-sdk/issues/611#issuecomment-1248691109_

leighmcculloch:
This issue should apply to all ScObject types.

graydon:
Actually 3 is the object ID, not the length! Maybe we should put a `#3` or something.

It's difficult to do this currently because the host is not kept alive by the `DebugEvent`s. The event buffer gets snapshotted into `HostError` which is then formatted lazily as a string when-and-if it's caught, quite late, and the host may be long gone by then. Keeping the host alive by holding a reference to it in the `DebugEvent` or `HostError` is possible, but even if we did that it's not necessarily the case that the host objects referenced are still alive at the time of a catch either (when we roll back a host, for example, objects get dropped).

We could format eagerly (when the event is saved to the log) but that would make the log operation much more expensive in cases when nobody looks at the results, and also require keeping strings in the log which .. I've been trying to avoid doing.

Another dimension of this is that a general value might be quite _big_ and we don't necessarily want to pay the storage costs nor spam the debug log with gigantic stringified values even in debug cases. There ought to be some kind of size limit / data elision on printing, even if we go a little ways into printing.

I'm not going to close this WONTFIX but .. I feel a little stuck with the current design pressures. I think I'd like to at least put this on hold until we decide on RawVal size and repertoire, whether we're going to support slices/user strings, and maybe whether/how we are going to define a "small object" type. Maybe we can find something good enough for users with some combination of those features, like "capture small objects by value, by snapshotting their XDR type and storing it to the side of the log buffer". Or maybe we need to never drop objects on rollback or something. Unclear.

One question for @leighmcculloch : would you be ok with this behaving differently in production vs. local testing? It's easier to tolerate expensive diagnostic stuff in the local testing case as it's not being metered / not a DoS vector.

[...]

PR #532, opened 2022-10-06
Improved built-in token error reporting.

dmkozh:
### What

Instead of returning a generic error, the contract returns more specific status code and creates a debug event.

This is the first shot at the error names; I'm open to suggestions on names and scope of the error types.

### Why

Improving the developer experience (https://github.com/stellar/rs-soroban-env/issues/526)

### Known limitations

Not updating the tests now, given that https://github.com/stellar/rs-soroban-env/pull/525 that adds a lot of error scenarios is still not merged.


soroban-env-host/src/native_contract/token/error.rs
```diff
@@ -1,35 +1,23 @@
...
 
-impl From<Error> for HostError {
-    fn from(e: Error) -> Self {
-        match e {
-            Error::HostError(he) => he,
-            Error::ContractError => crate::xdr::ScUnknownErrorCode::General.into(),
-        }
-    }
+pub fn contract_err(host: &Host, err: ContractError, msg: &'static str) -> HostError {
+    host.err(DebugError::new(Status::from_contract_error(err as u32)).msg(msg))
```
jayz22, line 22: I don't think you need this function. This is basically exactly what the `err_status_msg` (https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/err_helper.rs#L39) is for.

jayz22 (commented):
Thanks for making this change. One suggestion regarding reusing existing helper functions in `err_helper.rs`. 
A general comment, it seems you are using the more verbose form of error reporting everywhere (via `host.err`), with both the status and msg). However, most of the msgs are just repeats of the status itself without extra information. You may consider using the `err_status` method instead, this will shrink the code a bit. Or, if you do care about extra information, you can construct the `DebugError` and pass in extra `RawVal`s to be part of the debug event.

soroban-env-host/src/native_contract/token/error.rs
dmkozh: Thanks, I've only somehow noticed message-only and status-only fns. Changed code to use `err_status_msg`.

dmkozh:
> Thanks for making this change. One suggestion regarding reusing existing helper functions in `err_helper.rs`. A general comment, it seems you are using the more verbose form of error reporting everywhere (via `host.err`), with both the status and msg). However, most of the msgs are just repeats of the status itself without extra information. You may consider using the `err_status` method instead, this will shrink the code a bit.

I think whether or not we want to have messages depends on the ease of accessing the events from the user side. If the error text events are easy enough to access, I would prefer to have them, even though they could be redundant in some cases. That's because: 
- text is still more readable
- IIUC status reporting relies on sdk matching this exact enum
- in some cases text describes more specific issue and I'd rather be consistent everywhere
- we may want a bit less precise set of errors; I'm still not sure about this one

If you think that errors aren't going to be easy to access, then I guess we could remove the messages altogether, but then I'm not sure what would be the value/purpose of events.

dmkozh:
I figured that moving the balance transfer fns is very easy, so I decided to add that change to this PR in order to have a better understanding of the spectrum of the contract errors we have.

jayz22:
> > Thanks for making this change. One suggestion regarding reusing existing helper functions in `err_helper.rs`. A general comment, it seems you are using the more verbose form of error reporting everywhere (via `host.err`), with both the status and msg). However, most of the msgs are just repeats of the status itself without extra information. You may consider using the `err_status` method instead, this will shrink the code a bit.
> 
> I think whether or not we want to have messages depends on the ease of accessing the events from the user side. If the error text events are easy enough to access, I would prefer to have them, even though they could be redundant in some cases. That's because:
> 
> * text is still more readable
> * IIUC status reporting relies on sdk matching this exact enum
> * in some cases text describes more specific issue and I'd rather be consistent everywhere
> * we may want a bit less precise set of errors; I'm still not sure about this one
> 
> If you think that errors aren't going to be easy to access, then I guess we could remove the messages altogether, but then I'm not sure what would be the value/purpose of events.

I'm not opposed to the msg, it doesn't hurt. 
But I'm not sure what you mean by the ease of accessing events/error. From Paul's original request the events are dumped on error so they are clearly accessible (in the cli workflow at least). The whole purpose of splitting the error status from the debug event is because the error status should contain the minimal critical information that is sufficient in identifying the error type without needing a message to explain it. The `DebugEvent` provides capability to provide detailed information, beyond a string, by passing in RawVals. For example, in one of your messages:

> Err(e.err_status_msg(
                ContractError::NegativeAmountError,
                "negative amount is not allowed",
            ));

You could provide a RawVal representing the violating amount, or another RawVal representing the violating the identity to make the debugging info much richer.

dmkozh:
> But I'm not sure what you mean by the ease of accessing events/error.

I wasn't sure what are the scenarios where debug events and also how granular the errors should be. There is an ongoing discussion concerning these topics. I think for now it should be ok to have richer debug events; we can always remove them if there is any problem.

> You could provide a RawVal representing the violating amount, or another RawVal representing the violating the identity to make the debugging info much richer.

There wasn't a helper to do that; I've added a macro for now and updated the errors where it made sense to me. PTAL

jayz22 (approved):

leighmcculloch:
> > You could provide a RawVal representing the violating amount, or another RawVal representing the violating the identity to make the debugging info much richer.
> 
> There wasn't a helper to do that; I've added a macro for now and updated the errors where it made sense to me. PTAL

It's worth noting that any Object (in a RawVal) that is inserted into a DebugEvent will only be renderered as their object handle, not the value within until https://github.com/stellar/rs-soroban-env/issues/442 is addressed. 

This means if the "violating amount" is a BigInt you won't actually see the amount in the output.

dmkozh:
> It's worth noting that any Object (in a RawVal) that is inserted into a DebugEvent will only be renderered as their object handle, not the value within until https://github.com/stellar/rs-soroban-env/issues/442 is addressed.

Oh, that's unfortunate we don't have this yet. I would hope this to make it into the next release, so that we could merge errors with args now and fix the display later (from the host standpoint the event is a string and a bunch of raw vals anyway). WDYT?

dmkozh:
It would be nice to get this merged. Is there anything else blocking this PR?

leighmcculloch:
I don't think so, I assume it is just waiting on @sisuresh to review and merge it.

sisuresh (approved):
LGTM. Thanks!

PR #542, opened 2022-10-18
Correct / tidy up / comment calibration code.

graydon:
This does a bit of refactoring, commenting and correcting calibration routines. It also adds the ability to run a single calibration benchmark by name by passing it on the command line.

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -634,9 +634,10 @@ impl Host {
 
     pub(crate) fn charge_for_new_host_object(
         &self,
+        prev_len: usize,
         ho: HostObject,
     ) -> Result<HostObject, HostError> {
-        self.charge_budget(CostType::HostObjAllocSlot, 1)?;
+        self.charge_budget(CostType::HostObjAllocSlot, prev_len as u64)?;
         match &ho {
             HostObject::Vec(v) => {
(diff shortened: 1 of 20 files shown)


soroban-env-host/src/host.rs
```diff
@@ -634,9 +634,10 @@ impl Host {
 
     pub(crate) fn charge_for_new_host_object(
         &self,
+        handle: usize,
         ho: HostObject,
     ) -> Result<HostObject, HostError> {
-        self.charge_budget(CostType::HostObjAllocSlot, 1)?;
+        self.charge_budget(CostType::HostObjAllocSlot, handle as u64)?;
```
leighmcculloch, line 640: I think this is the first place we are passing a handle around outside of a `RawVal` or an `Object`. Given that `Object` is almost free, it might be worth passing the handle around as an `Object` and extracting its handle at this line.

leighmcculloch (commented):
I didn't review the benchmark logic in detail, but have one concern with the change in the host. But I don't feel strongly and doesn't need to block if you disagree.

soroban-env-host/src/host.rs
graydon: It's a synonym for "size of the object array" here -- the handle allocated is always "the previous size of the array". I can rename it for clarity.

PR #544, opened 2022-10-20
Pseudo random

graydon:
This adds a new subsystem to the host that stores a set of ChaCha20 CSPRNGs associated with each calling context (either a contract ID or the empty context, used for raw host functions). The PRNGs are lazily initialized and seeded separately with the SHA256 combination of a common per-host seed (default zero but can be set by embedder) and the context ID. The common per-host seed will be set, in the stellar-core embedding, to a hashed combination of the txset hash and the current tx position within the txset.

The idea here is to give users a fighting chance when it comes to PRNGs. The interface is simple and likely something we can support unchanged in the future even if we improve the seeding (eg. strengthen it to multi-party seeding from all voting validators) and/or stronger separation mechanisms in the host and embedder. For now, users will at least get a PRNG that has state that's as hard to guess as our existing front-running protection in txset ordering, and as hard to leak the state of as anything else held in the host, separated by contract contexts and only revealed through output-observations.

Hopefully it's enough to prevent users from trying to grow their own versions seeded with the ledger sequence number alone, or whatever. Comments welcome!


soroban-env-guest/src/guest.rs
```diff
@@ -161,6 +165,11 @@ impl EnvBase for Guest {
         // Intentionally a no-op in this cfg. See above.
         Ok(())
     }
+
+    fn prng_fill_slice(&self, mem: &mut [u8], finalize: bool) -> Result<(), Status> {
+        // Intentionally a no-op in this cfg. See above.
+        Ok(())
```
leighmcculloch, line 171: ❓ Why is this a no-op in the guest? We'd want this to call through to the `prng_fill_linear_memory` function wouldn't we, like how the bytes functions do above.

soroban-env-common/src/env.rs
```diff
@@ -96,6 +96,12 @@ pub trait EnvBase: Sized + Clone {
         vals: &[RawVal],
         strs: &[&'static str],
     ) -> Result<(), Status>;
+
+    /// Fill a byte slice from the host PRNG.
+    fn prng_fill_slice(&self, _mem: &mut [u8], _finalize: bool) -> Result<(), Status> {
+        // TODO: remove default impl when SDK env implements prng_fill_slice.
+        unimplemented!()
```
leighmcculloch, line 103: 💡 I think it's okay if you want to drop the default impl here, and we'll just impl this straight away in the SDK. I don't mind though. If you leave this out it will make sure we don't miss impl this on the SDK 😄

soroban-env-common/src/env.rs
```diff
@@ -461,6 +467,23 @@ macro_rules! call_macro_with_all_host_functions {
...
+                /// Return a u63 from the persistent, stateful PRNG associated
+                /// with the calling context, finalizing the PRNG if `finalize`
+                /// is set to `true`. Traps if PRNG is already finalized.
+                {"0", fn prng_next_u63(finalize:RawVal) -> RawVal}
+                /// Fill the region of linear memory between `lm_pos` and
+                /// `lm_pos` + `len` with bytes drawn from from the persistent,
+                /// stateful PRNG associated with the calling context,
+                /// finalizing the PRNG if `finalize` is set to `true`. Traps if
+                /// PRNG is already finalized. Returns void.
+                {"1", fn prng_fill_linear_memory(lm_pos:RawVal, len:RawVal, finalize:RawVal) -> RawVal}
```
leighmcculloch, line 484: 💭 In the SDK I can see us implementing an interface similar to that of the [`rand::Rng`](https://docs.rs/rand/latest/rand/trait.Rng.html) type, although much simpler, something in the form:

```
pub trait ... {
    fn gen<T>(&mut self) -> T;
    fn fill<T>(&mut self, dest: &mut T);
}
```

Where `gen` would probably support primitive types that we support across boundaries, like `u32`, `u64`, `i32`, `i64`, and maybe smaller types too I guess like 8-bit, 16-bit, and the size integer types.

Where `fill` would probably support byte slices, byte arrays, and `Bytes` and `BytesN` types.

For the `gen` case:
- ❓ I'd like `gen` to only return positive values for the integer types. That's what the Go stdlib's math/rand and crypto/rand functions do, and in my experience that is 99% what users want. However, it seems like the behavior of `rand::Rng` is to return values in the full range of the type. It might be better to do the full range way for consistency with other Rust programs. Wdyt @graydon?
- ✅  `prng_next_u32` looks good and we'd use it for the 8-bit, 16-bit, and 32-bit cases, truncating it in the formers.
- ❓ `prng_next_u63` looks good if we only support positive values on types like i64 if we model the Go API, but could be difficult to use if we model the Rust rand::Rng API, since in the SDK we'll surface it as a `i64` but in practice it will only ever generate half the range. We could add a special function like `gen_pos_i64` specifically, that would call through to this, that'd work.
- ❓ If we want full range integer types, what would be the recommendation for getting a u64, or full range i64? We probably need a `prng_next_u64` that we can use for those two cases.

For the `fill` case:
- ✅ The API looks good.

leighmcculloch (commented):
Oh this nice. I have some question inline about the interface.

soroban-env-host/src/host.rs
```diff
@@ -2760,4 +2806,53 @@ impl VmCallerCheckedEnv for Host {
...
+
+    fn prng_next_u63(
+        &self,
+        vmcaller: &mut VmCaller<Self::VmUserState>,
+        finalize: RawVal,
+    ) -> Result<RawVal, Self::Error> {
+        self.charge_budget(CostType::PrngGenerateBytes, 8)?;
+        self.with_prng_context_and_finalize_flag(finalize, |prngs, context, finalize| {
+            let u = prngs.next_u64(context, finalize)?;
+            Ok(unsafe { RawVal::unchecked_from_u63(u as i64) })
```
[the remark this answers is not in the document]
graydon: Yes, but `unchecked_from_u63` is unchecked: it just casts to u64, left-shifts the input and sets the low bit, throwing out the high; there are literally only 63 bits of storage available to it. It takes an i64 as input just by convention / a way to signal to the caller (somewhat) that this won't be all the bits of a u64 they're expecting.

(I don't love the u63 interface in general, but I think this is not misusing it as it stands; happy to redesign it!)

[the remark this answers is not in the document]
leighmcculloch: Ah, right, got it. 👍🏻 Nah this is fine.

soroban-env-guest/src/guest.rs
graydon: Oops, correct. Will fix.

soroban-env-common/src/env.rs
graydon: I can't recompile the examples without it (they use the SDK).

soroban-env-common/src/env.rs
graydon: Re: positive values, and the API shape in general: I think if we say we're generating a type, we should return the full range of the type.

But thinking about it a bit more, I also think we might want to steer users towards slightly higher-level APIs, which are what they usually want. Random number generation doesn't happen in a vacuum, it's usually task-driven. We could do some of these in the host also, they're standard-enough (and use host objects enough) to do there:

- Shuffle a host vec (do in host)
- Select a random element of a host vec or a map (do in host)
- Generate a value in a given `[lo,hi]` value-range (maybe do in guest)
- Generate a value with a given symbolic distribution (maybe do in guest)

I could definitely modify it to return i64 or u64 full values, if thoroughness-wrt-number-types is of interest, I just figured I'd do a non-object-allocating interface. Not sure which scenarios users are going to really want/need (or what our plans are for reforming the number types in general).

leighmcculloch: All of those things sound good to me.

The last two dot points I think the host should provide too. We may as well?

For i64 / u64 would it be okay for me to just make 2x next_u32 calls to construct a hilo? If that's the case, that's probably more efficient than doing one host fn call that constructs a host object which then has to be unpeeled via an extra host function call anyway.

For the select random element, that's essentially the same thing generating a value in a range, if we add the ability to access a map with an index.

MonsieurNicolas: I would stay away for now (as host functions) from providing those extra things as it's unlikely to be perf critical: the more useful thing to do is to be sure that we provide the right low level building blocks that will be implemented correctly.

At a minimum:
* fill an `[u8]` with random data (you already have that)
* construct "types of interest" from `[u8]`
   * I see that we already support this for `BigNum`. For now this is probably enough for "all integers" (including i64) as `BigNum` has `to_i64`
   * Should it be possible to generate random symbols, strings too?

soroban-env-common/src/env.rs
```diff
@@ -461,6 +467,23 @@ macro_rules! call_macro_with_all_host_functions {
                 {"0", fn verify_sig_ed25519(x:Object, k:Object, s:Object) -> RawVal}
             }
 
+            mod prng "p" {
+                /// Return a u32 from the persistent, stateful PRNG associated
+                /// with the calling context, finalizing the PRNG if `finalize`
```
MonsieurNicolas, line 472: why do we need the `finalize` functionality? I thought we'd expose something similar to `/dev/random` ?

graydon: See comment in prng.rs -- the point of finalizing is to help contracts defend against callers that keep poking them to try to leak bits of state from their PRNG.

soroban-env-common/src/env.rs
leighmcculloch: It's common in the Rust stdlib + community `rand` crate to provide a `Vec::choose` method, which can be given an RNG to choose a random value from the Vec.

I think what @graydon is suggesting here by offering host functions to select a random element from a Soroban Vec is approximately equivalent.

We can definitely do that on the SDK side instead, but it also seems like something the host could do entirely.

leighmcculloch: I'm less concerned with the performance benefit of doing it in the host at this point, and more interested in the fact that if we do that, we keep as much rand logic as possible in the host, so the SDK can be rather unintelligent about its provision of random capabilities.

MonsieurNicolas: I don't think we know what people will actually want to do with those things, and the more we add (in the SDK or in the host) the more we have to specify and test. It seems easy to add in the host or in SDKs later as part of the normal process.

soroban-env-common/src/env.rs
MonsieurNicolas: Oh I see. We can discuss this more on the mailing list when we get to that CAP too.

My intuition is that if a contract expects to make some secure decision based on PRNG we should never allow the non finalize behavior as it may be under attack.

Maybe a better version of this for v1 is to *not* expose this granular finalize and instead disallow PRNGs to be generated across contexts:
* `I` calls a method on `J`
    * `J`'s PRNG gets initialized the first time a random number is requested.
    * `J` calls `next_u32` as many times as it wants
* as soon as `J` returns, its PRNG gets finalized (note that it may be more complicated than this if we allow re-entrancy)
* if `I` tries to call `J` and `J` tries to generate a random number -> error

This still has the same vulnerability "ie `I` can poke `J` but requires an attacker to submit many transactions, which is not hard to do (ie `call lottery.pickLotteryWinner(); if !checkIWon() trap()` ) but a lot more expensive. We could make it that this "finalize" behavior is sticky for an entire ledger too but then maybe this is a way to DoS those contracts (``call lottery.pickLotteryWinner(); trap("LOL")` ).

There is actually a third option that might be better than all of this in v1: don't allow PRNGs to be used from a subcall (this may be also useful in general?). This would mitigate all problems above as attackers can't wrap calls with custom logic/conditions.
edit: oh actually there is one way. people could attack the contract using gas in some situations. If PRNG yields 2 outcomes `A` or `B`, and the attacker wants the cheaper outcome `A` it could provide enough gas to execute `A` but not enough to execute `B`. To mitigate this, we'd need a way to charge "constant gas" in some situations (which should not be that hard to do - but work).

graydon: I put `finalize` in the API exposed from host to guest so that the SDK could experiment with this question without the host having to change its behaviour.

soroban-env-common/src/env.rs
leighmcculloch: > We can punt on container-select / leave it to the guest if we're willing to expose a map-get-by-numeric-index method. We don't have one currently but it's plausible and more general, I don't mind adding that.

I know I suggested the indexing idea for maps a few messages up, but I regret is and want to walk it back. Reason being is that reflecting on it, this would probably be the only place where we'd use the index. And an index into a map would be in many ways a footgun to expose in the SDK so we'd probably just use it internally. So we'd be adding this index to access a random element, but not using the index elsewhere, so we may as well just have a host function that returns a random key from the map. Which I think is what you originally suggested.

soroban-env-common/src/env.rs
MonsieurNicolas: I think only option 1 can be implemented in the SDK?

leighmcculloch: I doubt the SDK can predict when finalize needs to be called. The SDK won't know if the user is going to call another function that will need random data. What am I missing?

MonsieurNicolas: I think the way you could do a simplified version of 1 is by forcing a pattern with RAAI so that if a block of code uses a PRNG you can force call finalize, so that any other block would fail to generate a PRNG?

soroban-env-common/src/env.rs
MonsieurNicolas: I think adding this without having a way to transform vectors is a missed opportunity to make it better for other real use cases. Spinning up a thread on Discord...

soroban-env-common/src/env.rs
MonsieurNicolas: I thought about this more, and while option 3 for example makes it better by orders of magnitude
* none of those options can resist biasing from an attacker (so an attacker can get 100x or 1000x higher chance of winning than good actors)
* those options don't really get better over time as we improve the PRNG seed.

I think the one pattern that would satisfy both of these properties is a select & reveal scheme:
* contract picks random values in one transaction and record the tx hash at the same time (so we need to have a way to get the hash of the current tx) - the contract should not leak any information about its choice, and because only contracts have access to their own state the "choice" cannot be attacked
* in a different transaction, the contract can act in a deterministic way based on choices made earlier.
     * For example, by seeding its local (deterministic) random generator with the random seed from step 1.
     * We'd need good pseudo-random library, but that should be easy to come by

soroban-env-common/src/env.rs
graydon: `swap` might be nice to add, but isn't adequate -- it still produces a new vector on each call. immutable objects! the point is to enable creating a shuffled vector efficiently, without O(n) intermediates.

I agree a number-in-a-range API is probably one of the more useful things to provide, to avoid footguns.

soroban-env-common/src/env.rs
graydon: I don't think contracts can ever have state that's secret-for-long. As soon as a txset is confirmed its order is known and the consensus input to any PRNGs accessed in that block is known. The whole approach I'm aiming for here is that txs inside that txset might use the PRNG state decided _for_ the txset _in_ the txset so that it's too late to abuse that knowledge in the next txset (and too soon for adjacent txs inside the txset to know the inputs).

The commit/reveal schemes I've seen for building PRNGs on chain have to do with secret state held off-chain by 3rd party oracles. I don't think it can really work on-chain.

graydon: Update: as predicted, people are "approximating" PRNG service by using various much-easier-to-guess constants such as pointer addresses. We should land something here ASAP.

graydon:
Update: as predicted, people are "approximating" PRNG service by using various much-easier-to-guess constants such as pointer addresses. We should land something here ASAP.

We've talked to a 3rd party that does "true" VRF/RNG oracle service and they made a few assertions:
  - People use PRNGs in contracts fairly often
  - Serious users with significant money they're worried about will either adopt a user-driven commit/reveal scheme or an external VRF-based oracle; people are sufficiently aware and concerned about MEV attacks by validators that they'll want to close that gap
  - Lower-stakes, early-stage or extremely cost-conscious users will roll their own even-more-attackable "PRNG" seeded with something terrible until they care, and this is the niche where it makes sense to have a standard function so long as the caveats (i.e. document how a validator may perform an MEV attack) are clearly marked on it.
  - Given the trust structure embedded in SCP (the fact that validators have business reputations to protect) the odds of an MEV attack are possibly lower anyways.

I still wonder if we could tweak SCP a bit to include some extra validator signatures (or say validator executions of VRFs with inputs declaratively stapled to the outside of transactions) to make stronger versions work, but I think we should do _something_ here soon, and the approach I've prototyped (modulo API questions above) is probably still worth doing for the lower-stakes / early-stage users. We can protocol-version it and upgrade it to something stronger in the future.

nano-o:
Ideally, we would want randomness that is unpredictable by anyone under the  failure assumptions of SCP and revealed only after transactions have been ordered. This is likely to be a hard problem.

One idea is to use a random beacon whose task is to provide a random seed after a block has been proposed but before execution begins. But I don't know of any algorithm that would work under the same failure assumptions as SCP.  Schemes based on threshold cryptography require an expensive setup which is a problem in a dynamic environment, and VDF-based solutions are probably too slow anyway (there would be a window of 2 or 3 seconds to provide the random seed before execution).

A good compromise may be to require block builders to provide a verifiable random seed for their blocks using a VRF evaluation tied to the block hash. VRFs give better guarantees that just hashing (unpredictable by others, uniform). In this scheme, SCP guarantees are still the same. BEV attacks are possible, but the block builder has to grind, and block builders that are caught in BEV attacks may lose the trust of other nodes, be removed from their slices, and thus won't get their blocks included in the chain anymore.

graydon:
We're running out of time on this bug and I'd like to ship something at least _plausibly_ useful in the first iteration (i.e. by the end of may). I'm willing to cut this PR down to its most minimal form to ship, can I get a _focused_ list of required changes from @MonsieurNicolas, @leighmcculloch and/or @nano-o ?

leighmcculloch:
> want randomness that is unpredictable by anyone under the failure assumptions of SCP and revealed only after transactions have been ordered

@nano-o My understanding is that by using the Tx Set Hash and the index of a tx in the Tx Set, the resulting value is not predictable until after the ledger is closed. Is that correct? Is that sufficient? It seems to me like using a hash of those inputs would meet the requirement of being revealed only after transactions have been ordered.

nano-o:
If we use the TxSet hash then things are predictable at least to the validator that builds the TxSet, who can therefore order and include/exclude transactions to its advantage. If we assume that nomination leaders (e.g. all of Tier 1) are honest then hashing the TxSet seems fine.

graydon:
> If we assume that nomination leaders (i.e. all of Tier 1 at least) are honest

I know this somewhat flies in the face of the MEV experience on other chains, but .. it is currently the case that whoever nominates a block can delay or drop transactions, or inject their own (with enough time to make multiple attempts at finding an order that would let them front-run). IOW users already trust them not to be doing so quite a bit.

MonsieurNicolas:
I think we can probably land the minimal version of this in the host so that people have a strong starting point for seeds:
* see https://github.com/stellar/rs-soroban-env/pull/544/files#r1000993383 (maybe just writing into linear memory is the mvp for this)
* do *not* include the `finalize` functionality that is insecure (this has to be solved in client code)
* when generating numbers, seed in a similar way that we do [elsewhere](https://github.com/stellar/stellar-core/blob/77c9b6914ae4dfb30a793dcd167c8ef4feaa5c54/src/transactions/CreateClaimableBalanceOpFrame.cpp#L301), mixing in the `txSet` hash for now (we can change this to something better later)

graydon:
@MonsieurNicolas when you say "don't include the `finalize` functionality", do you mean don't include the call as something explicit in the API, or don't allow finalization at all (for example we could still do what you suggested and auto-finalize any PRNG in use by a contract as soon as that contract returns and/or makes a call to another contract).

In terms of functions to reveal, I'm leaning towards this set:

  - `prng_get_u64_in_range(lo:u64,hi:u64) -> u64`: easy to get wrong, plain modulus will bias the result.
  - `prng_get_bytes(size:u32) -> BytesObject`: inefficient to do in guest if you want a lot of bytes.
  - `prng_shuffle_vec(v:VecObject) -> VecObject`: inefficient to do in guest and easy to get wrong.

These 3 primitives probably cover 90% of use cases they're all very easy, and each is either hard to do efficiently in the guest or easy to screw up (but have very standard implementations that all stdlibs do basically the same way).

graydon:
continued in https://github.com/stellar/rs-soroban-env/pull/802

PR #546, opened 2022-10-26
Synth wasm

graydon:
This adds a new non-published utility crate for synthesizing wasm blobs that interact with the soroban host through its interface directly, not using the SDK. It's primarily meant for testcase generation and exercising the host through the VM programmatically, though it may find other future uses.

The main things it adds (aside from a few setup/finish conveniences) are helpers that generate import directives and emit calls to host-functions; these are generated from the same macro that generates the host interface itself. There's an example showing this use in the test.rs file.

It uses `wasm-encoder` rather than `parity-wasm` for writing wasm; the latter is unmaintained at this point and the former is the official crate for such things, from the bytecode alliance.


soroban-env-common/src/raw_val.rs
```diff
@@ -59,6 +59,18 @@ pub enum Tag {
...
+    pub const fn rawval_const(&self) -> i64 {
+        ((*self as i64) << 1) | 1
+    }
+    pub const fn rawval_mask() -> i64 {
+        ((TAG_MASK as i64) << 1) | 1
+    }
+}
+
+sa::const_assert!(Tag::I32.rawval_const() == 3);
+sa::const_assert!(Tag::rawval_mask() == 0xf);
```
leighmcculloch, line 72: What are these for?

soroban-synth-wasm/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "soroban-synth-wasm"
```
leighmcculloch, line 2: Love the name. 👏🏻

leighmcculloch (approved):
This is very cool. 👏🏻

soroban-env-common/src/raw_val.rs
graydon: They're used in `assert_val_tag` just to check that a given RawVal has a given tag (simultaneously checking that it has the 1-bit set). We could use the existing constants but they don't include the low 1-bit, they're for checking the tag separately from the 1-bit, as done in the RawVal code.

PR #548, opened 2022-10-27
Catch panics from native contracts in try_call, fix #430

### LINKED ISSUE #430: Env::try_call will panic if a natively registered contract panics
Env::try_call will panic if a natively registered contract is invoked and that contract panics, but the function when deployed will never see a panic. This means during integration testing with Rust code instead of a WASM the behavior is different, which is not ideal, and may cause tests to break or misrepresent the actual behavior.

Discussion: https://discord.com/channels/897514728459468821/1019330208651821147

cc @graydon @jonjove

graydon:
This addresses #430 in a way that I _think_ is fairly close to what @leighmcculloch was asking for -- we discussed in some detail today and I'd appreciate confirmation that this is the track you'd like to take.

The goal here is to centralize the ability to "trap panics in test-only native contracts" in the host, to pave the way for removing the logic in the SDK that does the same. If an SDK user wants to trap a panic in a native contract they can/should use `env.try_call` which, as of this change, will do so in as close as possible emulation to the way a VM contract works: both errors and VM-traps (panics) get routed into status returns from `try_call`.

The main reason for doing this in the host rather than sharing responsibility with the SDK is to allow the host to manage _some_ degree of transactionality inside itself on this code path, specifically by trapping _inside_ the frame push/pop code, such that we manage to trap the error and run `rollback` before propagating it.

To get this working right and reporting decent-looking diagnostics required _not_ using `panic_any` (there's a comment about it in the code) but rather stashing status codes in a refcell in the `TestContractFrame`, which is a bit gross but it works.

With all of this wired up, we get a fairly nice diagnostic result. For example this is the output from a `call` (not `try_call`) to a native contract that, itself, calls `panic_error!` in the SDK (and thus `Host::fail_with_status`), which generates an error, which is escalated to a panic by the `impl Env for CheckedEnv::fail_with_status` wrapper, which unwinds and is caught by the invoking `Host::call_n` and converted to an error, which is then returned from `call` and escalated to a panic _again_ by the `impl Env for CheckedEnv::call`, and is ultimately caught and reported by the unit test runner:

```
thread 'test::hello_error_panic' panicked at 'HostError
Value: Status(ContractError(1))

Debug events (newest first):
   0: "caught panic with 'Status(ContractError(1))' in contract function 'Symbol(hello)'"
   1: "escalating error 'Status(ContractError(1))' to panic"
   2: "failing with contract error status code 'Status(ContractError(1))'"

Backtrace (newest first):
   0: soroban_env_host::host::err_helper::<impl soroban_env_host::host::Host>::err
             at /home/graydon/src/rs-soroban-env/soroban-env-host/src/host/err_helper.rs:18:37
   1: soroban_env_host::host::Host::call_n::{{closure}}
             at /home/graydon/src/rs-soroban-env/soroban-env-host/src/host.rs:871:33
   2: soroban_env_host::host::Host::with_frame
             at /home/graydon/src/rs-soroban-env/soroban-env-host/src/host.rs:403:19
   3: soroban_env_host::host::Host::call_n
             at /home/graydon/src/rs-soroban-env/soroban-env-host/src/host.rs:823:24
   4: <soroban_env_host::host::Host as soroban_env_common::vmcaller_checked_env::VmCallerCheckedEnv>::call
             at /home/graydon/src/rs-soroban-env/soroban-env-host/src/host.rs:2041:19
   5: soroban_env_common::vmcaller_checked_env::<impl soroban_env_common::checked_env::CheckedEnv for T>::call
             at /home/graydon/src/rs-soroban-env/soroban-env-common/src/vmcaller_checked_env.rs:168:43
   6: soroban_env_common::checked_env::<impl soroban_env_common::env::Env for T>::call
             at /home/graydon/src/rs-soroban-env/soroban-env-common/src/checked_env.rs:124:41
   7: <soroban_sdk::env::Env as soroban_env_common::env::Env>::call
             at /home/graydon/src/rs-soroban-sdk/soroban-sdk/src/env.rs:652:13
   8: soroban_sdk::env::Env::invoke_contract
             at /home/graydon/src/rs-soroban-sdk/soroban-sdk/src/env.rs:237:18
   9: test_result::ContractClient::hello::{{closure}}
             at src/lib.rs:12:1
  10: test_result::ContractClient::with_env
             at src/lib.rs:12:1
  11: test_result::ContractClient::hello
             at src/lib.rs:12:1
  12: test_result::test::hello_error_panic
             at src/lib.rs:89:21
  13: test_result::test::hello_error_panic::{{closure}}
             at src/lib.rs:81:5
  14: core::ops::function::FnOnce::call_once
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/ops/function.rs:248:5

', /home/graydon/src/rs-soroban-env/soroban-env-host/src/host.rs:1315:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test test::hello_error_panic - should panic ... ok

```


soroban-env-host/src/host.rs
```diff
@@ -1204,6 +1273,49 @@ impl VmCallerCheckedEnv for Host {
...
+    //
+    // It might seem like we ought to `std::panic::panic_any(e)` here, making
+    // the panic carry a `HostError` or `Status` and catching it by dynamic type
+    // inspection in the `call_n` catch logic. The reason we don't do so is that
+    // `panic_any` will not provide a nice printable value to the `PanicInfo`,
+    // it constructs, so when/if the panic makes it to a top-level printout it
+    // will display a relatively ugly message like "thread panicked at Box<dyn
+    // Any>" to stderr, when it is much more useful to the user if we have it
+    // print the result of HostError::Debug, with its glorious status code,
+    // site-of-origin backtrace and debug log.
```
leighmcculloch, line 1302: 💯 Agreed, this seems like the right trade-off for us to be making to not use that mechanism.

soroban-env-host/src/host.rs
```diff
@@ -799,9 +818,59 @@ impl Host {
             // maintains a borrow of self.0.contracts, which can cause borrow errors.
             let cfs_option = self.0.contracts.borrow().get(&id).cloned();
             if let Some(cfs) = cfs_option {
-                return self.with_frame(Frame::TestContract(id.clone(), func.clone()), || {
-                    cfs.call(&func, self, args)
-                        .ok_or_else(|| self.err_general("function not found"))
+                let frame = TestContractFrame::new(id.clone(), func.clone());
```
leighmcculloch, line 821: ❔ Does this change the behavior at all of a `Env::call` in a panic scenario? I don't think so, but not sure as I see changes in this `call_n` function.

leighmcculloch (approved):
The high level gist of this change makes sense to me. One question inline, but this looks good to me.

soroban-env-host/src/host.rs
```diff
@@ -799,9 +818,59 @@ impl Host {
...
+                                .arg::<RawVal>(func.into()),
+                        )),
+                        Err(_) => {
+                            // Return an error indicating the contract function
+                            // panicked. If if was a panic generated by a
+                            // CheckedEnv-upgraded HostError, it had its status
+                            // captured by VmCallerCheckedEnv::escalate_error_to_panic:
+                            // fish the Status stored in the frame back out and
+                            // propagate it.
+                            let msg = "caught panic with '{}' in contract function '{}'";
```
leighmcculloch, line 865: ❔ It appears this will make the debug events significantly different when running in WASM VM vs native. I'm not sure that's a good thing. Do we need to create the debug event for the escalation and panic caught? I'm not attached to them not being present, I'm just wondering if it will cause confusion.

graydon: I can make them _more similar_ if you like (eg. remove the escalation event, rephrase the "caught" message to be somewhat closer to the "VM trapped with host error" error we make if there is a VM trap with host error) but .. I'm a little hesitant here. I would also say:

1. Debug events are for _debugging_, not something users should rely on the content of. We want to retain freedom to change debug events fairly freely over time, and we're probably going to add quite a lot. It may well be that we will be emitting debug events on VM setup/teardown or host function calls -- this will necessarily make these two code paths emit different debug events _intentionally_, because different things are happening. 
2. VM-trap vs. panic-and-catch are fundamentally slightly-different mechanisms, and they're never going to be identical. For example an out-of-bounds memory read or divide-by-zero will deliver a signal to the process rather than panic, whereas a VM will trap them. And even more concretely: the budgeting of native contracts is fundamentally different, and panics arising from out-of-budget will always be different.

soroban-env-host/src/host.rs
graydon: I'm not 100% sure what you're asking. The .. whole point is to change behaviour! I'll try to answer by detailing the before and after behaviour and you can tell me if it does or doesn't meet expectations.

First: this only affects native contracts and calls between them. No change to VM-based calls.

Before this change:
  - If contract A `call`s contract B and B panics, the panic will unwind through the host (actually somewhat corrupting the host by failing to `rollback` the `TestContractFrame`) and all the way back to A. If A happens to have `catch_panic` or something they might catch it, otherwise it bubbles up to the outermost process catch handler which will probably print the panic.
  - If contract A `try_call`s contract B and B panics, the panic also bubbles up, `try_call` doesn't contain it any better.
  - If the panic originated from `impl Env for CheckedEnv` unwrapping a `HostError` from a failed host function called by B, the panic will contain a stringification of that `HostError` which captures where the error originates, as well as (I think?) a duplicate stack trace of where the `unwrap` failed (moments after the `HostError` was captured). 

After this change
  - If contract A `call`s contract B and B panics, the panic will unwind to `Host::call_n` and be caught and converted to a `HostError`. The host will `rollback` the `TestContractFrame` and then return `Err(HostError)` to `call`, which will return that `Err(HostError)` to `impl Env for CheckedEnv`, which will call back to `Host::escalate_error_to_panic`, which will panic and unwind into A, where its fate will be the same as the panic in the before-case.
  - If contract A `try_call`s contract B and B panics, the same initial sequence of events happens but when `Err(HostError)` is returned from `call_n` to `try_call`, it will be converted to `Ok(status)` which will pass through `impl Env for CheckedEnv` and return to B as a `RawVal`-that-happens-to-be-a-`Status`. No further unwinding happens, `try_call` contained the panic.
  - If the panic originated from `impl Env from CheckedEnv` escalating a `HostError` from a failed host function called by B, the panic's originating context (at least its stack trace and captured error buffer) will _mostly_ be eaten by `call_n` catching it. Its status will be propagated to the `HostError` returned from `call_n`

soroban-env-host/src/host.rs
leighmcculloch: I think removing the additional escalation event might be a good idea, but keeping the text for the other one would be fine. I agree calling it a VM trap would be misleading and we shouldn't say anything untrue. I agree people shouldn't depend on debug events and they should change, but if the debug events make the code and behavior appear significantly different that may be confusing.

soroban-env-host/src/host.rs
leighmcculloch: Got it. My question was directed at wanting to make sure the Env::call behavior wasn't changing significantly, or if it was, understanding the change. I think this sounds right as you describe it.

In some ways this erode the Rust testing experience because someone chucking a panic may expect their debugger to take them to exactly where the original panic occurred, but they won't have that experience anymore. I think we have to balance creating that ideal experience vs the ideal same experience as production, and I think this PR makes the right tradeoff.

leighmcculloch (approved):
🎉

soroban-env-host/src/host.rs
graydon: I checked this! As it happens, the vscode debugger (by default) sets a breakpoint on the panic-initiating site, so at the moment of the first panic you break into a debugger.

leighmcculloch: Oh nice. Yay!

leighmcculloch (approved):

PR #551, opened 2022-10-28
Use stellar/binaries for binary installs

leighmcculloch:
### What
Use stellar/binaries for binary installs.

### Why
To avoid rebuilding tools.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,7 +60,10 @@ jobs:
     - uses: stellar/actions/rust-cache@main
     - run: rustup update
     - run: rustup target add ${{ matrix.sys.target }}
-    - run: cargo install --target-dir ~/.cargo/target --locked --version 0.5.16 cargo-hack
+    - uses: stellar/binaries@v12
+      with:
+        name: cargo-hack
+        version: 0.5.16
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -60,7 +60,10 @@ jobs:
     - uses: stellar/actions/rust-cache@main
     - run: rustup update
     - run: rustup target add ${{ matrix.sys.target }}
-    - run: cargo install --target-dir ~/.cargo/target --locked --version 0.5.16 cargo-hack
+    - uses: stellar/binaries@v12
```
jayz22, line 63: Why do we want to use pre-built binary (that doesn't have the code change) for testing?

leighmcculloch: This installs a binary for cargo-hack. It installs the same version that was being installed via cargo before. The only significant difference is that we aren't compiling it here.

leighmcculloch: Why: rust compiles are slow. This specific tool isn't so bad once cached, but cargo-workspaces takes over 5 minutes which we use in some workflows. Since we have the infra for binaries for these tools now may as well use them for this one too.

PR #552, opened 2022-10-29
Capture user panic-strings in native builds, avoid spurious NoContractRunning error

graydon:
This improves slightly on #548 in two ways:

  - Removes a spurious `NoContractRunning` error that gets logged, harmlessly but confusingly, when escalating an error to a panic _outside_ a `try_call`, like if the user does `call` and then there's a panic
  - Catches and logs user strings, such that if a user happens to call `panic!("some string")` in their code (a fairly normal thing to do in Rust) then "some string" will wind up written in the debug log and so a unit test that does `#[should_panic(expected="some string")]` will work right.

cc @leighmcculloch


leighmcculloch:
🎉

PR #555, opened 2022-11-01
Bug 554 - fix stack overflow

### LINKED ISSUE #554: Stack overflow adding element to metered map past a certain size
Someone on Discord reported an issue where adding more than 2966 contract data entries resulted in an error:
```
error: transaction submission failed
```

I tried to make a small contract that reproduces the issue in https://github.com/stellar/soroban-examples/pull/165 and when I tried to add a 2208th item, I see a `fatal runtime error: stack overflow`. I've seen the following signal messages:
- `signal: 11, SIGSEGV: invalid memory reference`
- `signal: 6, SIGABRT: process abort signal`

Full output being:
```
thread 'test::test_stack_overflow' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass '-p soroban-increment-contract --lib'

Caused by:
  process didn't exit successfully: `/Users/leighmcculloch/Code/soroban-examples/target/debug/deps/soroban_increment_contract-58c64acdb16b6c5c 'test::test_stack_overflow' --exact --nocapture` (signal: 6, SIGABRT: process abort signal)
```

Discussion https://discord.com/channels/897514728459468821/1036512231556390932

cc @graydon @jayz22

graydon:
First things first: this _only_ applies to debug builds. In opt builds you can build gigantic maps with no stack overflows (run the included-but-marked-as-`ignore` tests with `cargo test --release unittests -- --include-ignored`).

However, users get a _debug_ build of the host when they run contract unit tests in native local testing mode, and they can (apparently have in the wild) done things that produce large maps in those tests, and that can produce stack overflows and process aborts. Bad experience.

What goes wrong?

The insert path of `im` allocates a handful of interior B-tree nodes "by value" / as local variables held on the stack while it's rebalancing its B-tree. Interior B-tree nodes are fixed-size arrays of 64 copies of a (key,val) entry, and for maps storing `LedgerKey` and `LedgerVal` -- which the `storage` subsystem of the host does -- the entry size adds up to over 300 bytes, such that each interior B-tree node is over 20kb. Empirically, if users add a few thousand nodes, we get a few levels of B-tree splitting deep, and _debug builds_ wind up using several hundred KB with those 20KB+ stack slots _in each frame_, so the deepest frame hits the conventional 2MB thread stack limit on linux.

A more-complete solution would probably be to patch `im` to allocate fewer interior nodes, and/or to make the rust compiler and/or LLVM do more aggressive reuse of stack slots in debug builds, or .. whatever. But the _easy_ fix is just to make the key and value types in the map smaller by indirection behind `Box<...>`, which is what I'm doing here.

I checked (and left it in another of the ignored tests) that this lets debug builds get up past 16 million entries without problem. Probably much further, but it takes many minutes to run that test in a debug build, so I stopped there. I left a 64k-entry test (that finishes in a reasonable time even in debug builds).

Fixes #554

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -63,7 +63,7 @@ use self::metered_vector::MeteredVector;
 // Notes on metering: `RollbackPoint` are metered under Frame operations
 #[derive(Clone)]
 pub(crate) struct RollbackPoint {
-    storage: MeteredOrdMap<LedgerKey, Option<LedgerEntry>>,
+    storage: MeteredOrdMap<Box<LedgerKey>, Option<Box<LedgerEntry>>>,
     objects: usize,
 }
 
--- a/soroban-env-host/src/host/metered_map.rs
+++ b/soroban-env-host/src/host/metered_map.rs
@@ -43,12 +43,21 @@ impl<K, V> Default for MeteredOrdMap<K, V> {
     }
 }
 
+fn check_size_is_small<T>(name: &str) {
+    let sz = core::mem::size_of::<T>();
+    if sz > 64 {
+        panic!("type '{}' is too big: {}", name, sz);
+    }
+}
+
 impl<K, V> MeteredOrdMap<K, V>
 where
     K: Ord + Clone,
     V: Clone,
 {
     pub fn new(budget: Budget) -> Result<Self, HostError> {
+        check_size_is_small::<K>("key");
+        check_size_is_small::<V>("val");
         budget.charge(CostType::ImMapNew, 1)?;
         Ok(MeteredOrdMap {
             budget,
@@ -57,6 +66,8 @@ where
     }
 
     pub fn from_map(budget: Budget, map: OrdMap<K, V>) -> Result<Self, HostError> {
+        check_size_is_small::<K>("key");
+        check_size_is_small::<V>("val");
         budget.charge(CostType::ImMapNew, 1)?;
         Ok(MeteredOrdMap { budget, map })
     }
--- a/soroban-env-host/src/storage.rs
+++ b/soroban-env-host/src/storage.rs
@@ -42,7 +42,7 @@ pub trait SnapshotSource {
 /// against a suitably fresh [SnapshotSource].
 // Notes on metering: covered by the underneath `MeteredOrdMap`.
 #[derive(Clone, Default)]
-pub struct Footprint(pub MeteredOrdMap<LedgerKey, AccessType>);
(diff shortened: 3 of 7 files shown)


leighmcculloch (approved):
🎉

jayz22:
🎉🎉 This is great! 
I have a couple of questions:
1. Is there/will there be a noticeable compute cost for having every entry on the heap?
2. Since the im library haven't changed, is it still making internal copies of the entries? The entries now are a few bytes (two 64-bit addresses) vs 300 bytes previously so why does the capacity improve by over 3 orders of magnitude (previously it failed to store 3000 entires?)

sisuresh (approved):

graydon:
> 1. Is there/will there be a noticeable compute cost for having every entry on the heap?

Noticeable in the sense that the number will be a little different: yes. Noticeable in the sense of being a different complexity class or even constant order of magnitude: almost certainly no. A heap allocation should only cost a small constant. I expect several operations may actually be cheaper since it makes B-tree nodes smaller (it can be hard to figure out cache effects of such a change -- the interior nodes are smaller so traversing them is faster, but comparing any two leaves or fetching a result requires dereferencing a pointer to a totally different cache line, so.. it might go either way).

> 2. Since the im library haven't changed, is it still making internal copies of the entries? The entries now are a few bytes (two 64-bit addresses) vs 300 bytes previously so why does the capacity improve by over 3 orders of magnitude (previously it failed to store 3000 entires?)

You have to look at this in terms of log-space.

The stack was overflowing because the `im` library was producing absolutely _gigantic_ stack frames (like 800kb in a single frame) during recursive B-tree split / rebalance operations inside the insert path. The number of such recursive B-tree split/rebalance frames is/was actually quite small, because it corresponds to _levels_ in the B-tree: the example I saw failing was only _3 frames_ deep into B-tree code, when storing 3000 entries, which is what we'd expect in a B-tree with 64-way fanout: it was just starting to recursively push entries out of the 2nd level of the B-tree (of which there are 64*64=4096 entries when perfectly balanced) and into the 3rd level (of which there are 64*64*64=262144 entries when perfectly balanced).

My change made the size of those gigantic stack frames much smaller. I did some casual measurements that suggested something like 10x-20x smaller (I stopped at a random deep insert and it was using 120kb of stack out of the available 2MB) which is about what you'd expect going from 300-ish bytes to 32-ish bytes. So there's now "room" for the B-tree code to make 10x as many _frames_ in its recursive rebalancing before running out of stack. But that's _30 frames_ which would only happen if the B-tree had populated _30 levels_. A 30-level B-tree with 64-way fanout has 64^30 elements -- far more than can exist in the universe.

PR #559, opened 2022-11-03
Disable default-features on stellar-xdr workspace dependency

leighmcculloch:
### What
Disable default-features on stellar-xdr workspace dependency..

### Why
When I recently moved the dependencies to the workspace, I didn't realize that the workspace dependencies were additive in regards to features. This meant that default features, which includes the `std` feature in the stellar-xdr, are enabled unless the workspace dependency explicitly turns it off. The default features are enabled even for crates that explicitly turn them off, because features are always additive in a workspace.

I've introduced a test a crate that imports the common crate, and adds a panic handler, which is a hack to detect if std was imported or not, as a way to prevent this from accidentally happening again.


soroban-env-common/tests/no_std/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "test_no_std"
```
leighmcculloch, line 2: I'm going to merge this to keep the release 🚂 going, but @graydon if you have any alternative ideas for how to test this here in this repo or general feedback with what I'm doing here, I'm all ears.

Note that I already tried making a integration test module that has the harness turned off that is nostd. That didn't work because tests have to be binaries and this needs to be a lib.

leighmcculloch:
Reported this issue to the cargo repo at https://github.com/rust-lang/cargo/issues/11329.

PR #561, opened 2022-11-03
Restructure benchmark framework, add calibration code for all `CostType`s.

jayz22:
### What

**Restructure the bench framework.** 
This PR splits the previous `HostCostMeasurement` into two places:
1. `soroban-env-host/src/cost_runner`: contains the runner of each cost type, as well as the structural definitions of the sample data. 
2. `soroban-env-host/benches`: contains the data generator for each cost type. 
Information dependency goes from 2-->1 but not the other way around. 

**Implement calibration for basically all cost types.** Except a few, see limitations. 

**Cleaned up, consolidated, and reindexed `CostType`.** 
See "Why" section for which types were removed and the reasoning. 

### Why
The main reason for the restructuring to run non-public functions in the host for calibration.
The current calibration/benchmark framework consist of roughly three parts:
1. Generation of data, whose structure requirement is defined by a particular cost type. 
2. The bench harness, dispatchers for kicking off a measurement for a (bench type, cost type, iteration) combination. 
3. Running a cost entity in the host. 
 
Previously all these sit with the "benches", which is outside of the host crate.  However, the host runner should belong to the host, since it needs access to the (non-public) host components. Same for the data structural definitions.

Removed `CostType`s:
- `*AllocCell`: such as `HostVecAllocCell`. The cost of adding a host object has already been covered by `HostObjAllocSlot` , and the cost of producing a particular host object has been covered elsewhere. 
- `BytesSlice`: this is essentially cloning, covered by `BytesClone`
- `BytesConcat`: covered by `BytesAppend`
- `PushFrame` and `PopFrame`: too granular. The cost of frame pushing and popping has been covered by `GuardFrame`, and push/pop are always paird. 
- `CallArgsUnpack`: this is too granular. The work of converting a guest function call argument from `ScVal` to a `RawVal` has been covered elsewhere. This operation just collects `RawVal`s into a `Vec`. 
- `CloneVm`: this is just a `Rc` clone, should be free. 

### Known limitations
Several types were not calibrated because we need to know the size limits of its data, in order to approximate the cost with a constant (upper bound) cost term:
- `CloneEvents`
- `ScVecFromHostVec`
- `ScMapFromHostMap`
- Storage related: need to know the size limit of keys which are LedgerKeys


soroban-env-host/src/cost_runner/runner.rs
```diff
@@ -0,0 +1,37 @@
...
+    /// machinery, so anything that happens during it will be considered part of
+    /// the cost for running the HostMeasurement at the returned input. Will call
+    /// `run_iter` with iter set to each number in 0..RUN_ITERATIONS.
+    /// Returns the total input consumed by this run.
+    fn run(host: &Host, sample: &mut Self::SampleType) -> u64 {
+        let mut iter_input = None;
+        for iter in 0..Self::RUN_ITERATIONS {
+            iter_input = Self::run_iter(host, iter, sample);
+        }
+        match iter_input {
```
graydon, line 32: The reason the previous `get_input` function was a separate call rather than an option-return was so that we did as little extra overhead-work besides the `run_iter` function under supervision of CPU-and-memory tracking. I'd prefer to keep it separate rather than mixing it into the measured work like this.

graydon (changes requested):
Looks mostly good, except I'd like to revert the change that moves `get_input` to a `Option<u64>` return.

soroban-env-host/src/cost_runner/runner.rs
jayz22: Makes sense. I've fixed it.

jayz22:
@graydon I pushed another change to construct and pass in `Vec<SampleType>` instead of `&mut SampleType` to the runner. Mainly because we want to remove internal cloning from the measurements. Now it constructs a vector by cloning sample data from the outside and then passes it into the runner to run the host, without accounting for sample cloning. 

During this change I discovered an issue. Previously the the mem_tracker start happened right before calling the `HCM::run`, and it was underflowing the `AtomicU64` counter after I made the above switch. I believe it's because the memory deallocation of the sample data container was being counted while the allocation (which happens before the tracker started) was not. I moved that call before the data setup, and it seem to give the correct results. I've left a comment in the code. Please let me know if my assessment is correct and if there is a better pattern to solve such issue (rearranging the code this way feels too subtle and error-prone).

soroban-env-host/benches/common/measure.rs
```diff
@@ -292,34 +289,44 @@ where
...
+        <HCM::Runner as CostRunner>::COST_TYPE
+    );
     loop {
+        // prepare the measurement
         let host = Host::default();
         host.with_budget(|budget| budget.reset_unlimited());
-        let mut m = match next_hcm(&host) {
+        // memory tracking needs to start here (instead of right before `run`) because the sample
+        // data container allocation needs to be counted. Otherwise, the allocation/deallocation
+        // count will be asymmetric, which will lead to mem tracker count underflow.
```
jayz22, line 302: @graydon Please see this comment, and my comment with this PR.

jayz22: @graydon and I discussed and decided we only count allocations not deallocations since we care about the worst-case estimate, and this mimics the arena model which is the worst case. I updated the `deallocated` and added a comment. These two lines are moved back to where it was before (right before running the host).

jayz22:
> @graydon I pushed another change to construct and pass in `Vec<SampleType>` instead of `&mut SampleType` to the runner. Mainly because we want to remove internal cloning from the measurements. Now it constructs a vector by cloning sample data from the outside and then passes it into the runner to run the host, without accounting for sample cloning.
> 
> During this change I discovered an issue. Previously the the mem_tracker start happened right before calling the `HCM::run`, and it was underflowing the `AtomicU64` counter after I made the above switch. I believe it's because the memory deallocation of the sample data container was being counted while the allocation (which happens before the tracker started) was not. I moved that call before the data setup, and it seem to give the correct results. I've left a comment in the code. Please let me know if my assessment is correct and if there is a better pattern to solve such issue (rearranging the code this way feels too subtle and error-prone).

Addressed in https://github.com/stellar/rs-soroban-env/pull/561#discussion_r1019815009

graydon (approved):

PR #571, opened 2022-11-11
Use the current SDK version for test WASMs.

dmkozh:
### What

Use the current SDK version for test WASMs.

This also makes all the examples use the single workspace version.

### Why

The examples used outdated SDK versions. This block the breaking changes to common env.

### Known limitations

N/A


leighmcculloch (approved):

leighmcculloch (approved):
This is so much simpler now that we don't have to have so many patch entries 👏🏻. It will probably be easier to keep updated I guess now.

leighmcculloch (approved):

jayz22:
cc @graydon who originally set up the repo structure this way in https://github.com/stellar/rs-soroban-env/pull/354. 
Making sure we are not over-simplifying or missing any use cases.

jayz22:
Thanks @leighmcculloch and @dmkozh, this is a great improvement!
I also wasn't aware of the workspace inheritance feature was added in Cargo 1.64 (2022-09-22) which was before the original PR. Glad to see things become more simple.

PR #576, opened 2022-11-18
Update stellar-xdr

leighmcculloch:
### What
Update stellar-xdr.

### Why
Get the latest stellar-xdr that has XDR version embedded.


leighmcculloch:
Uh, I can't upgrade the stellar-xdr package, because it looks like it has a significant amount of changes in the XDR files themselves.

Issue #579, opened 2022-11-19
panics in the host via mistaken comparison and conversion dispatch


graydon:
There's a fairly serious (and old!) bug in the way `EnvVal<WeakHost,RawVal>::cmp` calls `WeakHost::obj_cmp` which calls through to `impl Env for CheckedEnv` and treats it as infallibe. This is causing panics because it's not actually infallibe (specifically: comparing values can exceed budget and thereby return a `HostError`), and the latter `impl` escalates `HostError` to panic (which is appropriate for local-testing mode but inapprorpiate for production hosts).

This is happening because we (really just I) overlooked the fact that all the very-early plumbing (in April) that put in place originally to make `im` work (`EnvVal` and `WeakHost` and such) got made-useless by the slightly-later transition (in May) to infallible host functions and `CheckedEnv`. It all compiles and runs, but it's essentially incoherent: any error that happens inside an `im` call to `Ord::cmp` winds up escalating to a panic.

This is really quite bad. My best guess at a solution here is .. quite invasive and horrible:

- [x] DONE (https://github.com/stellar/rs-soroban-env/pull/608) Get rid of `EnvVal` and `WeakHost` entirely (eliminating the `Ord` impl along the way)
- [x] DONE (https://github.com/stellar/rs-soroban-env/pull/585) Get rid of `im`, and actually we can't really use anyone's containers because they all have this characteristic -- we have to write our own maps and vecs that have fallible lookup routines
- ~Carefully audit and/or possibly feature-gate code paths in the production host that might accidentally call through to the `impl Env for CheckedEnv` -- it should _really_ only be seen/called by contracts running tests. Possibly move it to the SDK to avoid the possibility of calling it in the host's own codebase.~
- [x] DONE (https://github.com/stellar/rs-soroban-env/pull/628) Refactor all the `TryFromVal` paths to enable systematic reform of errors
- [x] PENDING (https://github.com/stellar/rs-soroban-env/pull/633) Remove `impl Env for CheckedEnv` entirely and make all `Env` methods fallible with a configurable `Error` provided by the `Env` type.

graydon:
[...]

PR #585, opened 2022-11-27
Delete im

graydon:
This still has some TODOs in the form of error codes to consolidate/reassign, but the code is otherwise I think as it should be.

This PR deletes the `im` library previously backing the `Host` collection objects (maps and vecs), and also reworks the comparison system to avoid https://github.com/stellar/rs-soroban-env/issues/579. It replaces both vecs and maps with simple eagerly-copied `std::Vec`s (maps are sorted `std::Vec<(K,V)>`s). It also eliminates all the `WeakHost` and `EnvVal` usage in the `Host`; with a little more work plus catching up the SDK it should be possible to remove `EnvVal` entirely. In addition to removing `WeakHost` it also removes all the auxiliary pointers to `Budget` stored in "metered" host objects; a single `Budget` and/or `Host` is passed in to each operation that needs it.

The only odd part of this is that maps and vecs no longer "share substructure" via the BTree/RRB code in `im` -- every edit makes a full copy -- but some consideration suggests this probably isn't a _huge_ issue:
  - The host vecs and maps still "share substructure" of any host objects they refer to; they're just slabs of `RawVal`s, so are still "shallow" in that sense (and the storage map entries are changed from `Box` to `Rc`, so cloning them is "shallow" now too).
  - For small maps (less than 64 entries) `im` would be holding everything in a single BTree or RRB node anyways, which means this is no less efficient. The sharing only kicked in on larger structures, and ... we're likely not going to _have_ big maps or vecs available to users.
  - Since we're removing `EnvVal` (which used to comprise the map and vec entries) the data consumption actually _shrinks_ with this change (again so long as you're dealing with small maps or vecs).
  - I've been fairly careful to make all the operations that build new maps or vecs do so with a single exact-size allocation, again making these smaller for small maps and also relatively efficient. We can make them more efficient still by going to a per-host arena allocator in the future, if we want.

Comments welcome!


leighmcculloch:
Is it happening?

graydon:
@leighmcculloch I believe so, if you would be inclined to review! I have pushed a rebase and it should be ready now.

(It'd be good to do soonish since I will be out after thursday, and this will bitrot fairly quickly)

soroban-env-host/src/host/comparison.rs
```diff
@@ -0,0 +1,274 @@
...
+impl_compare_fixed_size_ord_type!(LedgerKeyClaimableBalance);
+impl_compare_fixed_size_ord_type!(LedgerKeyLiquidityPool);
+impl_compare_fixed_size_ord_type!(LedgerKeyContractCode);
+impl_compare_fixed_size_ord_type!(LedgerKeyConfigSetting);
+
+impl_compare_fixed_size_ord_type!(LedgerEntryExt);
+
+impl_compare_fixed_size_ord_type!(AccountEntry);
+impl_compare_fixed_size_ord_type!(TrustLineEntry);
+impl_compare_fixed_size_ord_type!(OfferEntry);
```
sisuresh, line 129: Do we need all of these? For example, when will we compare (or even use) offer entries in Soroban?

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
+        Self::from_vec(Vec::new())
+    }
+
+    pub fn from_array<const N: usize>(buf: [A; N], budget: &Budget) -> Result<Self, HostError> {
+        budget.charge(CostType::VecNew, N as u64)?;
+        Self::from_vec(buf.into())
+    }
+
+    pub fn from_vec(vec: Vec<A>) -> Result<Self, HostError> {
+        // No charge here: vector already allocated, charge happened in claler.
```
jayz22, line 56: nit: typo "caller"

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
+            budget.charge(CostType::VecNew, sz as u64)?;
+            // It's possible we temporarily go over-budget here before charging, but
+            // only by the cost of temporarily allocating twice the size of our largest
+            // possible object. In exchange we get to batch all charges associated with
+            // the clone into one (when A::IS_SHALLOW==true).
+            let vec: Vec<A> = iter.collect();
+            A::charge_for_clones(vec.as_slice(), budget)?;
+            Ok(Self { vec })
+        } else {
+            // TODO use a better error code for "unbounded input itertors"
```
jayz22, line 78: nit: typo "iterators"

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
+    // This doesn't take ExactSizeIterator since that is not implemented for Chain
+    // (see https://github.com/rust-lang/rust/issues/34433) but it only works
+    // with iterators that report an exact size_hint, and it constructs a new
+    // Vec from that iterator with a single allocation-and-copy.
+    pub fn from_exact_iter<I: Iterator<Item = A>>(
+        iter: I,
+        budget: &Budget,
+    ) -> Result<Self, HostError> {
+        if let (_, Some(sz)) = iter.size_hint() {
+            budget.charge(CostType::VecNew, sz as u64)?;
```
jayz22, line 69: nit: use `self.charge_new?

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
-    pub fn from_array<const N: usize>(budget: Budget, buf: [A; N]) -> Result<Self, HostError> {
-        let mut vec = Self::new(budget)?;
-        for v in buf {
-            vec.push_back(v)?;
+impl<A> MeteredVector<A>
+where
+    A: MeteredClone,
+{
+    pub fn new(budget: &Budget) -> Result<Self, HostError> {
+        budget.charge(CostType::VecNew, 0)?;
```
jayz22, line 46: nit: use `self.charge_new?

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
+        Self::from_vec(Vec::new())
+    }
+
+    pub fn from_array<const N: usize>(buf: [A; N], budget: &Budget) -> Result<Self, HostError> {
+        budget.charge(CostType::VecNew, N as u64)?;
```
jayz22, line 51: nit: use `self.charge_new?

jayz22 (commented):
Half way through my review, will continue tomorrow.

leighmcculloch (approved):
👍🏻. My knowledge of some of the intricacies is limited now so I think this should probably get an ✅ from another person, but it looks good to me overall.

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
-        self.budget.charge(CostType::ImVecImmutEntry, x)
+    fn charge_scan(&self, budget: &Budget) -> Result<(), HostError> {
+        budget.charge(CostType::VecEntry, self.len() as u64)
     }
 
-    fn charge_cmp(&self, x: u64) -> Result<(), HostError> {
-        self.budget.charge(CostType::ImVecCmp, x)
+    fn charge_binsearch(&self, budget: &Budget) -> Result<(), HostError> {
+        let mag = 64 - (self.len() as u64).leading_zeros();
+        budget.charge(CostType::VecEntry, mag as u64)
```
jayz22, line 37: Nit: I think if we want to ensure over-estimate, we should charge `mag+1`.

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
+    A: MeteredClone,
+{
+    fn charge_new(&self, size: usize, budget: &Budget) -> Result<(), HostError> {
+        budget.charge(CostType::VecNew, size as u64)
     }
 
-    fn charge_mut_access(&self, x: u64) -> Result<(), HostError> {
-        self.budget.charge(CostType::ImVecMutEntry, x)
+    fn charge_access(&self, count: usize, budget: &Budget) -> Result<(), HostError> {
+        budget.charge(CostType::VecEntry, count as u64)
```
jayz22, line 28: Are the `CostType::VecEntry` and `VecNew` charges are dependent on the element type `A`?
- For `VecNew` I think so since it needs to pre-allocate memory depending on `size_of<A>`.
- For `VecEntry` I think not because it's just moving a pointer? 

Am I correct?

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
 }
 
-impl<A: Clone + Ord> MeteredCmp for MeteredVector<A> {
-    fn metered_cmp(&self, other: &Self, budget: &Budget) -> Result<Ordering, HostError> {
-        assert!(Rc::ptr_eq(&self.budget.0, &other.budget.0));
-        self.charge_cmp(min(self.len(), other.len()) as u64)?;
-        Ok(self.vec.cmp(&other.vec))
+impl<Elt: MeteredClone> Compare<MeteredVector<Elt>> for Host
+where
+    Host: Compare<Elt, Error = HostError>,
```
jayz22, line 285: I think these two impls can be replaced by a generic:

`impl<B: AsBudget, Elt: MeteredClone> Compare<MeteredVector<Elt>> for B
where
    B: Compare<Elt, Error = HostError>,`

soroban-env-common/src/raw_val.rs
```diff
@@ -470,6 +470,11 @@ impl RawVal {
         Self(x)
     }
 
+    #[inline(always)]
+    pub const fn shallow_eq(&self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
jayz22: @leighmcculloch and @graydon I remember you guys had quite some back and forth regarding whether allowing shallow comparison for RawVal is the right thing to do. Are all of the concerns for not doing it back then gone/addressed?

soroban-env-common/src/compare.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn compare(&self, a: &Rc<T>, b: &Rc<T>) -> Result<Ordering, Self::Error> {
+        <Self as Compare<T>>::compare(self, &*a, &*b)
+    }
+}
+
+// Apparently we can't do a blanket T:Ord impl because there are Ord derivations
+// that also go through &T and Option<T> that conflict with our impls above
+// (patches welcome from someone who understands trait-system workarounds
+// better). But we can list out any concrete Ord instances we want to support
+// here.
```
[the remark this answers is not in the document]
graydon: Oh, sorry, it actually should be transplanted over to the the block of fixed-size comparisons in host::comparison! Fixing.

soroban-env-host/src/host/comparison.rs
```diff
@@ -0,0 +1,274 @@
...
+
+struct FixedSizeOrdType<'a, T: Ord>(&'a T);
+impl<T: Ord> Compare<FixedSizeOrdType<'_, T>> for Budget {
+    type Error = HostError;
+    fn compare(
+        &self,
+        a: &FixedSizeOrdType<'_, T>,
+        b: &FixedSizeOrdType<'_, T>,
+    ) -> Result<Ordering, Self::Error> {
+        self.charge(CostType::BytesCmp, core::mem::size_of::<T>() as u64)?;
```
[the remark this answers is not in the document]
graydon: Oh, good point! Yeah, it's not 100% stable. I'll switch this to just charging a constant 1. It's not a precise bound but it means "it's not free".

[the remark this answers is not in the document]
graydon: (And yes I'll add a note about the requirements, that's it exactly: nothing variable sized. Containing pointers is not really an issue if I stop using size_of and just charge a fixed constant, so long as the pointers are not to unbounded-size substructures.)

soroban-env-host/src/host/comparison.rs
graydon: Well, they're substructures of LedgerKey / LedgerEntry which are stored in the storage maps. I guess since we're just charging at the base level of them and not descending into their substructure I can eliminate the substructure cases.

graydon: (Sorry, wrote a wrong comment here): they're substructures of LedgerKey / LedgerEntry and we explicitly switch on the outer structure to special-case the charge for the one variable-sized case: `ContractData`. So we need to also handle the fixed-size cases, even if they never get stored in practice.)

soroban-env-host/src/host/metered_vector.rs
graydon: I think so, but as you mentioned above `size_of` is not stable / we can't rely on it. So we should just capture the linear input size and treat all such costs as the same.

soroban-env-host/src/host/metered_vector.rs
graydon: Oh, sure. Fixed.

soroban-env-host/src/host/metered_vector.rs
graydon: fixed.

soroban-env-host/src/host/metered_vector.rs
graydon: fixed.

soroban-env-host/src/host/metered_vector.rs
graydon: fixed.

soroban-env-host/src/host/metered_vector.rs
graydon: fixed.

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -1,250 +1,298 @@
...
+        F: FnMut(&A) -> Result<Ordering, HostError>,
     {
-        self.charge_immut_access(self.len() as u64)?;
-        Ok(self.vec.binary_search(value))
+        self.charge_binsearch(budget)?;
+        let mut err: Option<HostError> = None;
+        let res = self.vec.binary_search_by(|probe| {
+            // We've already hit an error, return Ordering::Equal
+            // to terminate search asap.
+            if err.is_some() {
```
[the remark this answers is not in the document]
graydon: I'm not 100% certain. In theory if `cmp` returns an error we store it in `err` and then return `Ordering::Equal` which _should_ terminate the binary search at that moment; but it's also possible that binary search does another comparison for some reason after the first `Ordering::Equal` so I put this branch in to ensure we "stick with our answer" and coax it into terminating ASAP.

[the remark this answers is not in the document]
graydon: (Fwiw the source to `binary_search_by` is here: https://doc.rust-lang.org/src/core/slice/mod.rs.html#2448 and I think it terminates immediately anyways, so it shouldn't ever happen.)

soroban-env-host/src/host/metered_vector.rs
graydon: I tried this but I get one of the perplexing trait errors: 

```
error[E0210]: type parameter `B` must be covered by another type when it appears before the first local type (`MeteredVector<Elt>`)
   --> soroban-env-host/src/host/metered_vector.rs:266:6
    |
266 | impl<B: AsBudget, Elt: MeteredClone> Compare<MeteredVector<Elt>> for B
    |      ^ type parameter `B` must be covered by another type when it appears before the first local type (`MeteredVector<Elt>`)
    |
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local, and no uncovered type parameters appear before that first local type
    = note: in this case, 'before' refers to the following order: `impl<..> ForeignTrait<T1, ..., Tn> for T0`, where `T0` is the first and `Tn` is the last
```

I've found basically that half or 2/3 of the times when I try to do something "more generic" the trait system prevents me from doing it, either due to one of these sorts of prohibitions or else by triggering infinite recursion or some other solver-conflict. @leighmcculloch has had a similar experience, and we've been leaning towards "more impls that are repetitive but at least work" in a lot of recent code.

soroban-env-host/src/host/comparison.rs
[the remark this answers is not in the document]
jayz22: > Yeah, it's not 100% stable.

Ah I see. That's disappointing, I was hoping the exact size of the type could be used for charging. After all, the structure is fixed and the size is known. I guess we will have to charge it some upper bound then.

soroban-env-host/src/host/metered_vector.rs
jayz22: Thanks for the explaination.

jayz22 (approved):

PR #597, opened 2022-12-02
Wasm instruction level calibration

jayz22:
### What

Add routines for doing host calibration at wasm instruction-level granularity . 
Estimate the cpu insn count per input for various wasm instruction types. 

Here are the calibration results (sorted high to low):
Wasm instruction type | Cpu insn count per input
-- | --
CallIndirect | 753
Call | 670
GlobalSet | 88
GlobalGet | 74
Select | 63
MemorySize | 50
I64Store32 | 47
I64Store | 46
I64Store16 | 46
I64Load | 46
I64Store8 | 45
I64Load16S | 45
I64Load32S | 45
I64Load8S | 44
BrTable | 39
I64DivS | 30
I64RemS | 29
I64Rotl | 26
I64Eq | 25
I64Ne | 25
I64LtS | 25
I64GtS | 25
I64LeS | 25
I64GeS | 25
LocalGet | 24
LocalSet | 24
I64Popcnt | 24
I64Add | 24
I64Sub | 24
I64Mul | 24
I64And | 24
I64Or | 24
I64Xor | 24
I64Shl | 24
I64ShrS | 24
I64Rotr | 24
I64Ctz | 23
I64Eqz | 23
Const | 22
LocalTee | 22
I64Clz | 22
Drop | 18
Br | 13

Overall, most of the wasm instructions are similar in cost, and falls into a distinct group:
- Function calling (Call and CallIndirect) are most expensive, over 1 order of magnitude more expensive than others. (~700)
- Global variables (~80)
- Memory-related instructions: MemorySize, I64Store etc. (~50)
- Arithmetic/logic instructions are all very similar. (~25)
- Control instructions (br, drop) are cheapest. (~15)

This suggest we might need a tiered approach to wasmi metering, dividing the instruction types into 3 or 4 buckets each charging a different upper bound cost. 

(Edit)
One exception is the `GrowMemory` operation (not listed above but see attachment), which is very expensive since each page is 64k and it works like growing a dynamic vector. We should consider limit this operation, either by disallowing it upfront (thus limiting contract linear memory size to 64k which is probably too restricting), or specify a new upper bound (say 256k) and only allow `GrowMemory` to happen once. 

Attachment of the full calibration results:
[output.txt](https://github.com/stellar/rs-soroban-env/files/10170613/output.txt)

(Edit - Dec 6)
The table above and the attached full results file have been updated, with the overhead being subtracted from measurements. 

### Why

Our current approach towards metering the wasm runtime (wasmi) treats all instructions indifferently, by counting the interpreter loop and batch-charging the host every 256 loops. 

This work aims to answer the key question: do all instructions cost the same/same ballpark. If the answer is no, then we need to account for that in our metering framework, by introducing more granularity. 

### Known limitations

Not all instructions are measured, only those we care about. Excluded ones: 
- Numeric types other than i64 (we either don't allow them such as FP, or they cost similarly to i64). 
- Some instructions are impossible to exist repeatedly on their own, e.g. end, return. 

~The measurements are only accurate to a certain degree. Some instructions that produce a value on the stack (e.g. `LocalGet`) need to drop the value at the end (So it's really `LocalGet` + `Drop`). The delta is around 10 insns/input.~


jayz22:
@graydon and I discussed earlier and there are a couple of fixes to be done:
1. most of these measured instruction have some overhead that needs to be estimated and subtracted. 
2. add `BrTable`
I will work on these and update the results in the description.

jayz22:
@graydon I've updated the results with overhead subtracted, also added measurement of `BrTable `. Please take another look.

graydon (approved):
LGTM, thanks!

PR #598, opened 2022-12-05
Upgrade crate-git-revision to 0.0.4

brson:
### What

Upgrade crate-git-revision to 0.0.4

### Why

Per https://github.com/stellar/crate-git-revision/pull/9, when used as a git dependency, crate-git-revision 0.0.3 fails to find the correct git directory for soroban-common-env, causing the crate to always be rebuilt.
This upgrade fixes the issue, which I have verified manually.

### Known limitations

The lockfiles here end up with multiple copies of crate-git-revision.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,6 +239,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "crate-git-revision"
+version = "0.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


leighmcculloch (approved):

brson:
Ah, I force-pushed to pick up changes on main just after you approved @leighmcculloch

leighmcculloch (approved):

PR #600, opened 2022-12-07
Remove EnvVal and {Try,}FromVal

graydon:
This is an as-minimal-as-I-can-make-it followup to the `im`-removal PR that removes `EnvVal`, `FromVal` and `TryFromVal`. It leaves `IntoVal` and `TryIntoVal` in place and adds a couple helpers to make it slightly more ergonomic to live in a world without `TryFromVal` and `FromVal`.

As with the `im`-removal change, this is likely to bitrot pretty quickly. It's also not 100% done, there's still a bug to isolate in its handling of the `TryIntoVal`  blanket on `Result`. But at the moment nothing is using that and I've just commented the module out -- I think I am happy to land it unfixed and, worst case, if I can't make the blanket work then just special-case a few specific `Result<Object,HostError>` etc. types in the future. So I think it could theoretically land asap and we deal with the fallout after. It deletes a lot of code that's needless friction / busywork.

After this, my plan is to try to split the remaining fallible `TryIntoVal` from infallible `IntoVal`, and either move the latter to the Guest or SDK crates or at _least_ re-parameterize them from `E:Env` to `E:CheckedEnv` and move the `impl<E:CheckedEnv> Env for E` problem child that caused #579 over to the SDK. To a large measure this is all goundwork for that, though I think any number of future tasks will be simplified by having fewer types, traits and impls kicking around.

(I would also vaguely like to rename these traits, but that's a much simpler PR and something we can discuss more later)


soroban-env-common/src/convert.rs
```diff
@@ -1,6 +1,37 @@
...
+
+/// A free function that exists strictly for readability in contexts where type
+/// ascription is necessary. Exists only to allow you to write
+/// `try_convert_to::<Foo>(x,env)` rather than the less-legible
+/// `<_ as TryIntoVal<E,Foo>>::try_into_val(x,env)`.
+pub fn try_convert_to<V, E: Env, T: TryIntoVal<E, V>>(
+    t: T,
+    e: &E,
+) -> Result<V, <T as TryIntoVal<E, V>>::Error> {
+    <T as TryIntoVal<E, V>>::try_into_val(t, e)
```
leighmcculloch, line 28: Is the goal to also get rid of TryIntoVal? it is a bit odd that `try_convert_to` wraps an `TryIntoVal`. Upon reading the function I would assume it provides a `TryConvert`.

leighmcculloch (approved):
This looks good to me. I _think_ this is going to land okay in the SDK, but I'm wondering what surprises there will be.

soroban-env-common/src/convert.rs
graydon: I'm hoping to eventually rename the traits to match better; I named `try_convert_to` in anticipation of the future name and also because .. if I call it `try_into_val` the diagnostic that fires is always wrong ("do you want to call the trait method" rather than "do you want to import the function you haven't imported yet")

graydon:
Continued in https://github.com/stellar/rs-soroban-env/pull/628

PR #620, opened 2022-12-29
Fix override of panic hook during cross-contract testing

brson:
### What

When configured for testing, in `Host::call_n`, call the new `call_with_suppressed_panic_hook` to catch panics. `call_with_suppressed_panic_hook`:

- performs a one-time override of the global panic hook
- wraps the original panic hook
- tracks a thread local contract-call recursion count
- when the recursion count is zero, forwards to the original panic hook, maintaining the original hook's behavior
- when the recursion count is non-zero, does nothing

The `call_with_suppressed_panic_hook` function is extracted out of the `call_n` function for easier reading, and gets its own module to avoid many extra `cfg`s.

### Why

The existing code is written to suppress the default printing of the global panic hooks when calling contracts in testing mode; later performing a single panic once all contract executes.

The existing code has race conditions that affect tests running in parallel, causing some tests to not report anything to stdout, and probably causing some tests to fail to suppress the recursive panics.

### Known limitations

This is hard to write tests for and I think not worth the effort. Please read carefully.

Unlike the existing code, this version never restores the original panic hook that does not contain the wrapper. Since this is only for test configurations that seems reasonable.

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -824,7 +824,7 @@ impl Host {
                 let panic = frame.panic.clone();
                 return self.with_frame(Frame::TestContract(frame), || {
                     use std::any::Any;
-                    use std::panic::{catch_unwind, set_hook, take_hook, AssertUnwindSafe};
+                    use std::panic::AssertUnwindSafe;
                     use crate::xdr::ScUnknownErrorCode;
                     type PanicVal = Box<dyn Any + Send>;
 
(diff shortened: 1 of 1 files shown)


brson:
I have tested this manually by running multiple panicking cross-contract-call tests in parallel.

graydon (approved):
LGTM! Nice catch, subtle fix. The only change I might suggest is to reduce the use of the term "recursive" here; IIRC the `call_n` function is the path we enter even when doing a top-level invocation of a contract. Not just contract-to-contract, but also transaction-to-contract (in which "recursion" is a potentially confusing term to use).

brson:
@graydon You are right. I removed the emphasis on recursion.

PR #624, opened 2023-01-06
Remove unnecessary value traits for the map.

dmkozh:
### What

Remove unnecessary value traits for the map.

### Why

Values don't need to be comparable.

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/host/metered_map.rs
+++ b/soroban-env-host/src/host/metered_map.rs
@@ -12,7 +12,7 @@ pub struct MeteredOrdMap<K, V, Ctx>
 where
     K: MeteredClone,
     V: MeteredClone,
-    Ctx: AsBudget + Compare<K, Error = HostError> + Compare<V, Error = HostError>,
+    Ctx: AsBudget + Compare<K, Error = HostError>,
 {
     pub(crate) map: Vec<(K, V)>,
     ctx: PhantomData<Ctx>,
@@ -22,7 +22,7 @@ impl<K, V, Ctx> Clone for MeteredOrdMap<K, V, Ctx>
 where
     K: MeteredClone,
     V: MeteredClone,
-    Ctx: AsBudget + Compare<K, Error = HostError> + Compare<V, Error = HostError>,
+    Ctx: AsBudget + Compare<K, Error = HostError>,
 {
     fn clone(&self) -> Self {
         Self {
@@ -36,7 +36,7 @@ impl<K, V, Ctx> MeteredOrdMap<K, V, Ctx>
 where
     K: MeteredClone,
     V: MeteredClone,
-    Ctx: AsBudget + Compare<K, Error = HostError> + Compare<V, Error = HostError>,
+    Ctx: AsBudget + Compare<K, Error = HostError>,
 {
     fn charge_new<B: AsBudget>(&self, size: usize, b: &B) -> Result<(), HostError> {
         b.as_budget().charge(CostType::MapNew, size as u64)
@@ -61,7 +61,7 @@ impl<K, V, Ctx> Default for MeteredOrdMap<K, V, Ctx>
 where
     K: MeteredClone,
     V: MeteredClone,
-    Ctx: AsBudget + Compare<K, Error = HostError> + Compare<V, Error = HostError>,
+    Ctx: AsBudget + Compare<K, Error = HostError>,
 {
     fn default() -> Self {
         Self {
@@ -81,7 +81,7 @@ impl<K, V, Ctx> MeteredOrdMap<K, V, Ctx>
 where
     K: MeteredClone,
     V: MeteredClone,
-    Ctx: AsBudget + Compare<K, Error = HostError> + Compare<V, Error = HostError>,
+    Ctx: AsBudget + Compare<K, Error = HostError>,
 {
     pub fn new(ctx: &Ctx) -> Result<Self, HostError> {
         ctx.as_budget().charge(CostType::MapNew, 1)?;


graydon:
This is not correct. Values do need to be comparable -- not for the purpose of looking things up in the map, but for the purpose of maps themselves being ordered. To order one map relative to another, the code will look at keys _and values_.

dmkozh:
> This is not correct. Values do need to be comparable -- not for the purpose of looking things up in the map, but for the purpose of maps themselves being ordered. To order one map relative to another, the code will look at keys _and values_.

Removing this and adding non-comparable values doesn't break anything, so I'm not convinced this is being used now... More importantly, that's unexpected interface for an ordered map. I'm not aware of any implementations having such requirement, including standard C++ and Rust ordered maps.

jayz22:
> Removing this and adding non-comparable values doesn't break anything

Both the env and the sdk?

dmkozh:
> > Removing this and adding non-comparable values doesn't break anything
> 
> Both the env and the sdk?

Hmm, I haven't tested this with SDK yet, so if it breaks, then we have something uncovered on the host side. In general I'd try to limit the amount SDK-exclusive features and we definitely need some test coverage for them.

graydon:
> Removing this and adding non-comparable values doesn't break anything, so I'm not convinced this is being used now... 

It is called from [this compare impl](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/metered_map.rs#L331-L348) which is called by [this path](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/comparison.rs#L49) of host object comparison. 

The places in this PR that you removed it are superfluous / documentary constraints -- they're struct definitions or impl blocks that don't happen to use the ordered-values constraint, and so removing them does not break anything. But you could remove the ordered-keys constraint also -- in the struct case you can remove all the constraints! -- and IMO it's misleading to do so. Those constraints tell the reader something, which is why we repeat the same type parameter constraints on all generic blocks here.

(If you were to remove the ordered-value constraint from the impl block that actually does the comparison, you'll see that it won't compile anymore.)

> More importantly, that's unexpected interface for an ordered map. 

All value types in soroban are a bit "enriched" by comparison to types in many other systems: every value is totally ordered, in addition to being immutable and serializable! These aren't requirements on most maps either. This is intentional, since the richer semantics make code less error prone, more interoperable, and subject to greater reuse (everyone has the same enriched semantics and they're implemented in the host).

dmkozh:
> But you could remove the ordered-keys constraint also -- in the struct case you can remove all the constraints! -- and IMO it's misleading to do so.

I'm pretty confused here... What is the value of annotating the struct besides the fact that non-comparable values won't compile now even though they would not be compared between each other? It seems like in a general case it would be nicer to let the users store arbitrary values and only require them to implement specific traits in case if they're used in the appropriate context. E.g. if a map ever gets a `Debug` implementation, would you also require `Debug` trait on keys/vals from every other trait?

> All value types in soroban are a bit "enriched" by comparison to types in many other systems: every value is totally ordered, in addition to being immutable and serializable! 

The reason I stumbled upon this is that I wanted to create a map that stores `RawVal`s and they conceptually shouldn't be comparable (or should they?). In my use case I also don't ever need to compare this map with anything. FWIW the use case is temporary contract storage.

graydon:
> I'm pretty confused here... What is the value of annotating the struct besides the fact that non-comparable values won't compile now even though they would not be compared between each other?

The value is in twofold:
  - Documentation and setting expectations -- it illustrates to the user what the expected use of the type is
  - Allowing `impl` blocks to use those expectations as they like, without adding or removing constraints to maintain absolute minimality on a change-by-change basis.

> It seems like in a general case it would be nicer to let the users store arbitrary values and only require them to implement specific traits in case if they're used in the appropriate context. E.g. if a map ever gets a `Debug` implementation, would you also require `Debug` trait on keys/vals from every other trait?

This map (`MeteredOrdMap`) is not a generally-useful "arbitrary values" library data structure for use outside the context it's intended for: it's an immutable sorted-vector map that does metering (via a host) on every operation; this is a very special-purpose data structure. It's designed to stay on the host, and only ever carry two types of values: `RawVal` and `Arc<LedgerEntry>`. Both of which are totally ordered.

> The reason I stumbled upon this is that I wanted to create a map that stores `RawVal`s and they conceptually shouldn't be comparable (or should they?). In my use case I also don't ever need to compare this map with anything. FWIW the use case is temporary contract storage.

`RawVal` is conceptually totally ordered. Its non-object-handle cases compare directly without helpers, and its object-handle cases are ordered by calling back into the host storing the objects (by deep structural comparison of the objects). Map objects in the host are all represented as `Vec<(RawVal,RawVal)>`, ordered by the first element of the pair, so `RawVal` had better be ordered!

This structure should not be used for storage inside guest (WASM) contract code at all. In native contracts (eg. the builtin token) it can be used as you describe, and the ordering constraints shouldn't be blocking use in that context.

dmkozh:
>  It's designed to stay on the host, and only ever carry two types of values: RawVal and Option<Rc<LedgerEntry>>. Both of which are totally ordered.

Why narrow the use case so much? My use case would benefit from this, as it needs exactly the same semantics, besides having different key/value types (i.e. I want to get the budget metering and rollbacks and I don't see why not reuse the metered map we already have).

> and its object-handle cases are ordered by calling back into the host storing the objects (by deep structural comparison of the objects).

That's *not* the behavior I'd want. In my case if I ever would want to compare RawVals, I'd compared them as ints, as I don't care about the value order. But a semantically correct thing would be to not compare the values at all.

I've actually noticed that the `Compare` is actually implemented, but only for `CheckedEnv` and I wanted to use just `Budget`. I guess I could just hack something to just make this compile then.

I realize that's probably my issue, but for me this design led to confusion and hard-to-parse compiler errors. I'm not sure if specialized use cases are good enough reason to not use the common data structure design patterns, especially given that the interface/functionality is mostly the same as for a regular ordered map.

graydon:
> Why narrow the use case so much? My use case would benefit from this, as it needs exactly the same semantics, besides having different key/value types (i.e. I want to get the budget metering and rollbacks and I don't see why not reuse the metered map we already have).

I'm not sure I understand; what key and val types are you storing? I thought you were storing `RawVal`s?

> > and its object-handle cases are ordered by calling back into the host storing the objects (by deep structural comparison of the objects).
> 
> That's _not_ the behavior I'd want. In my case if I ever would want to compare RawVals, I'd compared them as ints, as I don't care about the value order. But a semantically correct thing would be to not compare the values at all.

Comparing object handles as integers is certainly wrong; it gives you an order that's not meaningful at all. I agree the semantically correct thing to do usually is not compare them (at least not without a `CheckedEnv`); this (along with fallibility of comparison) is why `RawVal` does not implement standard rust comparison traits. But you _can_ compare them when you need to, such as when using a `RawVal` as a key for a map: such a `RawVal` might be a handle to another object, which you must then order with respect to other keys in the map.

> I've actually noticed that the `Compare` is actually implemented, but only for `CheckedEnv` and I wanted to use just `Budget`. I guess I could just hack something to just make this compile then.

It needs a full `CheckedEnv` to perform comparison, because it needs to resolve object handles.

> I realize that's probably my issue, but for me this design led to confusion and hard-to-parse compiler errors. I'm not sure if specialized use cases are good enough reason to not use the common data structure design patterns, especially given that the interface/functionality is mostly the same as for a regular ordered map.

I apologize. I did not mean to deny you the option of using this type, just explain my thinking about its current organization and design choices, what its assumptions are. Those can change! I had not foreseen anyone else besides the host wanting this functionality, and was surprised to hear you did. Could you tell me more about the use-case? If I understand, it is a map with some combination of:

1. non-`RawVal` keys
2. `RawVal` values
3. No access to a `CheckedEnv`, but access to a `Budget`

Is that right? What's being stored and where? Is this in the builtin contract, or a native test, or some other context?

dmkozh:
> I'm not sure I understand; what key and val types are you storing? I thought you were storing RawVals?

The keys are comparable (it's a tuple of contract id and `ScVal`), and the value is `RawVal` (see [draft](https://github.com/dmkozh/rs-soroban-env/blob/e9db1dda061d89a525dbfcd5b3e978ce06644503/soroban-env-host/src/storage.rs#L24)).

> Comparing object handles as integers is certainly wrong; it gives you an order that's not meaningful at all. 

Exactly; that's why I would prefer to not compare them at all. But if I had to do the comparison, I would compare them as ints, as I'm not actually interested in value ordering and I'm fine with any arbitrary order.

> Could you tell me more about the use-case? 

The use case is temporary storage, which works almost exactly the same as the regular storage (i.e. every contract owns its storage, it needs to be metered etc.). The only difference is that we don't flush anything to ledger and hence can benefit from just storing RawVals. See the [draft](https://github.com/dmkozh/rs-soroban-env/blob/e9db1dda061d89a525dbfcd5b3e978ce06644503/soroban-env-host/src/storage.rs#L310) for details.

> No access to a CheckedEnv, but access to a Budget

I'm not sure about CheckedEnv access; I don't understand the new design well enough yet to answer to this for certain. I've used the current storage as template and it does not point at CheckedEnv. I'd like to avoid any potential additional costs if possible (it *seems* like if I don't actually compare it, I shouldn't run into `RawVal` comaprison cases, but maybe there are other performance implications).

PR #625, opened 2023-01-06
Reduce TryInto{,Val} to a blanket, remove several redundant impls

graydon:
This is Yet Another Attempt at cleaning up the {Try,}{From,Into}{,Val} traits. The logic at work in this attempt is:

1. Try to follow the rust stdlib design where `Into` variants are _only_ defined as blankets that delegate to `From` variants. Specifically this means:
  - We keep the `Into` variants around as syntactic convenience (so we can use them as constraints in impls and call `.into()` or `.into_val()` or `.try_into_val()` when needed), which avoids the problem in earlier attempts of having to contort code to make up for only one direction being defined. Both are allowed and you can write whichever makes sense in a constraint or call.
  -  But: we do not define, or encourage users to define, any `Into` variants besides the blankets. **Only ever write `From` variants in our code, UDT-generated or otherwise**. This mostly accomplishes the original goal of "less conversion boilerplate".
2. Unfortunately, since all our impls are generic in `E:Env`, we can't have any blanket `TryFromVal` that delegates to `FromVal` or `From`: those are considered colliding by the coherence rules. So we can't _totally_ follow the rust stdlib design that does such delegation; we have to write a few redundant imlpls if we want `TryFromVal` for all types.
3. We could just ignore `TryFromVal` in cases we don't need it -- tell users to call `FromVal` or `From` as appropriate -- but the UDT code calls `TryFromVal` on everything, because it has no idea which types need it. So we do need the otherwise-redundant `TryFromVal` impls.

I haven't gone through the SDK yet to see how this approach fares there, but it appears to be not-terribly-disruptive in the env, so I'm hopeful. Will do the SDK next.

There is also a remaining design question to answer about how much to chain conversions of object types via `Object` specifically vs. converting to and from `RawVal` directly. But I think this is a step in the right direction.

(Incidentally I'm also thinking of renaming `TryFromVal` into `TryFromVia`, because the `Val` suffix really isn't meaningful here and the relevant difference is that it uses an auxiliary environment argument to help with the conversion. WDYT?)

### DIFF
--- a/soroban-env-common/src/array.rs
+++ b/soroban-env-common/src/array.rs
@@ -27,44 +27,26 @@ impl<E: Env, const N: usize> TryFromVal<E, RawVal> for [u8; N] {
     }
 }
 
-impl<E: Env, const N: usize> TryIntoVal<E, [u8; N]> for RawVal {
-    type Error = <[u8; N] as TryFromVal<E, RawVal>>::Error;
-    #[inline(always)]
-    fn try_into_val(self, env: &E) -> Result<[u8; N], Self::Error> {
-        <_ as TryFromVal<_, _>>::try_from_val(env, self)
+impl<E: Env> TryFromVal<E, &[u8]> for RawVal {
+    type Error = Status;
+    fn try_from_val(env: &E, v: &[u8]) -> Result<Self, Self::Error> {
+        Ok(env.bytes_new_from_slice(v)?.to_raw())
     }
 }
 
-impl<E: Env> TryIntoVal<E, RawVal> for &[u8] {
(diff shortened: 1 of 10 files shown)


leighmcculloch (approved):
Lgtm, although short, the complexity of this change is extraordinarily high so I don't have a complete mental picture for the impact these changes have. But nothing looks obviously out of sorts.

I'd like to look at the SDK change before we merge this if that is okay.

graydon:
re-attempted in https://github.com/stellar/rs-soroban-env/pull/628

PR #628, opened 2023-01-12
Reform TryFromVal to reduce number of impls.

graydon:
## Overview

This is a "minimal" version of the change I've been attempting to make to `TryFromVal` and friends for the past few weeks.

It defers mucking with the error types to a later PR (I have an idea for how to do that but it doesn't _have_ to happen here). The minimal changes here are just to get rid of so many impls to improve maintainability -- net removal of about 300 lines -- while expanding and systematizing overall coverage (the error stuff will also make things a bit more systematic).

The changes here are:

  - Remove non-Try {Into,From}Val. 
  - Blanket-impl TryIntoVal to use TryFromVal, removing all manual TryIntoVal impls.
  - Switch TryFromVal to take &T everywhere, removing the T-vs-&T duplication.
 
It seems to hold together, so I'm going to start on the SDK changes to match. Feedback welcome.

## Discussion

In the resulting code, I think nobody should implement anything except TryFromVal<T>, UDT code or otherwise.

You may find it easier to review just by opening the finished files and eyeballing them for "this looks nice". The specific changes are all just the 3 above applied mechanically, plus type-inference-massaging to match in a few cases.

The removal of non-Try {Into,From}Val might seem a little aggressive, but if you look at what it's used for and the bodies of the impls, I think it makes sense: most of the projection paths have a guest-side fallibility path anyway (eg. when given an object handle of the wrong type) for the few remaining injection paths that are  logically infallible (eg. "you should always be able to inject a u64 to a host object") it's not a lot of extra work in the guest to call .unwrap() or ? as appropriate, and I think it's fairly likely we're actually going to re-parameterize all these over EnvBase or merge Env and CheckedEnv, in which case all host methods will return a Result anyways plumb it out to the user where they can do an infallible-unwrap whenever `E::Error=Infallible`. Either way, removing them cuts the set of impls in half and I think the reduction in clutter is a net win in comprehensibility, even for users.

### DIFF
--- a/soroban-env-common/src/array.rs
+++ b/soroban-env-common/src/array.rs
@@ -1,14 +1,15 @@
 use crate::xdr::ScHostValErrorCode;
-use core::borrow::Borrow;
 use stellar_xdr::ScObjectType;
 
-use crate::{ConversionError, Env, Object, RawVal, RawValConvertible, Status, TryFromVal};
+use crate::{
+    ConversionError, Env, Object, RawVal, RawValConvertible, Status, TryFromVal, TryIntoVal,
+};
 
 impl<E: Env, const N: usize> TryFromVal<E, RawVal> for [u8; N] {
     type Error = Status;
 
-    fn try_from_val(env: &E, val: impl Borrow<RawVal>) -> Result<Self, Self::Error> {
-        let val = *val.borrow();
+    fn try_from_val(env: &E, val: &RawVal) -> Result<Self, Self::Error> {
+        let val = *val;
         if !Object::val_is_obj_type(val, ScObjectType::Bytes) {
             return Err(ScHostValErrorCode::UnexpectedValType.into());
         }
@@ -27,8 +28,8 @@ impl<E: Env, const N: usize> TryFromVal<E, RawVal> for [u8; N] {
 impl<E: Env> TryFromVal<E, RawVal> for Vec<u8> {
     type Error = Status;
 
-    fn try_from_val(env: &E, val: impl Borrow<RawVal>) -> Result<Self, Self::Error> {
-        let val = *val.borrow();
+    fn try_from_val(env: &E, val: &RawVal) -> Result<Self, Self::Error> {
+        let val = *val;
         if !Object::val_is_obj_type(val, ScObjectType::Bytes) {
             return Err(ScHostValErrorCode::UnexpectedValType.into());
         }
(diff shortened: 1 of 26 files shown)


soroban-env-common/src/env_val.rs
```diff
@@ -12,22 +13,27 @@ use super::{
...
-    fn try_into_val(self, env: &E) -> Result<V, Self::Error>;
+    fn try_from_val(env: &E, v: impl Borrow<V>) -> Result<Self, Self::Error>;
 }
 
-pub trait FromVal<E: Env, V>: Sized {
-    fn from_val(env: &E, v: V) -> Self;
-}
+// Blanket-impl uses of TryIntoVal to TryFromVal, so that we
+// only ever have to impl TryFromVal.
+impl<E: Env, T, U> TryIntoVal<E, T> for U
```
leighmcculloch, line 28: We used to have a blanket impl like this, and it had to be removed at some point for another reason. I think it related to UDTs, but I can't remember exactly. If we can have it again sounds good to me, but I wonder if somewhere downstream we'll see an issue with this. It's also possible the blanket impl was slightly different and that this doesn't face the same problems.

soroban-env-common/src/env_val.rs
```diff
@@ -12,22 +13,27 @@ use super::{
     Env,
 };
 
-pub trait IntoVal<E: Env, V>: Sized {
-    fn into_val(self, env: &E) -> V;
```
leighmcculloch, line 16: I'm suspicious of getting rid of `IntoVal`. The SDK uses `IntoVal` in a few locations for converting into types that should always succeed, such as converting from a fixed length array into a `RawVal`. Maps and Vecs require their T, K, and V types to be `IntoVal<E, RawVal>`.

Will we use `TryIntoVal` instead now? What will we do with the failure case?

Will we create a new trait for that into rawval case instead?

soroban-env-common/src/array.rs
```diff
@@ -1,24 +1,20 @@
...
-// TODO: these conversions happen as RawVal, but they actually take and produce
-// Objects; consider making the signatures tighter.
+use crate::{ConversionError, Env, Object, RawVal, RawValConvertible, Status, TryFromVal};
 
 impl<E: Env, const N: usize> TryFromVal<E, RawVal> for [u8; N] {
     type Error = Status;
 
-    fn try_from_val(env: &E, val: RawVal) -> Result<Self, Self::Error> {
+    fn try_from_val(env: &E, val: impl Borrow<RawVal>) -> Result<Self, Self::Error> {
+        let val = *val.borrow();
```
leighmcculloch, line 11: I think using Borrow<> is a great idea. One thing I am curious about is if this will result in code size changes. I saw using refs for some conversions increase WASM binary size, although not significantly. I don't think we should worry about this much, just wanted to share it.

leighmcculloch (commented):
Overall looks good. I'm concerned about one thing. My other comments are just informational.

leighmcculloch (approved):
I'm fine with moving ahead with this regardless of my concern above. I assume the IntoVal can be trivially re-added if needed.

soroban-env-common/src/array.rs
jayz22: Looking from this change set. It appears the usages of `Borrow<T>` are exactly like a `&T`, i.e. there is no additional `impl Borrow<T>` for any type (other than the implicit `T`, `&T`, `Box<T>` etc)? So I guess from the binary size perspective there should not be a difference.

I wonder then, what problem doe `impl Borrow<T>` (instead of `AsRef<T>`) solve?

jayz22 (approved):
Looks good to me.  It appears the `impl borrow<T>` is the main trick? I'm just curious how does the borrow trait solve problems where previous attempts couldn't.

soroban-env-common/src/array.rs
leighmcculloch: `Borrow` and `AsRef` do similar things but are subtly different. `Borrow` is implemented on all types automatically. 

Part of the contract of Borrow is that `Eq`, `Ord`, and `Hash` have to work identically for `T` and for the borrowed value of T.

AsRef does not have the same requirements. Any type can offer an AsRef of another type, even if those two types are very very different.

There's some details about it here: https://doc.rust-lang.org/std/borrow/trait.Borrow.html

jayz22: Right, I get those differences. 
I don't think (at least from looking at this change set) that we are replying on any other logic than those automatic implementations. So `*val.borrow()` is just `&val`, which should lead to no difference in compiled binary code (and also wasm code size)?

jayz22: My question was, if the `impl Borrow` are just automatic implementations, and we are only passing in `T` and `&T` as arguments in most cases, what problems did `impl Borrow` solve, there must be corner cases?

soroban-env-common/src/env_val.rs
graydon: So .. the "always succeeds" part here is .. the problematic part. All these impls are currently parameterized over `E:Env` which allows you to call host methods that "always succeed", except in reality they can fail (at minimum "out of gas" but also other things like bad handles -- stuff that turns into a `HostError` and is then escalated to a panic by `impl Env for CheckedEnv` or a VM trap by the dispatch function). When we get rid of the `impl Env for CheckedEnv` (or at least remove it form visibility in `Host`) we will either have to duplicate all of them, but in fallible form, for the `CheckedEnv` case (surfacing errors) or else certain conversions (eg. the current u64-to-object injection) will just disappear from the `Host` context (eg. won't be usable in the builtin contract).

What I propose happen (and am doing groundwork for here) is that _every_ conversion that uses the environment is _always_ considered fallible in the signatures-and-traits sense (returning a `Result<>`), and we merge `Env` and `CheckedEnv`, making the `Env` trait one where all host functions return `Result<>` with an associated type `Env::Error`, and `Guest::Error = Infallible`. That'll be a subsequent PR though.

soroban-env-common/src/array.rs
graydon: The use of `impl Borrow<T>` here is specifically (and solely) to allow us to write a single impl of TryFromVal that will work for _either_ T or &T arguments. We previously had -- in a variety of places, some impls but not all -- duplicate impls for both of those cases. The whole point of this PR is "reduce the number of impls".

jayz22: I see, thanks for the clarification.

graydon: @jayz22 I actually stared at this answer for a few minutes and wondered again how much pain it'd be to go with just &T everywhere, and I made a change to do that and it's .. not too bad. Actually lets me remove some of the additional type-inference helpers I put in before. It's very slightly less ergonomic when using `T::try_from_val` since you have to `&arg` the arguments, but for the more common case of calling `x.try_into_val()` it makes no difference since I already have the signature of `try_into_val(&self, e)` having that `&self` parameter -- rust will auto-ref the self arg.

Considering this has even less cognitive load and seems better behaved in terms of type inference I'm going to try going along with it. I think the ergonomic hit will not be so bad. @leighmcculloch mentioned there might be a slight codesize hit which I'll .. try to quantify, but again at this point I'm primarily concerned with getting this code under _control_ so it makes sense, is smaller, more consistent and maintainable.

PR #633, opened 2023-01-19
Env error cleanup

graydon:
This is one of the final steps in fixing https://github.com/stellar/rs-soroban-env/issues/579, wherein we remove `impl Env for CheckedEnv` and in fact merge `CheckedEnv` into `Env`, removing `CheckedEnv` entirely. All `Env` methods now return `Result<T,Env::Error>` and there is some new plumbing to interoperate with those `Result`s differently depending on whether the `Error` is `Infallible` or `HostError`. The panic-escalation path is feature-flagged _off_ entirely in the non-testutils host configuration.

This shouldn't land yet since it's still a _little_ speculative. I need to adapt the SDK to it first and check that everything works there and doesn't regress much in codesize. But this is the roughly shape I was aiming for.


graydon:
closing in favor of #638

Issue #463, opened 2022-09-21
Consider alternative forms for Symbol


leighmcculloch:
`Symbol` is a 60-bit encoding of a 10 character string with a limited character set of `a-zA-Z0-9_`.

The value is stored in a u64 and so is efficient to pass around.

We use it in a lot of places:
- Function names.
- Type names, field names.
- Enum variant names.
- As keys to data storage.
- so on

However, the 10 character limit is at times annoying. It can sometimes cause us to write code that looks like a human error. e.g. `contrct_id`.

We should consider other ways we could encode strings into Symbol's that would give us longer strings, more descriptive function names, variants, type names, etc.

cc @tomerweller @graydon

leighmcculloch:
One option I've heard discussed is something similar to how Ethereum uses the first 4-bytes of a hash of function names.

We could make `Symbol` a 60-bit hash of the string.

This would work in contexts where we could easily detect collisions, such as function names, but would be harder enforce in broader contexts like type names, enum variant names, etc.

leighmcculloch:
Another option I've heard discussed (cc @graydon?) is that we store the full length strings in a section of the WASM and reference them somehow. The host would be able to use those references to load the full string out of the data section of the WASM so we would only ever be transmitting to/from the host a handle.

tomerweller:
Thanks. It seems like a lot of developers are running into 10-char limit issue, based on discord questions

graydon:
I think there are several viable options here but they all have certain challenges:

1. We use hashes. This makes symbols useless for diagnostics / debugging, which I think is actually fairly serious. At present they can be used fairly casually and informatively, provide useful meaning. We also have to deal with collision probability separate from the issue of just organizing namespaces. Two names might collide / be equal in hash-space even if they look superficially different in source code.
2. We put names in data sections of contracts and then reference them by linear-memory address or wasm global number or whatever. This makes symbols _relative_ to a specific wasm, which has lots of problems: they can't be compared in any context outside of host functions while their owning guest VM is active (not in the guest, not in stellar-core outside of an activation, not in XDR function-invocation transaction inputs, not when captured in error buffers or emitted-events). It also represents a bit of an implicit information-flow / trust-risk between wasms, if wasm A passes a symbol to wasm B, it's implicitly telling wasm B it should read bytes out of A's linear memory, which might be cleverly malformed in such a way as to confuse B.
[...]

PR #644, opened 2023-01-30
Change `Storage`, `Footprint`, `Snapshot` to return `dyn Error`

jayz22:
### What

Change `Storage`, `Footprint`, `Snapshot` to return `dyn Error`.
Add example of how to retrieve custom error. 
Add a new trait to convert from `dyn Error` -> `HostError`;

### Why

The `Storage` and `Footprint` depends on external `Snapshot` which may need to define its own errors (for example https://github.com/stellar/soroban-tools/pull/269/files#diff-a218fbb339f2a7838711dbd1e77eecb588242d37b1bad5f44df33cf5d11028a4R39-R58).  

This PR changes those types to work with generic errors and adds the functionality to convert them into a HostError (by logging the error string as a `DebugEvent`). 

cc @2opremio @leighmcculloch 

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/storage.rs
+++ b/soroban-env-host/src/storage.rs
@@ -7,7 +7,7 @@
 //!   - [Env::put_contract_data](crate::Env::put_contract_data)
 //!   - [Env::del_contract_data](crate::Env::del_contract_data)
 
-use std::rc::Rc;
+use std::{error::Error, rc::Rc};
 
 use soroban_env_common::Compare;
 
@@ -52,8 +52,8 @@ impl Compare<AccessType> for Budget {
 /// A helper type used by [FootprintMode::Recording] to provide access
 /// to a stable read-snapshot of a ledger.
 pub trait SnapshotSource {
-    fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, HostError>;
-    fn has(&self, key: &LedgerKey) -> Result<bool, HostError>;
+    fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, Box<dyn Error>>;
+    fn has(&self, key: &LedgerKey) -> Result<bool, Box<dyn Error>>;
 }
 
 /// Describes the total set of [LedgerKey]s that a given transaction
@@ -193,9 +193,12 @@ impl Storage {
                 // In recording mode we treat the map as a cache
                 // that misses read-through to the underlying src.
                 if !self.map.contains_key::<LedgerKey>(key, budget)? {
+                    let val = src
+                        .get(key)
+                        .map_err(|_| HostError::from(ScHostStorageErrorCode::MissingKeyInGet))?;
                     self.map = self.map.insert(
                         Rc::new(key.metered_clone(budget)?),
-                        Some(Rc::new(src.get(key)?)),
+                        Some(Rc::new(val)),
                         budget,
                     )?;
                 }
@@ -284,7 +287,9 @@ impl Storage {
                 match self.map.get::<LedgerKey>(key, budget)? {
                     Some(None) => Ok(false),
                     Some(Some(_)) => Ok(true),
-                    None => src.has(key),
+                    None => src
+                        .has(key)
+                        .map_err(|_| ScHostStorageErrorCode::MissingKeyInGet.into()),
                 }
             }
             FootprintMode::Enforcing => {
@@ -395,9 +400,17 @@ mod test_footprint {
 
 #[cfg(test)]
 pub(crate) mod test_storage {
+    use core::fmt::Debug;
     use std::collections::BTreeMap;
 
-    use soroban_env_common::xdr::ScUnknownErrorCode;
+    #[derive(Debug)]
+    pub(crate) struct MyError;
+    impl std::error::Error for MyError {}
+    impl std::fmt::Display for MyError {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            <MyError as Debug>::fmt(self, f)
+        }
+    }
 
     use super::*;
     #[allow(dead_code)]
@@ -409,15 +422,15 @@ pub(crate) mod test_storage {
         }
     }
     impl SnapshotSource for MockSnapshotSource {
-        fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, HostError> {
+        fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, Box<dyn Error>> {
             if let Some(val) = self.0.get(key) {
                 Ok(val.clone())
             } else {
-                Err(ScUnknownErrorCode::General.into())
+                Err(Box::new(MyError))
             }
         }
 
-        fn has(&self, key: &LedgerKey) -> Result<bool, HostError> {
+        fn has(&self, key: &LedgerKey) -> Result<bool, Box<dyn Error>> {
             Ok(self.0.contains_key(key))
         }
     }
--- a/soroban-env-host/src/storage.rs
+++ b/soroban-env-host/src/storage.rs
@@ -74,7 +74,7 @@ impl Footprint {
         key: &LedgerKey,
         ty: AccessType,
         budget: &Budget,
-    ) -> Result<(), HostError> {
+    ) -> Result<(), Box<dyn Error>> {
         if let Some(existing) = self.0.get::<LedgerKey>(key, budget)? {
             match (existing, ty.clone()) {
                 (AccessType::ReadOnly, AccessType::ReadOnly) => Ok(()),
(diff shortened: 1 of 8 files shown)


soroban-env-host/src/storage.rs
```diff
@@ -193,9 +193,12 @@ impl Storage {
                 // In recording mode we treat the map as a cache
                 // that misses read-through to the underlying src.
                 if !self.map.contains_key::<LedgerKey>(key, budget)? {
+                    let val = src
+                        .get(key)
+                        .map_err(|_| HostError::from(ScHostStorageErrorCode::MissingKeyInGet))?;
```
leighmcculloch, line 198: Losing the original error here seems not ideal. Could we take the original error, generate text for it by using it's Display impl, and add that as a DebugEvent to the returned HostError?

soroban-env-host/src/storage.rs
```diff
@@ -409,16 +425,26 @@ pub(crate) mod test_storage {
...
     }
+
+    #[test]
+    fn test_retrieve_error() {
+        let snapshot = MockSnapshotSource(BTreeMap::new());
+        let lk = LedgerKey::Account(crate::xdr::LedgerKeyAccount {
+            account_id: util::generate_account_id(),
+        });
+        let err = snapshot.get(&lk).expect_err("key not exist");
+        assert_eq!(err.downcast_ref::<SnapshotError>().unwrap().code, 1234);
```
leighmcculloch, line 448: 👀 That's cool.

soroban-env-host/src/storage.rs
jayz22: Hmm I agree it would be nice to retain the original error. 
It would be difficult because the storage does not own a host and it needs it to add a DebugEvent. 
Changing the `Storage` to also return a `Box<dyn Error>` might be the more plausible way but that requires bigger change in the host. I will experiment with that.

jayz22: @leighmcculloch I have updated both `Storage` and `Footprint` to return `Box<dyn Error>` such that the upstream error is preserved. 
Also on the host side, I added the new functionality to map a `Error` into a `HostError`:  
- If the original error was a `HostError`, retrieve it. 
- Otherwise, it creates a `Bytes` object from its `fmt` string, and creates a debug event from it. (The `Bytes` representation can be improved later with better types, e.g. https://github.com/stellar/rs-soroban-env/issues/463)
Would love to have @graydon's opinion as well.

leighmcculloch (approved):
This looks good to me. I'd like @graydon to take a look too.

graydon:
Hmm, I'm not thrilled with this. It seems like a lot of churn, but more importantly it leads us away from uniform error reporitng by status/error codes, which we've repeatedly discovered is something we actually want, because they can cross the host/guest interface can also survive the termination of a VM and host, and de-allocation of its dynamic host objects.

I would prefer to just revisit the XDR allocation of error codes (we've been wanting to do that for a long time) and make sure they reasonably cover the cases storage providers need to report. It looks like the motivating case here is just reporting a more-detailed (base64) XDR error as a less-detailed one (unknown/xdr). I don't know that that's a problem -- HostError will already capture a stacktrace for context when it receives it from Storage, so we'll know "the storage module had an XDR error" which seems fairly close to the same fidelity you'll get with this change. And if it's not I suspect it's solvable with a more-precise XDR-related error code.

leighmcculloch:
> HostError will already capture a stacktrace for context when it receives it from Storage

How do we do this from within a `SnapshotSource` like the one @2opremio is implementing here?

https://github.com/stellar/soroban-tools/blob/55a27d1/cmd/soroban-rpc/lib/preflight/src/lib.rs#L33-L70

graydon:
> How do we do this from within a SnapshotSource like the one @2opremio is implementing here?

 Just do `SomeStatus.into()`: https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/error.rs#L109

No `Host` is required to build a `HostError`, just a `Status`. It captures a stacktrace on construction. It won't capture a Host's `DebugEvent` buffer if constructed that way, though if that difference matters we can probably also add a helper to `Host` that attaches such a buffer to an already-constructed event-buffer-less `HostError` returning on public-interface paths.

leighmcculloch:
> It captures a stacktrace on construction.

We're not just interested in the stack trace but what the error was, such as text from the error. Is there a way to include that? 

> add a helper to `Host` 

The `SnapshotSource` doesn't have a reference of a `Host` in this case. Do we need to change the SnapshotSource trait to pass the `Host` down? Or another idea?

graydon:
> We're not just interested in the stack trace but what the error was, such as text from the error. Is there a way to include that?

Well .. mainly I'm inclined to say "no" to that question, because we _explicitly removed_ error strings from HostError so that we wouldn't have to plumb them around, worry about ownership and encoding of them (and costs since errors are supposed to keep working even when budgets are exhausted) and worry about people pattern-matching on the string content of errors, as people were already starting to do. So I'm a bit skeptical about the idea that we should add them back again.

I guess I wonder why you're after "text from the error", exactly. The error example you pointed to has a few fairly predictable cases that have no accompanying text content. We could just .. add error codes for the couple cases currently routed into insufficiently-informative codes, no? The fact that we haven't added or changed error codes in the XDR at all for months while we've changed the code to have lots of new error paths, and left dozens of FIXMEs to add error codes for those paths, is IMO the issue to address here if we're having trouble with diagnostics.

We can theoretically thread a Host, or just a reference to the Event log, to other places if it's important to capture higher-detail debug events. But I don't see why that's necessary here, and it is not currently the case that everywhere that we generate a `HostError` we strictly ensure that there's an Event log at hand. For example all the errors generated in MeteredMap just do `Status.into()` (eg. https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/metered_map.rs#L133) and that's .. currently at least considered adequate. We assume that if we are hitting errors there we'll add enough codes to disambiguate them.

> The SnapshotSource doesn't have a reference of a Host in this case. Do we need to change the SnapshotSource trait to pass the Host down? Or another idea?

I'm suggesting that, depending on the questions above, it might not need one. I think it probably doesn't, as the callback interface to SnapshotSource is a fairly simple get/put interface. Errors originating _in_ it are probably able to self-describe with codes, and since SnapshotSource is called _by_ the Host, any error it returns with insufficient _context_ could be augmented as it passes through the host (by attaching the debug log). It wouldn't let SnapshotSource _add_ to the debug log, but .. again I'm really not clear on why it has to here.

leighmcculloch:
> I guess I wonder why you're after "text from the error", exactly.

Because some of the users of the host aren't constrained by the Stellar protocol. We can add codes for the cases we need to the protocol, but also it's a bit odd that codes that would never be generated on network would show up in the protocol. The code where we're using it is in the soroban-rpc. The soroban-rpc might change in ways that would require expanding the error code set in ways that would be unacceptable. For example, if the snapshot source in soroban-rpc was to encounter HTTP errors, would we add HTTP errors to the XDR?

jayz22:
> > I guess I wonder why you're after "text from the error", exactly.
> 
> Because some of the users of the host aren't constrained by the Stellar protocol. We can add codes for the cases we need to the protocol, but also it's a bit odd that codes that would never be generated on network would show up in the protocol. The code where we're using it is in the soroban-rpc. The soroban-rpc might change in ways that would require expanding the error code set in ways that would be unacceptable. For example, if the snapshot source in soroban-rpc was to encounter HTTP errors, would we add HTTP errors to the XDR?

@leighmcculloch's point above was why I originally felt the need to make this change. 
@graydon I want to point out the middle ground solution -- what's in this changeset minus the last two commits (before [2cd4814](https://github.com/stellar/rs-soroban-env/pull/644/commits/2cd48145952a53a053a071ebf42c1e1741e7b643)), which changes the Snapshot's error but not Storage and Footprint's. The snapshot originated errors are converted to a known ScStatus. It doesn't introduce nearly as much churn but also doesn't preserve the original error. (See conversation [above](https://github.com/stellar/rs-soroban-env/pull/644#discussion_r1091114033))

That said, with current ongoing effort of restructuring of the `ScVal` and the choice of exposing some but not all `ScVal` types, would it be better if we not expose all status codes or leave room for new ones to be added more conveniently?
Also, my understanding is we don't surface the status codes at the transaction level (they are not hashed), so they are technically not part of the protocol? cc @sisuresh

graydon:
> > I guess I wonder why you're after "text from the error", exactly.
> 
> Because some of the users of the host aren't constrained by the Stellar protocol.

I think this is a stretch. We stopped trying to build an "XDR definitions optional, Stellar-agnostic" version of Soroban a while back AFAICT, relying on the fact that it's still reasonably easy to fork-and-delete stuff if you want, but not explicitly making a config that you can build without (say) Stellar AccountIDs or the SAC or such.

As far as host-embedders (eg the RPC server) who just happen to be developed after-the-fact / downstream of setting down error codes in the XDR, we have basically unlimited error-code space and have already marked a part of it for not-defined-ahead-of-time error codes written by contract users. We could easily mark off another space for errors-from-callback-interfaces in the host.

Currently (intentionally) the set of such callback interfaces is _very_ small -- only one, the SnapshotSource provider -- and my intent is to keep it that way. So I think it's probably overengineering to expect a really wide variety of errors to need to be expressed here and expect we could just wire them all in to the XDR up front. But I can imagine our reorg of errors might wind up a little flexible on this front (we've discussed moving to less-specific broad categories for example).

> We can add codes for the cases we need to the protocol, but also it's a bit odd that codes that would never be generated on network would show up in the protocol.

What I'm saying here is that, as far as I can tell, the case you linked to only has meaningful errors to report that are of the same basic variety as error conditions that do show up in the protocol. I'm still not sure which of the ~8 lines of code linked to is generating a complex and inscrutable error. They look to me like "base64 decode failed", which is both fairly ubiquitous and fairly limited in content. It's not like we're going to print the failed-to-decode base64 string to the error log anyway.

> The code where we're using it is in the soroban-rpc. The soroban-rpc might change in ways that would require expanding the error code set in ways that would be unacceptable. For example, if the snapshot source in soroban-rpc was to encounter HTTP errors, would we add HTTP errors to the XDR?

No, but why does the end user benefit from in-band reporting of something at that level of detail? They're not running an RPC server. "Preflight failed due to IO error in ledger snapshot" is already far more detail than they will be able to act on, and the RPC-server operator has access to the RPC server logs themselves (which can write whatever they want).

Again, if this _is_ really important to signal at a high level of detail I'm not totally opposed to threading the Events object down to the Storage interface to let the embedder deposit a higher-fidelity error there, and I think that's a simpler fix than this PR; but I am also not even sure that's needed yet.

leighmcculloch:
> I think it's probably overengineering to expect a really wide variety of errors to need to be expressed here and expect we could just wire them all in to the XDR up front.

That's a really good point.

Also in the case we have we can probably get away with adding our own hooks into this specific implementation of the snapshot source, if we need it, where those hooks would log or report the finer error details, and that would create no burden on the host or protocol supporting custom errors.

PR #646, opened 2023-01-31
Do not include `Status` as the first arg of a `DebugEvent`

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/622.
Adds a doctest to demonstrate how to use it correctly.

### Why

The `DebugError` contains a `Status` and a `DebugEvent`, 
The `Status` is always present and belongs to the error reporting workflow.
The `DebugEvent` takes additional `RawVal`s as args for debug info. The `Status` should not be automatically included as the first arg, since it is 1. already included in the error flow and reported separately 2. easy to be used incorrectly, since it's not obvious to the user they need to pass in a fmt str with one additional slot for the status (this leads to bug https://github.com/stellar/rs-soroban-env/issues/622).

### Known limitations

[TODO or N/A]


sisuresh (approved):

graydon (approved):
Another case of me trying to be too clever, clearly!

Issue #604, opened 2022-12-08
Possibly split BytesClone into two cost types


graydon:
In https://github.com/stellar/rs-soroban-env/pull/585 there's a TODO in metered_clone.rs about possibly splitting `BytesClone` into two separate costs: one for allocating heap space for a new block-of-bytes, and a separate cost for doing a shallow-copy of data into the new block-of-bytes.

This TODO derives from two facts:

1. The change splits (in several places, eg. also in metered_vector) the costs of allocation from the costs of cloning. This is because we wish to charge for cloning _in bulk_ when doing shallow clones, rather than charging one-by-one. In the best case, both charging _and_ cloning will happen in bulk (i.e. cloning will just be a single call to memcpy). 
2. the fact that we're overloading the use of `BytesClone` here in the first place -- it was originally designed just to track the cost of duplicating a `Bytes` host object -- and maybe a better thing to do here is to make a separate, dedicated cost called `ShallowClone` that represents the cost of cloning "something shallow" without the cost of allocation.

Issue #680, opened 2023-02-09
Possibly differentiate various types of shallow clones


jayz22:
Currently in `metered_clone.rs` the `MeteredClone` has field `IS_SHALLOW` which indicate if the type contains growable substructures (if not, then it is "shallow"). 
However, this definition is too board and inaccurately underestimates the effort of doing a shallow copy
- A shallow copy is defined as a "ByteClone", which is calibrated with the work of copying a single byte. Most types' clone does more work than cloning a byte (e.g cloning a sha256 hash, or an Rc bump). 
- Even shallow types may contain substructures, such as `ScAddress` whose definition may expand in the future. 

One possible solution is to have our own version of "size_of" (similar to the [`mem::size_of`](https://doc.rust-lang.org/std/mem/fn.size_of.html) which is unstable) that contains all of our host types (which is a limited set). We can use that as the input to the `BytesClone` charge, so that all the shallow clones can still share the same cost type.

jayz22:
@graydon We had a brief discussion on this in https://github.com/stellar/rs-soroban-env/pull/585#discussion_r1041698235. 
I'd like to revisit the topic of charging clone by the size of the type, possibly using `https://doc.rust-lang.org/std/mem/fn.size_of.html`. From what I'm reading, this is a stable Rust function, and should work deterministically for primitive types. So maybe we can use this for primitive types and augment it for `struct`, `enum` and some of the imported xdr types.

jayz22:
Closed by https://github.com/stellar/rs-soroban-env/pull/690

PR #690, opened 2023-02-17
Reform metering for value cloning and memory allocation

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/604 and https://github.com/stellar/rs-soroban-env/issues/680.

This PR reforms `metered_clone`.
- Splits memory allocation from shallow copying
- Define the "shallow size" of `MeteredClone` types. 

A `metered_clone` consist of shallow copying of the type's memory footprint (for a `Vec` it's 24), plus recursively (deep) cloning any of the substructures. 
A `HostMemAlloc` charge occurs only when memory is allocated. I.e. types involving `alloc` -- `Box`, `Vec`, `Rc` etc.
The shallow copying of a type is charged the amount `ELT_SIZE` defined by the implementer, and there is a runtime check to ensure that amount is the upper bound of the actual memory size of the type `mem::size_of<T>`.
Adds a test in `budget_metering.rs` to demonstrate the metered cloning of a deep structure. 

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/host/metered_clone.rs
+++ b/soroban-env-host/src/host/metered_clone.rs
@@ -1,31 +1,45 @@
-use std::rc::Rc;
+use std::{mem, rc::Rc};
 
 use soroban_env_common::{
-    xdr::{BytesM, LedgerEntry, LedgerKey, ScAddress, ScMap, ScObject, ScVal},
+    xdr::{
+        BytesM, ContractEvent, ContractEventBody, LedgerEntry, LedgerKey, ScAddress, ScMap,
+        ScMapEntry, ScObject, ScVal,
+    },
     RawVal,
 };
 
 use crate::{
     budget::{Budget, CostType},
+    events::{DebugArg, DebugEvent, HostEvent, InternalContractEvent, InternalEvent},
     host::Events,
     storage::AccessType,
     xdr::{AccountId, Hash, ScContractCode, ScVec, Uint256},
     HostError,
 };
 
-// Charge for an N-element "shallow copy" of some type, not cloning any
-// substructure. In a better world we would multiply this by size_of<Self> but
-// that's not guaranteed to be stable. Also this has a fairly serious level of
-// inaccuracy since we call the same charge for any shallow type, whether it's a
-// simple byte-copy or a sequence of Rc<>.clone calls (each of which bumps a
-// refcount -- shallow but not as cheap as memcpy).
-//
-// TODO: this is correct for CPU-cost but not memory-cost, since there's no heap
-// allocation implied by a shallow clone. We should (maybe) split BytesClone
-// into ByteCopy (the CPU byte-copying part) and HeapAlloc (the heap-allocating
-// part). See https://github.com/stellar/rs-soroban-env/issues/604.
-fn charge_shallow_copy(n_elts: usize, budget: &Budget) -> Result<(), HostError> {
-    budget.charge(CostType::BytesClone, n_elts as u64)
+extern crate static_assertions as sa;
+
+// Charge for an N-element "shallow copy" of some type, not cloning any substructure. The charge
+// unit is number of elements `n_elts` multiply by size of each element. In a better world we would
+// multiply by size_of<Self> instead but that's not guaranteed to be stable, which might cause
+// metering to differ across compilations, causing serious problems in concensus and replay.
+fn charge_shallow_copy<T: MeteredClone>(n_elts: u64, budget: &Budget) -> Result<(), HostError> {
+    // Ideally we would want a static assertion. However, it does not work due to rust restrictions
+    // (e.g. see https://github.com/rust-lang/rust/issues/57775). Here we make a runtime assertion
+    // that the type's size is below its promised element size for budget charging. This assertion
+    // only happens in debug build. In optimized build, not satisfying the asserted condition just
+    // means an underestimation of the cost.
+    debug_assert!(mem::size_of::<T>() as u64 <= T::ELT_SIZE);
+    budget.charge(CostType::HostMemCpy, n_elts * T::ELT_SIZE)
+}
+
+// Let it be a free function instead of a trait because charge heap alloc maybe called elsewhere,
+// not just metered clone, e.g. Box::<T>::new().
+fn charge_heap_alloc<T: MeteredClone>(n_elts: u64, budget: &Budget) -> Result<(), HostError> {
+    // Here we make a runtime assertion that the type's size is below its promised element size for
+    // budget charging.
+    debug_assert!(mem::size_of::<T>() as u64 <= T::ELT_SIZE);
+    budget.charge(CostType::HostMemAlloc, n_elts * T::ELT_SIZE)
 }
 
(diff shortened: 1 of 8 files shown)


soroban-env-host/src/test/budget_metering.rs
```diff
@@ -142,3 +144,51 @@ fn map_insert_key_vec_obj() -> Result<(), HostError> {
...
+    ];
+
+    v.charge_for_clone(host.as_budget())?;
+    // Transition #                  (1)           (2)             (3)
+    // Type(size, count):  Vec(24,1) ---> Box(8,3) ----> Vec(24,3) ------> ScMapEntry(80,3x2)
+    // MemAlloc:                     8x3           24x3            80x3x2                     = 576
+    // MemCpy:             24        8x3  8x3      24x3  24x3      80x3x2  80x3x2             = 1176
+    // At the alloc boundary there is some double-charging happening.
+    // For example in transition #1, `Vec` charges for (1) allocation of 8x3 bytes, (2) copying of
+    // 8x3 bytes data into it. The `Box` charges for 8x3 for its own size.
```
jayz22, line 180: There is a bit of double counting going on. It comes from my current design and I’m not 100% sure if we can totally eliminate it: each type takes care itself (shallow copying of its own “pointer” size) and its substructures (allocating buffer, and shallow copying member “pointers” into the buffer). There isn’t really a way to know if itself is a substructure of something else, in which case its parent may have taken care of the shallow copying its pointer part. Thus leads to the shallow copy being double counted.

soroban-env-host/src/host/metered_clone.rs
```diff
@@ -121,82 +185,256 @@ impl MeteredClone for ScVal {
...
+        }
+    }
+}
+
+mod test {
+    #[allow(unused)]
+    use super::*;
+
+    #[test]
+    fn test_elt_size() {
```
jayz22, line 357: I'm not sure if there's a way to automatically cover this test on all types implementing the `MeteredClone` trait.

soroban-env-host/src/host/metered_clone.rs
```diff
@@ -1,31 +1,45 @@
...
-// TODO: this is correct for CPU-cost but not memory-cost, since there's no heap
-// allocation implied by a shallow clone. We should (maybe) split BytesClone
-// into ByteCopy (the CPU byte-copying part) and HeapAlloc (the heap-allocating
-// part). See https://github.com/stellar/rs-soroban-env/issues/604.
-fn charge_shallow_copy(n_elts: usize, budget: &Budget) -> Result<(), HostError> {
-    budget.charge(CostType::BytesClone, n_elts as u64)
+extern crate static_assertions as sa;
+
+// Charge for an N-element "shallow copy" of some type, not cloning any substructure. The charge
+// unit is number of elements `n_elts` multiply by size of each element. In a better world we would
```
graydon, line 23:
```suggestion
// unit is number of elements `n_elts` multiplied by a declared size of each element. In a better world we would
```

soroban-env-host/src/host/metered_clone.rs
```diff
@@ -121,82 +185,256 @@ impl MeteredClone for ScVal {
...
-            // Shallow types we can batch-up our charging for.
-            budget.charge(CostType::BytesClone, self.len() as u64)?;
-            Ok(self.clone())
-        } else {
-            self.iter().map(|elt| elt.metered_clone(budget)).collect()
+
+    const ELT_SIZE: u64 = 24;
+
+    fn charge_for_clone(&self, budget: &Budget) -> Result<(), HostError> {
+        // first take care of Vec clone, which involves allocating memory, (shallowly) memcpy the
```
graydon, line 235: Hm, this part I think isn't right -- it removed the special case of charging once for a clone of vectors of shallow elements.

graydon: Except, wait, no, you _are_ charging for shallow copies of C, but then also charging for non-shallow cloning. I think I'm confused about what's going on here.

soroban-env-host/src/host/metered_clone.rs
```diff
@@ -121,82 +185,256 @@ impl MeteredClone for ScVal {
...
     const IS_SHALLOW: bool = false;
 
-    fn metered_clone(&self, budget: &Budget) -> Result<Self, HostError> {
+    const ELT_SIZE: u64 = 8;
+
+    fn charge_for_clone(&self, budget: &Budget) -> Result<(), HostError> {
+        // first take care of the Box part
+        charge_shallow_copy::<Box<C>>(1, budget)?;
+        charge_heap_alloc::<C>(1, budget)?;
+        charge_shallow_copy::<C>(1, budget)?;
```
graydon, line 257: Similarly this looks like it's charging for copying C .. twice? once shallow and once deep.

soroban-env-host/src/host/metered_clone.rs
```diff
@@ -121,82 +185,256 @@ impl MeteredClone for ScVal {
...
+
+    const ELT_SIZE: u64 = 80;
+
+    fn charge_for_clone(&self, budget: &Budget) -> Result<(), HostError> {
+        charge_shallow_copy::<Self>(1, budget)?;
+        if self.args.is_heap() {
+            // equivalent to charging for a `Vec` clone.
+            charge_shallow_copy::<Vec<DebugArg>>(1, budget)?;
+            charge_heap_alloc::<Vec<DebugArg>>(self.args.len() as u64, budget)?;
+            charge_shallow_copy::<Vec<DebugArg>>(self.args.len() as u64, budget)
```
graydon, line 293: another case of the shallow-type vector double charging situation here

soroban-env-host/src/host/metered_clone.rs
```diff
@@ -35,14 +49,20 @@ pub trait MeteredClone: Clone {
     // `charge_for_clone` (correctly charging for cloning the substructure).
     const IS_SHALLOW: bool = true;
 
+    // Size (num of bytes) of a single element. This value determines the input for budget charging.
+    // It should be the upperbound (across various compilations and platforms) of the actual type's
+    // size. Implementer of the trait needs to decide this value based on Rust's guideline on type
+    // layout: https://doc.rust-lang.org/reference/type-layout.html
```
graydon, line 55:
```suggestion
    // layout: https://doc.rust-lang.org/reference/type-layout.html
    //
    // Note: When `Self::IS_SHALLOW = false`, `ELT_SIZE` should only cover any _extra_ size of `Self` above
    // and beyond the size of its substructure. For example for `Vec<C>` we have `ELT_SIZE=24` to account
    // for the size of the vector header / fat-pointer part, _not_ the size of the elements. For a struct type
    // that is not shallow, typically `Self::ELT_SIZE` will be zero, or nearly zero (possibly covering any padding
    // that occurs between elements).
```

jayz22:
@graydon I have reformed it once again after our discussion. All the previous double-charging of shallow copies should be eliminated. I believe this is ready for merge. Please take another look.

graydon (approved):

PR #698, opened 2023-02-21
Remove the unnecessary error events that appeared in non-error scenarios

dmkozh:
### What

- The account now returns an option and it's up to the client to return an error.
- Use `ok_or_else` instead of `ok_or`

### Why

Remove confusing error messages that don't mean anything.

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/auth.rs
+++ b/soroban-env-host/src/auth.rs
@@ -318,7 +318,10 @@ impl AuthorizationManager {
                             // so that it's possible to create the auth manager
                             // without knowing the invoker beforehand and also
                             // to not keep calling into `source_account` function.
-                            let source_addr = ScAddress::Account(host.source_account()?);
+                            let source_addr =
+                                ScAddress::Account(host.source_account().ok_or_else(|| {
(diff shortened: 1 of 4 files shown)


willemneal:
Is clippy part of the CI for this repo?  This seems like something it would have caught.

dmkozh:
> Is clippy part of the CI for this repo? This seems like something it would have caught.

I highly doubt this would be possible to find with clippy. There is nothing particularly wrong with the old code and `host.err_general` producing events is a design quirk that we can't do much about.

sisuresh (approved):

willemneal:
@dmkozh Here is a similar issue for options found by clippy: https://rust-lang.github.io/rust-clippy/rust-1.66.0/index.html#or_fun_call

So it looks like we should submit a new lint!

dmkozh:
> @dmkozh Here is a similar issue for options found by clippy: https://rust-lang.github.io/rust-clippy/rust-1.66.0/index.html#or_fun_call
> 
> So it looks like we should submit a new lint!

I see, that's neat!

PR #699, opened 2023-02-21
feat: copy `create_ledger_footprint` from CLI

willemneal:
This was in the utils.rs file in the cli and should be here.

### DIFF
--- a/soroban-env-host/src/storage.rs
+++ b/soroban-env-host/src/storage.rs
@@ -9,6 +9,7 @@
 
 use std::rc::Rc;
 
+use soroban_env_common::xdr::{self, LedgerFootprint};
 use soroban_env_common::Compare;
 
 use crate::budget::Budget;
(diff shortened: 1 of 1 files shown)


soroban-env-host/src/storage.rs
```diff
@@ -116,6 +117,23 @@ impl Footprint {
             Err(ScHostStorageErrorCode::AccessToUnknownEntry.into())
         }
     }
+
+    pub fn create_ledger_footprint(&self) -> Result<LedgerFootprint, xdr::Error> {
```
dmkozh, line 121: Would `to_xdr` name make more sense?

dmkozh (approved):

soroban-env-host/src/storage.rs
willemneal: Makes sense to me.

Issue #605, opened 2022-12-09
Charge comparison of `FixedSizeOrdType` an amount corresponding to its type.


jayz22:
In https://github.com/stellar/rs-soroban-env/blob/a7843346bf2db743291aca6575ec82f6a5e7f1a3/soroban-env-host/src/host/comparison.rs#LL100C31-L100C31 
We are charging 1 for all types. Ideally this should be `mem::size_of<T>` but this interface is [not stable](https://doc.rust-lang.org/std/mem/fn.size_of.html#:~:text=In%20general%2C%20the%20size%20of%20a%20type%20is%20not%20stable%20across%20compilations%2C%20but%20specific%20types%20such%20as%20primitives%20are.) and we can't use it. 
Instead, we should pass the size of the structure, if it's known for the type, or an upper bound size into the macro, and charge the corresponding amount to the budget.

PR #714, opened 2023-02-28
Split out `DeclaredSizeForMetering` from `MeteredClone` and use it for `Compare`

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/605. 
It splits out the `DeclaredSizeForMetering` from `MeteredClone` (previously defined as `MeteredClone::ELT_SIZE`) out of metered clone and share it between clone and comparison. 
Adapt to the object overahaul, which introduced many new types. Declared all the RawVal-wrapping types to be fixed size. 
Fix `try_compare_to_small` to use metered compare. 

### Why

### Known limitations

[TODO or N/A]


soroban-env-host/src/host/declared_size.rs
```diff
@@ -0,0 +1,438 @@
...
+}
+
+impl<C> DeclaredSizeForMetering for Box<C> {
+    const DECLARED_SIZE: u64 = 8;
+}
+
+impl<C: DeclaredSizeForMetering> DeclaredSizeForMetering for Option<C> {
+    // Size of C plus an 8 byte alignment overhead. If we need to handle types with larger
+    // alignment size, we need to increase the overhead.
+    const DECLARED_SIZE: u64 = C::DECLARED_SIZE + 8;
```
graydon, line 186: .saturating_add(8)

soroban-env-host/src/host/metered_map.rs
```diff
@@ -41,7 +41,7 @@ where
     Ctx: AsBudget + Compare<K, Error = HostError> + Compare<V, Error = HostError>,
 {
     fn charge_new<B: AsBudget>(n_elts: u64, b: &B) -> Result<(), HostError> {
-        let n_bytes = <Vec<(K, V)> as MeteredClone>::ELT_SIZE * n_elts;
+        let n_bytes = <(K, V) as DeclaredSizeForMetering>::DECLARED_SIZE * n_elts;
```
graydon, line 44: .saturating_mul(n_elts)

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -22,7 +22,10 @@ where
 {
     /// Covers the cost of allocating a new Vec<A> with size in number of elements.
     fn charge_new(n_elts: u64, budget: &Budget) -> Result<(), HostError> {
-        budget.charge(CostType::VecNew, n_elts * A::ELT_SIZE)
+        budget.charge(
+            CostType::VecNew,
+            n_elts * <A as DeclaredSizeForMetering>::DECLARED_SIZE,
```
graydon, line 27: .saturating_mul(...)

soroban-env-host/src/host/declared_size.rs
```diff
@@ -0,0 +1,438 @@
...
+    const DECLARED_SIZE: u64 = 16;
+}
+
+impl<K, V> DeclaredSizeForMetering for (K, V)
+where
+    K: DeclaredSizeForMetering,
+    V: DeclaredSizeForMetering,
+{
+    const DECLARED_SIZE: u64 = <K as DeclaredSizeForMetering>::DECLARED_SIZE
+        + <V as DeclaredSizeForMetering>::DECLARED_SIZE;
```
graydon, line 160: .saturating_add(...)

soroban-env-host/src/host/declared_size.rs
```diff
@@ -0,0 +1,438 @@
...
+        expect!["24"].assert_eq(size_of::<ScSymbol>().to_string().as_str());
+        // composite types
+        expect!["16"].assert_eq(size_of::<&[ScVal]>().to_string().as_str());
+        expect!["48"].assert_eq(size_of::<(RawVal, ScVal)>().to_string().as_str());
+        expect!["200"].assert_eq(size_of::<[ScVal; 5]>().to_string().as_str());
+        expect!["24"].assert_eq(size_of::<BytesM<10000>>().to_string().as_str());
+        expect!["24"].assert_eq(size_of::<StringM<10000>>().to_string().as_str());
+        expect!["24"].assert_eq(size_of::<Vec<ScVal>>().to_string().as_str());
+        expect!["8"].assert_eq(size_of::<Box<ScVal>>().to_string().as_str());
+        // For these below we don't reply on the size_of, since it is the lower bound
```
graydon, line 309: I think you can still assert_le!(usize::from_str(expect["8"]).unwrap(), size_of::<...>())

graydon (changes requested):
Looks pretty good! handful of nits

soroban-env-host/src/host/declared_size.rs
jayz22: Interesting. This works:
```
assert_eq!(
    usize::from_str_radix(expect!["1"].data(), 10).unwrap(),
    size_of::<bool>()
);
```
I need `assert_eq` here because here I'm trying to get the exact number from `size_of` for visualization. The section below (the various `assert_mem_size_equals_declared_size` calls) are the actual assertions. There we might consider switching `eq` with `le` but I think having a more strict bound is helpful unless it turns out to be too much maintenance burden. 
Also `assert_le!` is not an std macro, it is available from a 3rd party lib e.g. https://docs.rs/more-asserts/latest/more_asserts/macro.assert_le.html

Overall the above expression is more verbose than the original so I'm leaning towards keeping the original. What do you think?

soroban-env-host/src/host/declared_size.rs
jayz22: Fixed.

soroban-env-host/src/host/metered_map.rs
jayz22: Fixed.

soroban-env-host/src/host/metered_vector.rs
jayz22: Fixed.

soroban-env-host/src/host/declared_size.rs
jayz22: Fixed.

soroban-env-host/src/host/declared_size.rs
graydon: It's fine, I'm not saying the order of the instruction is wrong, I was just suggesting that you could make an assertion of a bound rather than equality in the cases where you are avoiding measuring size altogether (you have a comment there saying you don't rely on size measurement in some cases)

graydon (approved):

PR #715, opened 2023-03-02
Implement the Arbitrary fuzzing trait for contract value-types

brson:
### What

Implement the Arbitrary trait for Symbol, BitSet, Static, and Status.

### Why

This is a prerequisite for more complete fuzzing support in soroban-sdk: https://github.com/stellar/rs-soroban-sdk/pull/878
This patch is only needed if that patch is to be accepted.

This patch also tweaks and reexports the `call_with_suppressed_panic_hook` function,
which is used by the sdk patch to provide users a helper function for catching panics in fuzz tests.

A more complete description is in the above link.

### Known limitations

See sdk PR.


brson:
Rebased.

dmkozh:
Just a heads up: we are in the middle of a significant env/sdk rewrite that is definitely going to affect this and SDK PRs to some degree. Due to that I would be inclined to wait until the rewrite is landed (should happen in a few days). Thanks for the work and sorry for slow review.

brson:
Rebased.

soroban-env-common/src/arbitrary.rs
```diff
@@ -0,0 +1,48 @@
...
+extern crate alloc;
+
+use crate::xdr::ScStatus;
+use crate::Status;
+use crate::SymbolSmall;
+use alloc::string::String;
+use alloc::vec::Vec as RustVec;
+use arbitrary::{Arbitrary, Unstructured};
+
+impl<'a> Arbitrary<'a> for SymbolSmall {
```
dmkozh, line 14: What is the use case for this now? FWIW the SDK doesn't expose `SymbolSmall` explicitly and instead provides a wrapper `Symbol` type that can be either small or env-based value. So maybe it would make more sense to implement this together with SDK fuzzing support (I suppose it would be similar to fuzzing any env type, like bytes or string).

soroban-env-common/src/arbitrary.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#![cfg(feature = "testutils")]
+
+extern crate alloc;
+
+use crate::xdr::ScStatus;
+use crate::Status;
+use crate::SymbolSmall;
+use alloc::string::String;
+use alloc::vec::Vec as RustVec;
```
dmkozh, line 11: There is no name collision here with some other `Vec`, is there? Up to you though.

soroban-env-common/src/arbitrary.rs
brson: If the SDK doesn't expose this, then it probably isn't needed. I just rebased this week and adapted this code from the prior code for Symbol. I need to go back and reevaluate which types need arbitrary impls.

brson:
Is the `Status` types still exposed by the sdk? If not this patch is no longer needed at all.

soroban-env-common/src/arbitrary.rs
brson: I've removed it.

brson:
I removed the implementation of Arbitrary for SymbolSmall. The only Arbitrary implementation left now in this crate is for Status. I'm not clear on whether contract authors can still access this type or not. It may be that none of the non-object contract types need to implement Arbitrary.

dmkozh:
> Is the Status types still exposed by the sdk?

Yeah, it's still exposed, so probably we still need this.

> I removed the implementation of Arbitrary for SymbolSmall. The only Arbitrary implementation left now in this crate is for Status. I'm not clear on whether contract authors can still access this type or not. It may be that none of the non-object contract types need to implement Arbitrary.

`SymbolSmall` is accessible via `soroban-env-common`, but it isn't too useful as SDK exposes a higher level `Symbol` type that can be either small or object-based symbol, so I think it would make more sense to implement Arbitrary for the SDK `Symbol` instead.

dmkozh (approved):

dmkozh:
Let's wait a bit with merging this to get it into the next release (the current release happens right now and there is also the SDK change which we won't have time to merge into it)

PR #725, opened 2023-03-13
Move `InvokerType` from common env to internal host implementation.

dmkozh:
### What

Move `InvokerType` from common env to internal host implementation. Also removed the respective host function that gets the invoker type.

### Why

We no longer need to expose this externally thanks to auth next. The questions about the call stack can be better/more generically via the call stack getter host fn.

### Known limitations

N/A


graydon (approved):
LGTM! thanks

PR #728, opened 2023-03-15
chore: clippy fixes

willemneal:
### What

Add a clippy check to CI and make initial pass at fixing clippy errors.

At first the number of lints was way too high to tackle all at once.  However, disabling some that could be debatable allowed a lot of the obvious fixes to be done. 

In future PRs the disabled lints can be addressed in batches to ensure so that it's not too overwhelming.

### Why

Currently the only repo that requires that clippy is the CLI and this repo seems more important to have standard coding practices. Clippy can even help catch some logic errors.

### Known limitations

[TODO or N/A]


willemneal:
One other thing to note.  Currently it should fail when run for the first time.  The reason is that the latest commit to main added two methods which return a result, but will never return an error. I kept it in partly as a check that CI should fail the first time and unlike the initial batch of changes it is an API change to remove the Result.

willemneal:
The CI failed as expected an I disabled `unnecessary_wraps` for now so it should pass now.

soroban-test-wasms/src/lib.rs
```diff
@@ -42,25 +42,22 @@
 //!      documentation purpopses without having to worry about breaking tests in
 //!      the host here.
 
-pub const ADD_I32: &'static [u8] =
-    include_bytes!("../wasm-workspace/opt/example_add_i32.wasm").as_slice();
-pub const CREATE_CONTRACT: &'static [u8] =
+pub const ADD_I32: &[u8] = include_bytes!("../wasm-workspace/opt/example_add_i32.wasm").as_slice();
```
dmkozh, line 45: Interesting, I don't quite understand why static is gone here

soroban-synth-wasm/src/mod_emitter.rs
```diff
@@ -56,6 +56,12 @@ pub struct ModEmitter {
     import_refs: HashMap<(String, String, Arity), FuncRef>,
 }
 
+impl Default for ModEmitter {
```
dmkozh, line 59: That's weird, why is this needed?

dmkozh (commented):
Thanks for this cleanup. This looks good to me for the most part, besides a couple questionable diagnostics/changes I've pointed in comments.

soroban-test-wasms/src/lib.rs
willemneal: It's implied since the const lives for the whole program.

soroban-synth-wasm/src/mod_emitter.rs
willemneal: It's a suggestion. Since `new` doesn't take any arguments it's akin to `default`. Can suppress this one if it is unwanted.

dmkozh: Sure, I realize it's a suggestion. I'm just trying to understand which diagnostics don't seem to be useful. IMHO this one is not super useful, especially given that host implementation is pretty self-contained. It feels a bit weird to introduce custom trait implementations that aren't actually being used anywhere.

soroban-env-host/src/host.rs
```diff
@@ -299,7 +294,10 @@ impl Host {
     {
         match self.0.ledger.borrow_mut().as_mut() {
             None => Err(self.err_general("missing ledger info")),
-            Some(li) => Ok(f(li)),
+            Some(li) => {
```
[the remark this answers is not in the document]
willemneal: https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg

soroban-synth-wasm/src/mod_emitter.rs
willemneal: Gone!

soroban-env-common/src/env_val.rs
```diff
@@ -143,7 +143,7 @@ impl<E: Env> TryFromVal<E, RawVal> for i128 {
             let obj = v.try_into()?;
             let lo = env.obj_to_i128_lo64(obj).map_err(|_| ConversionError)?;
             let hi = env.obj_to_i128_hi64(obj).map_err(|_| ConversionError)?;
-            let u: u128 = (lo as u128) | ((hi as u128) << 64);
+            let u: u128 = u128::from(lo) | (u128::from(hi) << 64);
```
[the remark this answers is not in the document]
willemneal: Currently it's defined by this macro, which is a const and `inline(always)` so it should be the same. However, if you think it's worth removing I can disable the lint  for this function or file.

```rs
// Conversion traits for primitive integer and float types
// Conversions T -> T are covered by a blanket impl and therefore excluded
// Some conversions from and to usize/isize are not implemented due to portability concerns
macro_rules! impl_from {
    ($Small: ty, $Large: ty, #[$attr:meta], $doc: expr) => {
        #[$attr]
        #[rustc_const_unstable(feature = "const_num_from_num", issue = "87852")]
        impl const From<$Small> for $Large {
            // Rustdocs on the impl block show a "[+] show undocumented items" toggle.
            // Rustdocs on functions do not.
            #[doc = $doc]
            #[inline(always)]
            fn from(small: $Small) -> Self {
                small as Self
            }
        }
    };
    ($Small: ty, $Large: ty, #[$attr:meta]) => {
        impl_from!($Small,
                   $Large,
                   #[$attr],
                   concat!("Converts `",
                           stringify!($Small),
                           "` to `",
                           stringify!($Large),
                           "` losslessly."));
    }
}
```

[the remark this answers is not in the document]
dmkozh: Thanks a lot for digging into this! It looks like it is the same, so it's probably ok to leave the diagnostic. If we ever feel like this is an issue, we can revert this particular diagnostic.

soroban-env-host/src/host.rs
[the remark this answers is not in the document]
dmkozh: Hmm, the diagnostic looks useful in general, even though it's awkward in this particular case. So let's leave it be - it's probably worth one a bit awkward implementation.

dmkozh:
CI fails with `error: rustup is not installed at '/home/runner/work/rs-soroban-env/rs-soroban-env/.cargo/ci/config.toml'`

I'm not sure I understand how to fix it, but I suppose the reason has something to do with the newly introduced `config.toml`

willemneal:
@dmkozh fixed it. Was an ENV VAR for cargo home, which was from a previous attempt to have the clippy lints in a file.

dmkozh:
Seems like there are some failures in guest library.

willemneal:
@dmkozh I also seem to get the following error if I ran that command locally:

```
cargo clippy --locked --target wasm32-unknown-unknown --no-default-features
```
I get: 
```
error[E0152]: found duplicate lang item `panic_impl`
  --> soroban-env-common/tests/no_std/src/lib.rs:17:1
   |
17 | fn handle_panic(_: &core::panic::PanicInfo) -> ! {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: the lang item is first defined in crate `std` (which `soroban_env_common` depends on)
   = note: first definition in `std` loaded from /Users/willem/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-a9c931054ca2959f.rlib
   = note: second definition in the local crate (`test_no_std`)

For more information about this error, try `rustc --explain E0152`.
```

dmkozh:
> @dmkozh I also seem to get the following error if I ran that command locally:
> 
> ```
> cargo clippy --locked --target wasm32-unknown-unknown --no-default-features
> ```
> 
> I get:
> 
> ```
> error[E0152]: found duplicate lang item `panic_impl`
>   --> soroban-env-common/tests/no_std/src/lib.rs:17:1
>    |
> 17 | fn handle_panic(_: &core::panic::PanicInfo) -> ! {
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    |
>    = note: the lang item is first defined in crate `std` (which `soroban_env_common` depends on)
>    = note: first definition in `std` loaded from /Users/willem/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-a9c931054ca2959f.rlib
>    = note: second definition in the local crate (`test_no_std`)
> 
> For more information about this error, try `rustc --explain E0152`.
> ```

That's likely because you're missing `--features testutils` (probably that should be fixed separately...)

graydon:
Hey @willemneal sorry for the delay (and inevitable bitrot).

I've merged a run of `cargo clippy --fix` in https://github.com/stellar/rs-soroban-env/pull/790 and turned on clippy in CI _without_ the `style` group in https://github.com/stellar/rs-soroban-env/pull/796. So that accomplishes maybe half or a third of what this PR was trying to do, but lays groundwork. Assuming you want to enable all of the extras (or even all the way to enabling the full `style` group) I'd recommend refreshing this PR in stages -- doing a handful of new lints at a time, since it's easier to review a small or medium-sized PR than a giant one.

Issue #681, opened 2023-02-09
Cache contract id as object -- possibly use ubiquitous interning


graydon:
We fetch contract IDs as hashes fairly often, which means we're either copying around 32-byte hashes in the host or converting them to 8-byte `Object`s that hold handles pointing to bytes in the host object array. That latter approach is ok as far as it goes, but we don't _unique_ those bytes, and we do a fair number of conversions, so it's possible that we're winding up with quite a lot of copies of a given contract ID in the host object array.

We've talked about doing ubiquitous "interning" (forcing host objects to be unique when added to the table) and this is still a bit of a design decision to work through, but it should at least be possible to _optionally_ ask for an object to be added uniquely, to cut down on cases like this.

graydon:
Decision: we're not going to do ubiquitous interning for now. It penalizes the common case (allocating actually-new objects) in favor of a rarer case (allocating duplicate objects). For objects we know we need to cache a widely-reused value of, we can cache that value.

graydon:
(we should still cache the contract ID, probably)

graydon:
Should probably also cache a variety of continuously-reused symbols and event-topic-vectors-containing-just-those-symbols.

anupsdf:
Discussed with @graydon and this can move to `post V1 launch` if we run out of time to get this done for v1.

Issue #653, opened 2023-02-02
Auth Next: Update the token interface for better Auth Next compatibility


dmkozh:
Auth Next allows us to simplify/extend the token interface. Here are some ideas:

- Remove `admin` address from the admin functions. The admin address is already retrieved from storage and we can just use it directly to `require_auth`
- Introduce a function (e.g. `xfer_max`) that requires user to sign the max amount of token to be spent, but also takes the actual transfer amount. This allows contracts to 'squash' `incr_allow` and `xfer_from` calls into one token invocation.
- Support 'ephemeral' storage for approvals that only exist either during the operation invocation. Evaluate whether we should support non-ephemeral/temporary approvals at all

PR #732, opened 2023-03-17
Introduce temporary storage in Soroban host.

dmkozh:
### What

Temporary storage is basically a contract-owned map that isn't persisted in the ledger, but may be used to share values between multiple contract calls within the same top-level invocation.

### Why

 This is useful to implement safe allowance-style patterns without the risk of the allowance to stay in the ledger indefinitely (e.g. to transfer token to a dynamic, arbitrary number of recipients).

### Known limitations

N/A


sisuresh (approved):
We should add some simple test cases, but that doesn't to block this PR.

soroban-env-host/src/storage.rs
```diff
@@ -294,6 +301,81 @@ impl Storage {
...
+
+/// A special-purpose map from arbitrary contract-owned values to arbitrary
+/// values.
+///
+/// Since `TempStorage` stores `RawVal`s, it has to be attributed to the host
+/// instance and can't be transferred between host instances.
+///
+/// Semantically, this is similar to the `Storage`, but it is never persisted
+/// and hence doesn't need to support the footprints and convert keys to the
+/// ledger-compatibile types.
```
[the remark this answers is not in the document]
dmkozh: Done.

soroban-env-host/src/storage.rs
```diff
@@ -7,17 +7,24 @@
...
 use crate::Host;
 use crate::{host::metered_map::MeteredOrdMap, HostError};
 
 pub type FootprintMap = MeteredOrdMap<Rc<LedgerKey>, AccessType, Budget>;
 pub type StorageMap = MeteredOrdMap<Rc<LedgerKey>, Option<Rc<LedgerEntry>>, Budget>;
+// Alias for the `RawVal` that has the cheap 'shallow' comparison defined - it
+// just compares the payloads without looking at the value semantics.
+// This is used in `TempStorageMap` to make sure that we don't do expensive
+// value comparisons.
+pub type ShallowComparableRawVal = RawVal;
```
graydon, line 26: This doesn't define a different type (for the sake of impls, for example). The `impl<ShallowComparableRawVal> for Budget` below is impl'ed on `RawVal` too, the two types are considered identical by this line.

You probably want `struct ShallowComparableRawVal(RawVal)` or .. something. But I am also not sure why you're doing this. Map-key comparison is actually supposed to be deep, otherwise you can put an entry under, say, `vec [1,2]` and later fail to retrieve it by looking it up with `vec[1,2]` with a superficially different object-handle identity.

soroban-env-host/src/storage.rs
```diff
@@ -294,6 +301,81 @@ impl Storage {
     }
 }
 
+impl Compare<ShallowComparableRawVal> for Budget {
```
graydon, line 304: I'm _pretty_ sure you don't want to do this, see above.

soroban-env-host/src/host.rs
```diff
@@ -2377,6 +2384,67 @@ impl VmCallerEnv for Host {
...
     }
 
+    // Notes on metering: covered by components
+    fn put_tmp_contract_data(
+        &self,
+        _vmcaller: &mut VmCaller<Host>,
+        k: RawVal,
+        v: RawVal,
+    ) -> Result<Void, HostError> {
+        let key = self.from_host_val(k)?;
```
graydon, line 2394: I .. don't think you want to do this conversion from k to host val / ScVal key. That's just needless overhead, you can use k as a key directly.

soroban-env-host/src/host.rs
```diff
@@ -2377,6 +2384,67 @@ impl VmCallerEnv for Host {
...
+        self.0.temp_storage.borrow_mut().put(
+            self.get_current_contract_id_internal()?.0,
```
graydon, line 2396: I think before this PR we should implement https://github.com/stellar/rs-soroban-env/issues/681 so that we have a cached `RawVal` on hand we can just reuse as the current contract ID.

graydon (changes requested):
I'm not 100% clear on the patterns that motivate this -- could you describe in a little more detail?

Regardless of that, there are also implementation issues I think should be addressed before merging. It seems to be both mis-keying data by comparing rawval payloads and also inefficiently converting keys to ScVals. And I _think_ it'd be good to do #681 before this (though if you're really pressed for time I guess we can come back and do that later, fixing this code to use it too)

soroban-env-host/src/host.rs
jayz22: +1 to @graydon. You shouldn't convert `RawVal` to `ScVal`. 
Also, you should just keep `Hash` instead of `[u8; 32]` for better clarity. `MeteredMap` can accept any key type that implements `Compare`. So `(Hash, RawVal)` will work. 
Also, you want to make the key type `Rc<(Hash, RawVal)>`, since `MeteredMap`'s metering is calibrated to small keys comparison. This is why the existing `Storage` has `Rc<LedgerKey>` as the key type.

soroban-env-host/src/storage.rs
dmkozh: This is not for keys, this is for values. I'm actually thinking that we probably need to come back to https://github.com/stellar/rs-soroban-env/pull/624 - I still believe that the value comparison requirement doesn't make sense for the map because the values should never be compared (unless we have a bug and allow non-unique entries). Instead it should do sorting only by keys.

soroban-env-host/src/host.rs
dmkozh: Agree on the rc part, I'll do the update. I also can use hash, I suppose it won't make much difference.

However, I'm not sure about using  `RawVal` for keys here. The semantics of the temp storage are the same as for the regular storage and regular storage uses ledger keys. Are we confident that `RawVal` comparisons are always equivalent to the corresponding `ScVal` comparisons, including UDTs etc?

soroban-env-host/src/host.rs
dmkozh: I don't think this should depend on such optimizations - with rc the key sizes will be sufficiently small and this anyway is a very minor optimization comparing to the regular storage that stores entire ledger keys.

dmkozh:
> I'm not 100% clear on the patterns that motivate this -- could you describe in a little more detail?
> 

I've provided an example in the PR description - this is to implement patterns like temporary allowance (increase token allowance, then use it within the same host instance). This will be mostly useful for tokens (both built-in and custom) and maybe some similar contracts. So this has pretty narrow use and is more optimal then the regular storage, so I wouldn't be worried about minor optimizations that aren't possible yet, like caching contract ids.

soroban-env-host/src/storage.rs
graydon: As I said over there: all scvals are comparable because they can occur as keys in maps, because maps are polymorphic -- they can be keyed by any scval. You can make a map that is keyed by other maps. It's not _common_ but we'd have to develop much more elaborate special-casing of comparable and non-comparable scvals to prohibit it. the general case is "all scvals are comparable".

soroban-env-host/src/host.rs
graydon: If they are not identical, it is a very serious bug, because we interconvert between them constantly, we use binary search to look up map entries in host objects, and "map is always sorted" is an invariant lots of stuff relies on.

I will not claim this bug is impossible or even that we have enough testing to rule it out in the current codebase. Merely that it is essential that it not ship in the final product already.

soroban-env-host/src/storage.rs
dmkozh: This is not about scvals or any types being comparable. The issue is that currently the map requires comparison for *values* (not just keys). I got convinced that it's ok-ish in the initial discussion in #624, but now that I think about it, it seems just wrong because values should never be compared in the map (even if we implement map via sorted tuple vector) and if they are, then something is wrong with the implementation (unless the map is a multimap, that is).

soroban-env-host/src/host.rs
dmkozh: Ok, if that's the invariant, then it should be ok to use the `RawVal`s. I hope we'll find out soon enough if there is some bug...

graydon:
>  this is to implement patterns like temporary allowance (increase token allowance, then use it within the same host instance)

Sorry, I wasn't clear: I saw that example at the top of the PR, I just don't understand what it means. I want to make sure we're designing an interface that makes sense given whatever it means. 
  - Why would allowances temporarily increase?
  - Why temporary rather than permanent? Why the difference in semantics?
  - Are temporary allowances supposed to be observed by just one VM-frame-worth of a contract and its internal methods (such that it could pass the temporary increase around as a local argument during execution) or is it necessary that the temporary increase is observed by all VM-frame activations of the same contract?
  - If this has to do with allowances, and allowances are normally stored as persistent contract state, does this mean that any contract that checks its allowance has to check both its persistent _and_ its "temporary" allowance and sum them up before acting? If it has both a temporary and permanent allowance, which one is correct? Should they both change when a transfer is made? This all seems a bit dangerous to leave in the hands of users; I would have assumed the interface instead would be something like an override of persistent storage with a temporary entry that is defined to revert at the end of the transaction, or something. But I'm not sure.

Again, I'm not saying "this is bad", just that I don't _understand_ what a temporary allowance is useful for or how it's intended to be used, so it's difficult to comment on the design.

You can tell me that it's not important to get any input on the design from me, I don't need to understand it, and I guess I won't hold things up on the basis of my not-understanding if "it's important for the builtin contract" (not all of which I understand anyways), but I was asking in case it was easy to explain, because it's a new interface representing stateful and supposedly meaningful semantics and I am a little nervous randomly approving stuff I don't understand the motivation for.

dmkozh:
> Why would allowances temporarily increase? Why temporary rather than permanent? Why the difference in semantics?

The 'classic' ERC-20 style allowances were/are widely used in various contexts just due to the fact that it's hard to bundle authorization together due to reliance on msg.sender. Basically any sort of scenario that can be templated as 'allow contract to spend token in order to do something' had to be performed via allowance in two transactions. That's the reason for why the allowance needed to be persistent in the first place. It also led to people becoming lazy and using unlimited allowance, which in turn led to exploits that abused these allowances. Of course, there were some efforts to improve the allowance safety, including limiting the time scope of the allowance.

 In Soroban thanks to the ability to make authorized cross-contract calls easily a lot of these usage patterns don't need allowance at all. For example, a simple 'transfer token to contract, do something' scenario can be performed via a simple `xfer` called from the contract. There are ideas of how this can be generalized to support more scenarios - https://github.com/stellar/rs-soroban-env/issues/653 discusses some of these. 

However, allowance-based approach still seems useful for certain scenarios. For example, if the token balance needs to be distributed between multiple parties that change dynamically, the signature that lists all the approved `xfer` operations would become invalid really quickly. This can still be worked around by withdrawing all the token amount to the contract and then refunding the unspent amount. However, this is not very intuitive and probably is error-prone. Instead, we could create a temporary allowance and then do the distribution via `xfer_from` - that would have more familiar semantics and won't require contract writers to come up with the refund logic and manual balance tracking.

Now to the question about persistence: considering the security implications, I'd say the question should rather be 'why persistent and not temporary?'. Basically the shorter the allowance timespan, the better. For the scenarios similar to what I've described above, there is no need to persist anything at all, hence there is really no reason to persist the allowance in the first place.

The discussion on whether persistent allowances are needed at all is still on-going. I would be happier if we could get rid of them completely, but maybe there are some legitimate scenarios where it's safe to persist the allowance. This is still TBD.

To be clear, allowance is just the most obvious example. I believe there are other scenarios where storing the data temporarily would be cheaper and safer then creating and then cleaning up the persistent data.

> Are temporary allowances supposed to be observed by just one VM-frame-worth of a contract and its internal methods (such that it could pass the temporary increase around as a local argument during execution) or is it necessary that the temporary increase is observed by all VM-frame activations of the same contract?

The temporary entries should be alive for the top-level contract call execution time in order to allow persisting data between sub-contract calls (so that's the latter in your question). There is no value for single-frame 'storage' - that would just be a local variable, right?

> If this has to do with allowances, and allowances are normally stored as persistent contract state, does this mean that any contract that checks its allowance has to check both its persistent and its "temporary" allowance and sum them up before acting? 

I don't think so. The exact token interface is TBD, but I don't think these should be mixed. Probably the default interface should only work with temp allowances and persisted allowances should be used separately with some explicit interface and disclaimer on safety issues. This is still up to debate though.

> This all seems a bit dangerous to leave in the hands of users; I would have assumed the interface instead would be something like an override of persistent storage with a temporary entry that is defined to revert at the end of the transaction, or something.

Mixing the persistent and temp storage seems less efficient to me. Besides the obvious benefit of not paying for the IO and ledger rent, temp storage also can safely bypass the footprints (which reduces transaction size) and avoids ledger key conversions. I also think that mixing these two is a bit confusing, as it's not clear as to what should happen if there is an entry in persistent storage with the same key (simple revert won't work - imagine you have persistent allowance of X, then you add Y temp allowance and spend Y + X allowance - the final persisted value should be 0, not X). I might be wrong though - maybe SDK interface should make these interchangeable. But it seems to me like it is more efficient to distinguish between these two at the host level.

soroban-env-host/src/host.rs
dmkozh: Updated to use RawVals.

soroban-env-host/src/storage.rs
graydon: Ok. I think you shouldn't have to impl anything here at all -- `Compare<RawVal>` is impl'ed already -- but regardless I've redone (and expanded a bit) the task you set on in #624 in #739 and am ok with it if you really feel strongly.

soroban-env-host/src/host.rs
graydon: *Shrug* ok I would like to do it eventually but if people are blocked on this we can land now.

graydon:
@dmkozh Thanks for the detailed explanation, it makes more sense to me in that light and I don't mean to impede it. I'll try to help get the test working and I think if we get the comparable-value issue cleared up we should be good to go.

soroban-env-host/src/storage.rs
dmkozh: Yes, I've switched to `RawVal`s for keys and removed this alias completely.

graydon (approved):

PR #739, opened 2023-03-22
Remove unnecessary trait bounds on MeteredVec and MeteredOrdMap

graydon:
This is a repeat (and moderate expansion) of https://github.com/stellar/rs-soroban-env/pull/624 basically because the redundant trait bounds are still annoying @dmkozh and I don't feel strongly enough to argue the point anymore. This came up again in https://github.com/stellar/rs-soroban-env/pull/732 and removing them is harmless, besides the point I was trying to make earlier that they are informative / documentation. They're unnecessary.


dmkozh (approved):
Thank you for this change! Sorry for insisting on this so much, I realize it's not a huge deal. I do feel safer though knowing that we don't accidentally compare values, even given the map's narrow use cases.

Issue #311, opened 2022-08-03
(Epic) Define semantics (and testing strategy) when SDK version is revved


MonsieurNicolas:
This is similar to #310 but scenarios are around SDK changes that are not impacting the core protocol.

Context:
* contract `A` and `B` are compiled with SDK 1.0.0
* contract `A` depends on `B` (ie: `A` invokes `B`)

Scenario:
* `A`'s developer wants to upgrade to SDK 2.0.0 and wants to test what happens when calling `B` compiled with version 1.0.0
* `B` gets recompiled using SDK 2.0.0, `A`'s developer wants to test and understand how `A` would be impacted.

This is somewhat related to #289

leighmcculloch:
> * contract `A` depends on `B` (ie: `A` invokes `B`)

There are a couple variations of this. For example:
1. contract `A` invokes contract `B`
2. contract `A` imports the Rust crate for contract `B`

For (1) the compatibility layer is the host layer, and so the SDK plays very little role in there being compatibility or incompatibility between `A` and `B`. Except for the contract spec. `B`s contract spec might contain elements that are new if it is using a new SDK, in which case it would be ideal if `A` fails to build gracefully, indicating that a newer or unrecognized version of the contract spec is in use and `A` needs to use a newer SDK. The plan is to not make any breaking changes to the contract spec, so if `A` is using a newer contract spec that should have no bearing on anything. The contract spec has versioning already embedded into it for these cases.

For (2) things are much more fuzzy and I don't think we should do much other than use semver to the best of our ability. If anything in the SDK is a breaking change, increment the major version, and developers will use Cargo and versioning definitions to appropriately constrain their compatibility like they do for any Rust crate.

MonsieurNicolas:
> For (1) the compatibility layer is the host layer, and so the SDK plays very little role in there being compatibility or incompatibility between A and B. Except for the contract spec

yes, this is why I separated the "host compat issues" and this issue that tracks changes that may impact the contract spec in the broader sense (behavior included).
This includes issues related to custom types (UDTs), but I am sure there are more issues like if certain shared (embedded) functions/types behave slightly differently for example (therefore testing is needed).

[...]

PR #747, opened 2023-03-31
Sketch of a subtype judgment on contract specs

graydon:
This is a sketch of a subtype relationship on soroban contract specs. The idea here is to support the host function that updates / overwrites contracts with a judgment it can make that a new contract is compatible with the spec (type signature) of an old contract, such that anyone using the old contract will continue to be able to use the new contract. 

Of course it's just "spec-level" compatibility, not like "does the contract do the same thing" -- we can't judge that -- but I think it'll prevent a fairly wide class of upgrade mistakes if we can at least catch spec mismatches.

The subtyping relationship here is intended to be _extremely textbook_ and hopefully corresponds to the same rules (covariant returns, contravariant arguments, record extensions, union restrictions) you'll find in any standard treatment of immutable structural types and functions.

We haven't 100% decided we're going to do this yet, I don't think, but I suspect something like this (possibly strengthened or weakened somewhat) will prevent a lot of errors, and I wanted to sketch it out as part of discussing the lifecycle, evolution and versioning of contracts in the field (eg. as part of https://github.com/stellar/rs-soroban-env/issues/311 though that bug somewhat mixes up issues of contract compatibility and SDK compatibility, I think the spec mechanism is likely to be at least part of how we express SDK versioning anyways)

cc @sisuresh, @MonsieurNicolas, @dmkozh and @leighmcculloch

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -52,6 +52,7 @@ pub(crate) mod metered_clone;
 pub(crate) mod metered_map;
 pub(crate) mod metered_vector;
 pub(crate) mod metered_xdr;
+mod spec;
 mod validity;
 pub use error::HostError;
 
--- a/soroban-env-host/src/host/spec.rs
+++ b/soroban-env-host/src/host/spec.rs
@@ -1,0 +1,371 @@
+use crate::{budget::Budget, xdr, HostError};
+use soroban_env_common::xdr::{
+    ScSpecFunctionV0, ScSpecTypeDef, ScSpecUdtEnumCaseV0, ScSpecUdtEnumV0,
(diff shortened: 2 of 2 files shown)


leighmcculloch:
I think this functionality could live in the soroban-cli. I can imagine the soroban-cli having a subcommand for checking the compatibility of two contracts, either two on disk, or one deployed and one on disk.

I'm hesitant about this living in the host because other than this the host has no knowledge about SDK types or contract interfaces. Adding this logic to the host itself would as far as I can tell make this a part of the protocol proper, rather than being details external to the protocol.

tsachiherman:
I think that these are very strong "language features" that would be great to add to the cli.
Moreover, for implementing dynamic cross contract calls, I think that these would become useful.
Reading your description, it sounds like a linked list of specEntries. I'm wondering if these could get to a tree like or circular list. ( Since the cli would need to avoid these )

Issue #734, opened 2023-03-21
Decouple "iteration" from "input" in budget charging


jayz22:
During review of https://github.com/stellar/rs-soroban-env/pull/704, @dmkozh helped pointing out that in our current budget framework we have a mixing of the two concepts. 
_Input_ - which is the dimension going into the cost model to compute the cost.
_Iteration_ - how many times to charge a model.
A linear model, such as `ValSer`, multiplies a constant by the input (length of the buffer) to produce a cost, this reflects the linear complexity of serialization. 
For a constant-complexity operation, such as the `WasmInsnExec`, we batch multiple wasm instructions and charge them together to reduce the number of host roundtrips. We should be charging them as `constant model * num of iterations`. However, in our current treatment, we treat the model as linear and "num of iterations" as the input. 
We need to restructure the budget charging framework to separate out these as two separate dimensions.

jayz22:
Resolved by https://github.com/stellar/rs-soroban-env/pull/753

Issue #709, opened 2023-02-24
Cost parameter extraction


jayz22:
We currently have the calibration code in place to produce the cost characteristics for each [cost type](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/budget.rs#L11). However, we lack the heuristics and automatic decision tools to extract those into the table of [cost parameters](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/budget.rs#L457). We need:
- Heuristics to distinguish linear, constant and sublinear-with-upperbound cases. 
- For each case, use a fit function to extract parameters. 

These parameters should be stored in a table inside the repo for reference, along with some meta information (e.g. versioning) for reproducibility.

jayz22:
Resolved by https://github.com/stellar/rs-soroban-env/pull/753

PR #753, opened 2023-04-06
Decouple `CostModel` `inputs` from `iterations`; fit and extract model params

jayz22:
### What

Mainly addresses https://github.com/stellar/rs-soroban-env/issues/734 and https://github.com/stellar/rs-soroban-env/issues/709, along with a few other improvements. 

The cost model for cpu and mem budgeting has been enhanced to `f(x) = N * (a + b * Option<x>)` -- the input `x` is now optional, and iterations `N` is explicit and always present. 
This split has major implication on the bench setup, which is the major part of this change set.

This PR also adds model fitting and parameter extraction:
- For a constant model, simply take the average of the costs over the number of iterations.
- For a linear model, fit a linear regression model over the `(inputs, costs)` measurement.

The cost parameters have been updated with the extracted results.

This PR also adds several "ease of life" improvements, such as
- Adds comments explaining what `HostCostMeasurement` is doing and what different types of overheads are.
- Adds a smoke test on the budget with some random input to prevent accidental change of model parameters.
- Improve display of Budget (splits `Debug` from `Display`), improves the table display of the benchmark results.
- Add command line filter to which wasm insns type to run calibration for.

See full output of the calibration: ~[output8.txt](https://github.com/stellar/rs-soroban-env/files/11172250/output8.txt)~
[output10.txt](https://github.com/stellar/rs-soroban-env/files/11226807/output10.txt)

### Why

This is to address the overloaded use of `input` previously
- For a linear model, the input `x` is the characteristic representing the work's complexity, e.g. length of the buffer when doing a memcpy. 
- In a constant model, the work is a constant cost that is independent any the input. In other words, the input is undefined. 
Unfortunately, previously the input was entangled with the number of iterations the const work is repeated (due to e.g. batching), and thus many cost types were mis-characterized (e.g `VecEntry`). 

### Known limitations

[TODO or N/A]


soroban-env-host/src/cost_runner/cost_types/compute_ed25519_pubkey.rs
```diff
@@ -22,10 +22,11 @@ impl CostRunner for ComputeEd25519PubKeyRun {
     }
 
     fn run_baseline_iter(
-        _host: &crate::Host,
+        host: &crate::Host,
         _iter: u64,
         sample: Self::SampleType,
     ) -> Self::RecycledType {
+        black_box(host.charge_budget(Self::COST_TYPE, 1, None).unwrap());
```
graydon, line 29: I feel like you could make `run_baseline_iter` an defaulted member of the trait if `Self::RecycledType : From<Self:::SampleType>`, no?

soroban-env-host/benches/common/measure.rs
```diff
@@ -127,31 +137,32 @@ impl Measurements {
...
             .padding(5)
             .alignment(Alignment::Right);
 
-        writeln!(
-            &mut tw,
-            "input\tinput_baseline\tcpu_insns\tcpu_baseline\tinsns/input\tmem_bytes\tmem_baseline\tbytes/input\ttime nsecs\tnsecs/input"
-        )
-        .unwrap();
-        for m in self.measurements.iter() {
+        writeln!(&mut tw, "iterations\tinputs\tcpu_insns_total\tcpu_insns_base\tmem_bytes_total\tmem_bytes_base\ttime_ns_total\ttime_ns_base\tinput/iter\tnet_insns/iter\tnet_bytes/iter\tnet_ns/input").unwrap();
```
graydon, line 149: it seems to me this is missing net_insns/input which is what the model-fit will report below -- might be nice to eyeball that?

soroban-env-host/src/cost_runner/cost_types/compute_ed25519_pubkey.rs
jayz22: In most cases the `RecycledType` contains the `SampleType` plus the return type of the function under measurement. So we can't require `Self::RecycledType : From<Self::SampleType>`?

soroban-env-host/benches/common/measure.rs
jayz22: Ok, will add that.

soroban-env-host/benches/common/mod.rs
```diff
@@ -40,70 +40,83 @@ fn should_run<HCM: HostCostMeasurement>() -> bool {
     }
 }
 
+fn should_run_wasm_insn(ty: WasmInsnType) -> bool {
+    if let Some(bench_names) = get_explicit_bench_names() {
+        let name = format!("{:?}", ty).split("::").last().unwrap().to_string();
+        bench_names.into_iter().find(|arg| *arg == name).is_some()
```
dmkozh, line 46: nit: `.any()`?

soroban-env-host/src/budget.rs
```diff
@@ -449,79 +547,239 @@ impl Default for BudgetImpl {
...
+                }
+                CostType::MapEntry => {
+                    cpu.const_param = 56;
+                    cpu.lin_param = 0;
+                }
+                CostType::VecNew => {
+                    cpu.const_param = 0;
+                    cpu.lin_param = 0;
+                }
+                CostType::VecEntry => {
```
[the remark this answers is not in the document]
dmkozh: Also, another consideration: if the cost is 0 both for memory and CPU, why have it in the first place? Maybe we should eliminate no-ops?

dmkozh (commented):
This approach looks much easier to digest, thanks! Just left a few minor nitpicks and a question.

soroban-env-host/src/budget.rs
```diff
@@ -118,23 +123,38 @@ impl CostType {
 
 #[derive(Default, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct CostModel {
-    pub const_param: u64,
-    pub lin_param: u64,
+    pub(crate) const_param: u64,
+    pub(crate) lin_param: u64,
+    pub(crate) iterations: u64,
```
graydon, line 128: I don't think `iterations` should be stored in the cost model, should it? I thought it was passed in from the budget-charging call site.

jayz22: Yes, it is now part of the input `f(x) = N * (a + b * Option<x>)` as `N`. 
It is part of the model because `N` is tied to a particular `x`, i.e. charges with different inputs cannot be charged together.

soroban-env-host/src/budget.rs
```diff
@@ -118,23 +123,38 @@ impl CostType {
...
 }
 
 impl CostModel {
-    pub fn evaluate(&self, input: u64) -> u64 {
-        let mut res = self.const_param;
-        if self.lin_param != 0 {
-            res = res.saturating_add(self.lin_param.saturating_mul(input));
-        }
-        res
+    /// Evaluates the linear model `f(input) = const_param + lin_param * input`.
```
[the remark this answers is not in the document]
jayz22: I will update. Thanks for the catch.

soroban-env-host/src/budget.rs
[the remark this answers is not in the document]
jayz22: Yeah it is a little suspicious these are 0 (the baseline run costs more than the actual runs) I'll double check.

soroban-env-host/benches/common/measure.rs
```diff
@@ -160,32 +171,63 @@ impl Measurements {
...
+        fit_model(x, y)
     }
 }
 
-/// HostCostMeasurement is an interface to measuring the memory and CPU
+/// HostCostMeasurement (HCM) is an interface to measuring the memory and CPU
 /// costs of a CostType: usually a block of WASM bytecode or a host function.
+/// All types of costs are modeled by the linear function (see [`CostModel`]):
+///
+///     f(x) = N_x * (a + b * Option<x>)                                    [1]
```
graydon, line 206: Hm, I think this should be `Option<b>` not `Option<x>`, right? And really `None` just means `0` in the model?

I'm also unclear on why `N_x` is something you're putting in the model rather than just tracking during measurement. The number of iterations spent _measuring_ doesn't seem to me conceptually related to the predicted cost when evaluating the model on a particular input size.

soroban-env-host/src/budget.rs
```diff
@@ -98,12 +101,14 @@ impl CostType {
 
 /// We provide a "cost model" object that evaluates a linear expression:
 ///
-///    f(x) = a + b*x
+///    f(x) = N * (a + b * Option<x>)
```
graydon, line 104: I .. am unsure about this. My instinct is that the iteration count shouldn't wind up in the cost model at all. I think iteration counts should be limited to the calibration harness code. Especially I am unsure about the thing you're doing here where you describe the iteration count as similar to the _batching_ count that we use in callers (but then in the actual batching calls, eg. in `MeteredClone`, you're not using this, but passing `1` for iteration and `Some(n_elts.saturating_mul(T::DECLARED_SIZE))`.

I think there are 3 separate issues at work here:

1. A statistical iteration count that runs the same trial multiple times to average-out noise during calibration.
2. Charging the budget for a "bulk operation" like memcpy or memcmp where there's both an "element size" and an "element count"
3. Making a single budget _charge call_ that does a "bulk" call to `Budget::charge` that combines multiple logically-separate calls to `Budget::charge`, to avoid too much CPU overhead from the budget-charging-machinery

Cases 2 and 3 look similar but they're not, right? Like a call to something like `memcpy` is case 2 and _not_ case 3, because a single call to `memcpy` has a fixed constant term `a` -- the overhead of calling the function and setting up the copy loop -- and its variable cost is entirely in the `b` term, not the `N` term. Case 3 only occurs when we have a charge that has an `N` term that varies, because you want to multiply `a*N` to capture "a batch of `N` calls to `Budget::charge` where each call has constant term `a`", right?

So .. looking through the code I am not sure case 3 ever actually happens (outside of calibration). Or at least that it's correct for it to happen. I see you're _not_ doing case 3 for things like `memcmp`, but _are_ doing it for things like `VecEntry`, which .. I guess is plausible? But then if that ever _is_ the case, you're no longer dealing with a simple linear model, you've got a _bi-linear_ model right? Which .. I don't think is actually what the model fitter is fitting, and also doesn't quite look like what the evaluation function is doing (also I suspect it's overkill for these few cases, i.e. I don't expect `N*(a+bx)` for `N` vector element accesses is going to be a lot different from `a+b(N*x)` which is just `a+bx` for a different `b` (which I think will get divided-out during measurement anyways).

soroban-env-host/src/budget.rs
```diff
@@ -118,23 +123,38 @@ impl CostType {
...
+            Some(input) => {
+                let mut res = self.const_param;
+                if self.lin_param != 0 {
+                    res = res.saturating_add(self.lin_param.saturating_mul(input));
+                }
+                res
+            }
+            None => self.const_param,
+        };
+        iterations.saturating_mul(eval)
```
graydon, line 150: So here you're using the iteration count _passed in_ and ignoring the one stored in the model. Which I think is closer to making sense for case 3 above, but .. again I'm not 100% sure the cases that use it outside of calibration really need it.

graydon (changes requested):
I've pointed out some places where I think the model code is a bit confused in terms of what `iteration` means and where it's used. It's conceivable that this is useful for batching calls to the charge interface but I'm not sure that's actually how it's being used in the code nor that it's conceptually necessary in very many places.

Moreover, reading through the code in the actual call sites, I'm .. fairly uncomfortable with this change as-is. It seems like the call sites have actually become _more_ error-prone since it's unclear to someone writing metered code which calls should be `charge(ty, 1, Some(val))`, or `charge(ty, 0, Some(val))`, or `charge(ty, val, None)`, or `charge(ty, val, Some(other))` or whatever. Like I'm not sure the ones that do `charge(ty, val, None)` _should_ have `None` as their third parameter. Isn't that up to the model?

I think there are relatively few cases where we do a non-1 iteration count; maybe we could just add a method called `charge_many(ty, iter, input)` and then leave `charge(ty, input)` as-is?

jayz22:
@graydon thanks for your review and detailed thoughts. Here is an overall reply to your three comments above, since they are all touching similar problems. 
Regarding the model representation, the two issues you brought up:
1. `Option<x>` vs `Option<b>`: Should x be optional, or should b? I pretty strongly believe `Option<x>` is the more accurate form. What we are trying to do is: express all types of cost with a general model up to order 1. Which means 
Screen Shot 2023-04-12 at 11 39 59 AM
`x` should be optional, meaning the model is "up to order 1". `Option<b>` would mean the model *is* order 1 but its coefficient may or may not exist, which does not make sense.
2. Should `N` be part of the model? Initially I thought so. I thought of bulk charging as the use case, Whether bulk charging N constant costs (of the same type, e.g. `WasmInsnExec`), or bulk charging N linear costs (e.g. memcpy of N elements of the same type). I thought N should act as a constant scaling factor in the model that is associated with `X` -- we cannot bulk costs with different `X` together. I.e. the old `input` is now a tuple `(N, X)` However, this has turned out to be misused (as you correctly pointed out above), and is hard to use correctly. It turned out to be just an "accounting" factor that doesn't belong as part of the model. In terms of reporting, the tracker reduces `{(N, X)}` into a `(sum({N}, sum({X}))`, destroying their connections. I will remove N out of the model and make it part of the cost charging.

In terms of the use cases. It is interesting you laid out case 2 and 3 as different cases. I just thought of them as "bulk charging", for both things that are identical (memcpy of N bytes) and things that are logically separate but governed by the same formula (under the same `CostType` and having the same `X`) such as `WasmInsnExec`. I don't feel it's worth to differentiate them too much as long as we make sure they are both done correctly (our model is not first-principle based but empirical based). 

The "memcpy" use case is an interesting one, which I have missed and it was an oversight. I would have put `n_elts` as the "iteration" and `Some(declared_size<T>)` as the input had I not missed it. But as you pointed out, memcpy maybe more accurately expressed as a constant setup cost (setting up the loop, locating the memory) plus a linear component w.r.t. the num_of_bytes (copying the bytes). So it might be more accurate as a linear model with no bulk factor (N=1), which is what it is now. 

If this in general sounds reasonable to you, let me update this PR to remove N out of the model, and make the "bulk" vs "single_charge" interfaces more clear and separate, and then let you judge if it all makes more sense to move forward or if bigger changes are required. What do you think?

jayz22:
@graydon I've pushed a change to remove iterations from cost model and splitted out the batch-charging case from normal charging, which retains the old interface and is the vast majority. I still retained the `Option<X>` as explained above. This should clean up the cost-charging interface quite a bit. 
Let me know if this is something we can work with, or if you feel bigger changes are still necessary.

graydon:
@jayz22 Yeah, ok, I .. don't feel strongly enough about the location of the `Option` to argue it further, and with these changes the ambiguity at call sites is limited to `batched_charge` sites, of which are few. I'll approve and merge, thanks!

graydon (approved):

graydon:
Oh, actually @jayz22 it looks like @dmkozh has 2 remaining nits plus the issue we both noted concerning the remaining 0-cost cases. Can you take a look at those before we merge?

soroban-env-host/src/cost_runner/cost_types/compute_ed25519_pubkey.rs
jayz22: I know you've previously resolved this but I just wanted to give this suggestion (what I understood) a try. 

I tried defining the `RecycledType` as a tuple `(Option<SampleType>, Option<ReturnType>)` and to implement the `run_baseline_iter` as "charge + return the sample". This won't work for a couple of reasons:

1. Associated type is an unstable feature, without it the code in  `HostCostMeasurement` looks very ugly, since it needs to set up a container `Vec<RecycledType>` the recycled results. (This is bad but not deal breaker)
2. The `SampleType` is not always returned in full, in some cases, part of it is consumed by the runner and other part of it needs to be returned. So the general definition above doesn't work, and each runner type really needs to define its own RecycledType.

soroban-env-host/benches/common/measure.rs
jayz22: I have added what you suggested and also renamed some column names e.g. "net_insns/iter" --> "ave_net_insns" to make it cleaner.

soroban-env-host/benches/common/measure.rs
```diff
@@ -60,19 +61,28 @@ pub struct Measurement {
 pub struct Measurements {
     pub baseline: Measurement,
     pub measurements: Vec<Measurement>,
+    pub net_meas_per_iter: Vec<Measurement>,
```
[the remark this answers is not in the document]
jayz22: Fixed the naming.

soroban-env-host/benches/common/mod.rs
jayz22: Fixed.

soroban-env-host/src/budget.rs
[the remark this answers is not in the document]
jayz22: I found out in `VecNewRun`, the sample was not recycled properly in the "baseline run" and it was doing a deallocation, therefore deflating the baseline case quite a bit. Resulting in the actual result (sample run minus baseline run) being all zeros (actually way less than 0). 

After correcting it, the sample run and baseline run results are nearly identical. The `VecNew` (and `MapNew` too) is not doing anything significant, it is just wrapping the input (`Vec<T>`) in a struct. 

As a result, I've removed `VecNew` and `MapNew` and their corresponding measurement, runner etc. (This also cleans up the existing code quite a bit, a lot of the callers don't need to pass in a budget any more)

jayz22:
@graydon @dmkozh All comments have been addressed, thank you.

soroban-env-host/src/budget.rs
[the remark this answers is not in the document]
jayz22: (I've also updated the output file in the PR description above)

soroban-env-host/src/cost_runner/cost_types/compute_ed25519_pubkey.rs
graydon: ok, cool. thanks for exploring!

PR #754, opened 2023-04-06
Add soroban_env_host::host::LegacyEpoch

graydon:
This adds a new mechanism for gating old behaviour in the host. Currently it's not used, but the idea is that as we discover divergence we will need to add branches to the code here and there to run "as if" we were an old version.

This is _similar_ to the way protocol-version (and occasionally ledger-number) gating of backward-compatibility code happens in stellar-core, except that (a) ledger-number gating is both brittle and uninformative and (b) we don't have clear enough insight _in advance_ of making a breaking change that we're doing so, due to the size of the dependency tree. So instead the embedder is forced to do a multi-step upgrade process, with two versions of the entire host active at once. The embedder _should_ be able to do this correctly since it both sees the entire dependency tree of each host version it embeds, and can compare the history segment between the two for divergence and add legacy epochs to the host as necessary in order to achieve accurate replay.

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -160,8 +160,82 @@ impl Default for DiagnosticLevel {
     }
 }
 
+/// LegacyEpoch defines a set of _past_ (legacy) behaviour that we have logic in
+/// the host to handle, that can be activated by the embedder to replay past
+/// contracts "the same way they ran before".
+///
+/// Legacy epochs are added when some change in behaviour is detected by the
+/// embedder, during the second phase of the embedder upgrading the host, when
+/// the embedder performs a replay of some past transactions in order to expire
+/// old versions of the host, and notices some behaviour that diverges: that
+/// does not replay identically. At that point the _current_ behaviour is always
+/// considered part of `LegacyEpoch:Current` and a new _earlier_ epoch is added
+/// before it, that is named for the _previous_ behaviour.
+///
+/// To illustrate this with an example, first suppose that host versions are
+/// being _sequentially numbered_ by the embedder as v74, v75, v76, v77, etc. We
+/// do not assign such versions in the host itself, since the host does not know
+/// precisely when the embedder takes updates (and in any case the embedder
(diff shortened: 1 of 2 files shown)


soroban-env-host/src/host.rs
```diff
@@ -158,8 +158,31 @@ impl Default for DiagnosticLevel {
...
+/// the host to handle, that can be activated by the embedder to replay past
+/// contracts "the same way they ran before".
+///
+/// Legacy epochs are added when some change in behaviour is detected by the
+/// embedder -- because behaviour does not replay correctly -- at which point
+/// the _current_ behaviour is always considered part of `LegacyEpoch:Current`
+/// and a new _earlier_ epoch is added before it, that is named for the
+/// _previous_ behaviour.
+///
+/// For example, if we had a change to hashing that made it much cheaper, when
```
dmkozh, line 171: nit: I think this example is a bit misleading because IIUC the metering calibration data is configured via a separate mechanism (config settings) and there shouldn't be an issue about e.g. optimizing the function without changing the calibration (or vice versa). I suppose this should be meaningful for the logic changes (e.g. if a function had a bug or if pre-conditions have changed), but you probably have a better idea of what this covers.

jayz22: This can happen if the input or number of iterations of work has changed, while the unit of work (parameterized and stored in config settings) has not. 
Going along with this example, if the unit of work is hashing 1 byte of data, and is a constant cost. With the faster hashing algorithm, it requires hashing fewer bytes of data for the same object, thus fewer iterations. Then the overall cost will be reduced.

graydon: Ok, I'll make a different example

graydon: @dmkozh I've updated the comment here with much more detail as well as an example that's more realistic / not related to metering cost differences. Let me know if this is adequate.

dmkozh: This looks good, thanks!

dmkozh (approved):

Issue #743, opened 2023-03-24
Verify that equivalent RawVals and ScVals compare the same


brson:
Per https://github.com/stellar/rs-soroban-env/pull/732/commits/225cf4ff02b7eaaba898645591985b577c5d1a1b#r1144119247 it is crucial that for all equivalent pairs of RawVals and ScVals, the pair has the same comparison results.

It looks to me that RawVal uses a custom Compare trait - RawVals don't appear to implement Ord; ScVal does implement Ord. So presumably the result of Compare on RawVals needs to be equivalent to the results of Ord::cmp (and PartialOrd::partial_cmp).

This should be testable with the fuzzing infrastructure in https://github.com/stellar/rs-soroban-sdk/pull/878 so I intend to try that. Fuzzing won't guarantee all possible values compare the same, but better than nothing.

cc @graydon @dmkozh

brson:
I have started fuzzing as described and have found differences in how RawVal and ScVal implement comparison.

ScVal's derive Ord, and RawVal implements Compare by hand, so to remediate issues I am currently focusing on making the RawVal implementations match the derived ScVal implementations.

It may be prudent though to write the Ord impls for XDR types by hand if it is important that they not change. As an example, the ScVal enum's Ord impl depends on the lexical order of ScVal variants, something that is prone to changing as the XDR definitions change.

brson:
If it is crucial for these comparison functions to be the same then perhaps it is possible to just remove the Ord impl from ScVal and always do comparisons aRas wVals. It seems very high-risk to have these two impls that might diverge over time, even if we fix them once with some confidence.

brson:
Thanks for the clarifications @dmkozh.

I've continued this fuzzing under the assumption that the only comparisons that matter are those where one of the RawVal or ScVal comparisons returns Ordering::Equal.

Is it important that `impl Compare<ScVal> for Budget` has the same logic as `impl Compare<RawVal> for Host` and/or `PartialOrd for ScVal`?

I have found one bug in `Compare<ScVec> for Budget` where differing vector lengths are not accounted for.

graydon:
[...]

PR #762, opened 2023-04-10
Comparison fixes

brson:
### What

Three patches toward making `Compare<RawVal> for Env` and `Compare<ScVal> for Budget` behave identically to `Ord for ScVal`, per https://github.com/stellar/rs-soroban-env/issues/743.

These are the initial patches described in my comment here: https://github.com/stellar/rs-soroban-env/issues/743#issuecomment-1496661230

### Why

As discussed in that issue, all three of `Compare<RawVal> for Env`, `Compare<ScVal> for Budget` and `Ord for ScVal` need to behave the same. These patches fix the first three discrepancies I have found through fuzzing.

### Known limitations

This adds some hand-written unit tests to prevent regressions, but I intend to add comprehensive proptests / fuzz tests to the soroban-sdk crate eventually. Such tests depend on some fuzzing-related patches that are not yet ready to land.

### DIFF
--- a/soroban-env-common/src/status.rs
+++ b/soroban-env-common/src/status.rs
@@ -61,8 +61,8 @@ impl PartialOrd for Status {
 impl Ord for Status {
     #[inline(always)]
     fn cmp(&self, other: &Self) -> Ordering {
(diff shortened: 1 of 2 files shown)


graydon (approved):
Great, thanks!

PR #765, opened 2023-04-14
Don't allow invalid chars in Symbol objects.

brson:
### What

Validate that large Symbol values do not contain chars that are not allowed in Symbol types.

### Why

The conversion from str to Symbol falls back to the object path if a SymbolSmall cannot be created. The Host::symbol_new_from_slice function though does not do the required validation:

```rust
impl<E: Env> TryFromVal<E, &str> for Symbol {
    type Error = ConversionError;

    fn try_from_val(env: &E, v: &&str) -> Result<Self, Self::Error> {
        if let Ok(ss) = SymbolSmall::try_from_str(v) {
            Ok(Self(ss.0))
        } else if let Ok(so) = env.symbol_new_from_slice(v) {
            Ok(Self(so.0))
        } else {
            Err(ConversionError)
        }
    }
}
```

Symbol::new_from_slice creates an `ScSymbol` and puts it into the environment, but ScSymbol is (currently) allowed to contain invalid chars:

```rust
    fn symbol_new_from_slice(&self, s: &str) -> Result<SymbolObject, HostError> {
        self.add_host_object(ScSymbol(s.as_bytes().to_vec().try_into()?))
    }
```

This patch changes symbol_new_from_slice to prevent invalid ScSymbols from being injected into the environment.

This patch adds several test cases. Only the `invalid_chars` case was failing. The others I added because it wasn't obvious to me whether the cases were being checked or not; especially it was not obvious to me from the docs whether zero-length symbols are allowed (they are). The docs for Symbol should be updated to indicate zero-length symbols are ok, but that patch needs to live in the sdk.

### Known limitations

na

### DIFF
--- a/soroban-env-common/src/symbol.rs
+++ b/soroban-env-common/src/symbol.rs
@@ -157,6 +157,25 @@ impl<const N: u32> TryFrom<&StringM<N>> for SymbolSmall {
 }
 
 impl SymbolSmall {
+    #[doc(hidden)]
+    pub const fn validate_char(ch: char) -> Result<(), SymbolError> {
+        match SymbolSmall::encode_char(ch) {
+            Ok(_) => Ok(()),
+            Err(e) => Err(e),
+        }
+    }
+
+    const fn encode_char(ch: char) -> Result<u64, SymbolError> {
+        let v = match ch {
+            '_' => 1,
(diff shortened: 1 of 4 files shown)


dmkozh (approved):
Thanks for fixing this!

leighmcculloch (approved):
🎉

graydon (approved):

PR #767, opened 2023-04-19
Fix comparison of RawVals of differing types and objectness

brson:
__Edit: I have posted a prerequisite to this patch in https://github.com/stellar/rs-soroban-env/pull/773__

### What

Make comparison of RawVals agree with comparison of ScVals in the case where one value is an object and one value is a smallval, and they are not the same ScValType.

Also fix an unsoundness in RawVal::get_tag due to incorrect assumptions in an unsafe cast.

### Why

Per https://github.com/stellar/rs-soroban-env/issues/743, comparison operations for RawVals and ScVals must agree.

The final fallthrough case of Host::obj_cmp does a comparison based on the Tag discriminant, and the Tag discriminants do not have the same ordering as the ScVal discriminants. This patch adds a method Tag::get_scval_type, and changes obj_cmp to use that instead.

This adds a thorough test case which does an exhaustive conversion between u8s and tags, and an exhaustive match on Tags. The exhaustive match triggered a segfault due to an unsound cast from u8 to Tag in RawVal::get_tag. That function assumed that Tag's discriminants were contiguous between the "marker" discriminants `ObjectCodeLowerBound` and `ObjectCodeUpperBound`, which was not true. get_tag looks like

```rust
    #[inline(always)]
    pub const fn get_tag(self) -> Tag {
        let tag = self.get_tag_u8();
        const A: u8 = Tag::SmallCodeUpperBound as u8;
        const B: u8 = Tag::ObjectCodeLowerBound as u8;
        const C: u8 = Tag::ObjectCodeUpperBound as u8;
        if !((tag < A) || (B < tag && tag < C)) {
            return Tag::Bad;
        }
        // Transmuting an integer to an enum is UB if outside the defined enum
        // value set, so we need to test above to be safe. Note that it's ok for
        // `get_tag` here to be a _little_ slow since it's not called in a lot
        // of small/tight paths, only when doing a switch-based comparison. Most
        // small paths call `has_tag` which tests a _known_ enum case against
        // the tag byte, and therefore doesn't need the range check.
        unsafe { ::core::mem::transmute(tag) }
    }
```

In safe code it may not be possible to trigger the unsoundness - it may require unsafely constructing RawVal. It is not clear to me that contracts can find this code path, but it's probably best to fix it.

I changed the discriminants of Tag to make RawVal::get_tag always correct, and added docs, but the code is brittle to changes. I considered how to make it less error prone and add tests, but could not come up with a better test case than the one that is already in this patch.

This adds a dev-dependency on itertools for the cartesian_product method.

### Known limitations

na


brson:
Actually the test here is not exhaustive in the conversion of u8s to Tags. There could be a better test of RawVal::get_tag, but I'll wait for some review before thinking about it more.

brson:
I can reproduce the CI failures. Seems like I've broken something.

brson:
The CI failure was because changing Tag discriminants is ABI-breaking and requires rebuilding the soroban-test-wasms. I tried, but the workspace does not build presently. Instead I wrote a testcase and simple, possibly temporary, fix and posted a separate pr https://github.com/stellar/rs-soroban-env/pull/773

soroban-env-common/src/raw_val.rs
```diff
@@ -183,6 +183,60 @@ impl Tag {
...
+            Tag::I64Small => Some(ScValType::I64),
+            Tag::TimepointSmall => Some(ScValType::Timepoint),
+            Tag::DurationSmall => Some(ScValType::Duration),
+            Tag::U128Small => Some(ScValType::U128),
+            Tag::I128Small => Some(ScValType::I128),
+            Tag::U256Small => Some(ScValType::U256),
+            Tag::I256Small => Some(ScValType::I256),
+            Tag::SymbolSmall => Some(ScValType::Symbol),
+            Tag::LedgerKeyContractExecutable => Some(ScValType::LedgerKeyContractExecutable),
+            Tag::SmallCodeUpperBound => panic!(),
```
graydon, line 218: Since this code is called in the host and is under the control of the guest, I think it has to be panic-free. Unless I'm misunderstanding how that's being ruled out.

graydon (changes requested):
I think this needs to be panic-free

soroban-env-common/src/raw_val.rs
brson: The three panic cases are the "marker" tags, which I think can't be constructed under guest-controlled code paths. The only calls to `get_scval_type` are in `obj_cmp` here:

```rust
            None => {
                let atype = a.get_tag().get_scval_type();
                let btype = b.get_tag().get_scval_type();
                if atype == btype {
                    // This shouldn't have happened, but if it does there's a logic error.
                    return Err(
                        self.err_general("equal-tagged values rejected by small-value obj_cmp")
                    );
                }
                atype.cmp(&btype)
            }
```

Called on the result of `get_tag` which returns `Tag::Bad` for the cases that might otherwise return the three marker tags.

So I think there is no way for a guest to call this function and trigger a panic.

I debated this myself though and can appreciate an argument for being defensive here and returning `None`.

brson: I have removed the panics.

brson:
I removed the panics from Tag::get_scval_type. I also realized that I had other comparison tests located in another module, so I moved the tests from this PR there.

brson:
Also did a rebase to pick up the commits from https://github.com/stellar/rs-soroban-env/pull/773

soroban-env-common/src/raw_val.rs
graydon: Marker tags can definitely be constructed by guest code -- guest code can construct a `RawVal` with any bit pattern it likes -- but I see what you mean now that the panicking calls should be guarded by `get_tag()`. But yeah, in general I'm aiming for as few such implicit "this panic is safe because..." latent assumptions in the code; it'll be too easy for someone to break such an assumption with a future refactor. Thanks!

graydon (approved):
LGTM, thanks!

Issue #305, opened 2022-08-02
Load cost model parameters from CONFIG_SETTING Ledger entries


graydon:
Continuation of the last work-item in #160, which is somewhat involved.

Currently cost models have their parameters (eg. "CPU cycles per WASM instruction", "memory bytes per map cell", etc.) hard-wired into the budgeting system. They're calibrated using benchmarks but the results of that calibration are then just written as numbers into the source code.

This is acceptable for a test / preview version of the code but longer term (before deployment to a live / public network) they _must_ transition to loading from `CONFIG_SETTING` ledger entries so they can change over time / be loaded from the specific ledger in which a contract runs. Further, these LEs are agreed-to by upgrades during consensus, and so will also require an interface to setting and querying them by operators. There's actually a lot of plumbing to do to make this all work.

PR #770, opened 2023-04-21
Use network configs for metering

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/305

### Why

### Known limitations

[TODO or N/A]


jayz22:
@graydon I've addressed your feedback on removing `CostModel`, replacing it with a trait impl on `xdr::ContractCostParamEntry`. Also updated the xdr hash, this is ready for review.

soroban-env-host/src/budget.rs
```diff
@@ -111,33 +33,75 @@ impl CostType {
 /// The parameters for a `CostModel` are calibrated empirically. See this crate's
 /// benchmarks for more details.
 
-#[derive(Default, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct CostModel {
-    pub(crate) const_param: u64,
-    pub(crate) lin_param: u64,
+// #[derive(Default, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
graydon, line 36: If this commented-out code is dead I think it should be removed.

soroban-env-host/src/budget.rs
```diff
@@ -111,33 +33,75 @@ impl CostType {
...
+//         self.const_param = 0;
+//         self.lin_param = 0;
+//     }
+// }
+
+pub trait HostCostModel {
+    fn evaluate(&self, input: Option<u64>) -> Result<u64, HostError>;
+
+    #[cfg(test)]
+    fn reset(&mut self);
```
graydon, line 77: I don't see anything using this `reset()` method (and it seems a little odd since it's just writing to public fields of a set of parameters that ought to never be changed -- not a tracker or budget limit). Can it be removed?

soroban-env-host/src/budget.rs
```diff
@@ -111,33 +33,75 @@ impl CostType {
...
-    /// model.
-    pub fn evaluate(&self, input: Option<u64>) -> u64 {
+impl HostCostModel for ContractCostParamEntry {
+    fn evaluate(&self, input: Option<u64>) -> Result<u64, HostError> {
+        if self.const_term < 0 || self.linear_term < 0 {
+            // TODO: consider more concrete error code "malformed xdr input"
+            return Err(ScUnknownErrorCode::Xdr.into());
+        }
+
+        let const_term = i64::from(self.const_term) as u64;
```
graydon, line 87: Did we figure out whether the cost model types can be stored as u32?

graydon (changes requested):
Small set of nits, nearly done!

soroban-env-host/src/budget.rs
jayz22: Oops sorry I forgot to clean it up. 
Deleted.

soroban-env-host/src/budget.rs
jayz22: The model `reset()` is being called by budget `reset_models()`, which is used in metering tests: create a Host, resets all models to zero, then set individual models of interest to the values we want, run the test and record the budget.

soroban-env-host/src/budget.rs
jayz22: Last time we discussed why some the other network configuration entries have a mixture of unsigned and signed numbers. 

The `i32` type for the cost formula terms were chosen because of generality consideration: 1. if for some rare reason we may encounter future cost types that decrease with the declared input.  2. we may add higher order (beyond linear) terms (via `ext`) and negative parameters may become necessary. Both of these cases are far-fetched for now but wouldn't hurt to preserve that possibility. 
`i32` instead of `i64` was chosen because it should have enough digits to represent the largest single cost. Although it isn't out of the question we might need a larger number (a single const term with i32::MAX representing 2B instructions or 2GB memory) so we may want to consider i64.

soroban-env-host/src/budget.rs
graydon: ok

soroban-env-host/src/budget.rs
graydon: *shrug* ok

soroban-env-host/src/budget.rs
graydon: thanks!

graydon (approved):

PR #773, opened 2023-04-22
Fix unsoundness in RawVal::get_tag

brson:
### What

Fix a soundness bug in RawVal::get_tag and add a test case that it stays fixed.

### Why

While working on https://github.com/stellar/rs-soroban-env/pull/767 I was flummoxed by an error in RawVal::get_tag.

This function does an unsafe cast from u8 to Tag. It does so by using three marker Tags: `SmallCodeUpperBound`, `ObjectCodeLowerBound`, and `ObjectCodeUpperBound`. It assumes that u8's between 0 and `SmallCodeUpperBound`, and between `ObjectCodeLowerBound` and `ObjectCodeUpperBound` are valid, and does a transmute if so, like this:

```rust
    const fn from_u8(tag: u8) -> Tag {
        const A: u8 = Tag::SmallCodeUpperBound as u8;
        const B: u8 = Tag::ObjectCodeLowerBound as u8;
        const C: u8 = Tag::ObjectCodeUpperBound as u8;
        if !((tag < A) || (B < tag && tag < C)) {
            return Tag::Bad;
        }
        unsafe { ::core::mem::transmute(tag) }
    }
```

The assumption is not true though and there are gaps in the Tag discriminants that result in creating Tags that don't exist, which leads to segfaults if those Tags are used.

The main thing this patch does is add a test case that ensures that the conversion from u8 to Tag is correct for all u8s. It does this by using the [int-enum](https://docs.rs/int-enum/latest/int_enum/trait.IntEnum.html) crate as a dev-dependency, which uses a proc macro to create the conversion code. A test case compares the hand-coded conversion to the macro-coded conversion. This ensures the hand-coded conversion is correct while not pulling the `int-enum` crate as a dependency for contracts.

It also:

- Extracts the actual conversion from `RawVal` to `Tag::from_u8`, putting the conversion closer to the converted type.
- Fixes the conversion for the two missing cases. The comments in this code indicate it is somewhat perf sensitive, but not critically so. The new cases will add a couple of branches or a jump table - I haven't checked the output.

I investigated changing the discriminants of Tag so there are no gaps but it isn't easy to change:

- The soroban-test-wasms must be rebuilt, and they currently do not build.
- Tags are part of the ABI anyway and someday will not be able to change ever.
- It seems best to future-proof this code anyway.

These tests will ensure that if somebody does remove the gaps in Tag prior to launch, adds more Tags, or otherwise changes how the conversion is done, then the conversion function will stay correct.

Note also that the new `Tag::from_u8` function is public, creating a way to test whether a u8 is a valid Tag. I need this ability in a test case of https://github.com/stellar/rs-soroban-env/pull/767.

### Known limitations

Would be nice to get rid of the two gaps in the Tag discriminants.


brson:
> Nice catch. I would like to actually eliminate the gaps though, as it is perf sensitive (and obviously a bit of a footgun in terms of logic). Why aren't the test wasms building?

User error - I didn't realize I had to pair a compatible local sdk with the test wasms build.

It's fixed now. The PR removes the discriminant gaps. It still extracts the conversion function to a `Tag::from_u8` and makes it pub, and adds a test using the int-enum dependency.

graydon (approved):
Great, thanks!

PR #780, opened 2023-04-28
Allow re-registering over a contract

leighmcculloch:
### What
Allow re-registering over an existing test contract.

### Why
Our registeration functions in the SDK allow for registering over the top of existing contracts. It's possible to register a test contract over a real contract, or a real contract over a real contract. It's an anomally in that context that you can't register a test contract over a test contract, and it prevents developers from doing some things in tests like replacing a contract with a different test contract. I'm running up against this limitation at the moment.

The only argument I can think of for why it could be beneficial to prevent re-registration is to catch human error when doing so is done by mistake, but given that I haven't seen or heard of folks accidentally doing that with any of the other ways to register contracts, I doubt it is a big issue. The inconsistency is a bigger issue.


graydon (approved):
LGTM, especially since it's just testutils -- if people don't like this we can easily change it _back_

PR #783, opened 2023-04-28
Allow getting top-level authorizations in enforced mode

leighmcculloch:
### What
Allow getting top-level authorizations in enforced mode by replacing the recording-mode only accessor with an accessor that returns all authenticated authorizations regardless of mode.

### Why
The recorded top-level authorizations accessor returns all recorded authorizations that were the first authorization in each call tree for each address. Recorded authorizations are authenticated authorizations. On that basis this functionality, accessing the list of authenticated authorizations, works for both enforced and recorded authorizations. This allows SDK developers to get access to top-level authorizations regardless of whether the environment is in recording or enforcing mode.

This is important for supporting introspecting authentication authorizations in tests where recording isn't in use, but some mocking is taking place and developers wish to also validate that their mock is in use.

I didn't add a test for this change because this function is a test utility, it doesn't appear to itself be tested, since it is something used in tests naturally.


dmkozh:
This looks good, but shall we wait until the release with this or will you have time to update the SDK/examples before it?

dmkozh (approved):

leighmcculloch:
I can update the examples. I'll need to update the examples due to the renames of terminology from recorded to mocked in any case and I can do this at the same time.

I think I need this to unify mock_all_auths with mock_auths so that having an accessor of auths that were seen works for both.

PR #787, opened 2023-05-01
Expose a list of all auths

leighmcculloch:
### What
Expose a list of all auths in addition to the top-level auths.

### Why
The concept of top-level auths is pretty confusing when learning about how auth works. We have a longer term goal of building a tree-like structure for folks to interact with auth, but that is deferred as it is a lot of work. It would be simpler conceptually for people if when they are thinking about the auths that have occurred if it was just all auths rather than needing to learn about top-level auths.

This change exposes all auths as a flat list. This has limited utility as a developer cannot reason about the relationship that one auth has with another without the tree structure. So it is imperfect, but hopefully a little less confusing.

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -1250,6 +1250,28 @@ impl Host {
             .get_authenticated_top_authorizations())
     }
 
+    // Returns the authorizations that have been authenticated for the last
+    // contract invocation.
+    //
+    // Authenticated means that either the authorization was authenticated using
+    // the actual authorization logic for that authorization in enforced mode,
+    // or that it was recorded in recording mode and authorization was assumed
+    // successful.
+    #[cfg(any(test, feature = "testutils"))]
+    pub fn get_authenticated_authorizations(
+        &self,
+    ) -> Result<Vec<(ScAddress, Hash, ScSymbol, ScVec)>, HostError> {
+        Ok(self
+            .0
+            .previous_authorization_manager
+            .borrow_mut()
+            .as_mut()
+            .ok_or_else(|| {
+                self.err_general("previous invocation is missing - no auth data to get")
+            })?
+            .get_authenticated_authorizations())
+    }
+
     #[cfg(any(test, feature = "testutils"))]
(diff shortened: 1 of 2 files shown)


soroban-env-host/src/host.rs
```diff
@@ -1250,6 +1250,28 @@ impl Host {
...
+    // contract invocation.
+    //
+    // Authenticated means that either the authorization was authenticated using
+    // the actual authorization logic for that authorization in enforced mode,
+    // or that it was recorded in recording mode and authorization was assumed
+    // successful.
+    #[cfg(any(test, feature = "testutils"))]
+    pub fn get_authenticated_authorizations(
+        &self,
+    ) -> Result<Vec<(ScAddress, Hash, ScSymbol, ScVec)>, HostError> {
```
dmkozh, line 1263: I'm afraid this might end up too misleading and confusing. Since the call stack is ignored here, there will be a lot of ambiguity in terms of what the actual auth trees were. Also this might create a wrong impression that e.g. in order to do something like authorized A->B call one needs to sign A and B and not a tree of A->B.

leighmcculloch: I agree it is imperfect, but it's a lot simpler to understand than top-level authorization, which I've discovered through writing tests I still don't completely understand. I think other developer will hit this problem too.

Being able to assert did an auth happen, without needing to assert the tree it occurred in, seems reasonable, and only possible if we provide visibility into all auths.

leighmcculloch: This change doesn't prevent us from improving this testing story to include the ability to assert on a tree.

dmkozh: Yeah, I guess we can update this once more when the tree support is there...

dmkozh (approved):

soroban-env-host/src/host.rs
leighmcculloch: Yup. I think if we see folks using this in a way that's bad, when we update with the tree, we can just replace it as a breaking change to the test api. If folks are using it okay, we can probably have side-by-side if we discover between now and then that there's value.

PR #789, opened 2023-05-01
Remove redundant `clone` calls

dmkozh:
### What

Remove redundant `clone` calls in auth manager

### Why

Optimization

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/auth.rs
+++ b/soroban-env-host/src/auth.rs
@@ -138,15 +138,15 @@ pub(crate) struct AuthorizedInvocation {
 
 impl AuthorizedInvocation {
     fn from_xdr(xdr_invocation: xdr::AuthorizedInvocation) -> Result<Self, HostError> {
-        let sub_invocations_xdr = xdr_invocation.sub_invocations.to_vec();
+        let sub_invocations_xdr = xdr_invocation.sub_invocations.into_vec();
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):

soroban-env-host/src/auth.rs
```diff
@@ -138,15 +138,15 @@ pub(crate) struct AuthorizedInvocation {
 
 impl AuthorizedInvocation {
     fn from_xdr(xdr_invocation: xdr::AuthorizedInvocation) -> Result<Self, HostError> {
-        let sub_invocations_xdr = xdr_invocation.sub_invocations.to_vec();
+        let sub_invocations_xdr = xdr_invocation.sub_invocations.into_vec();
         let sub_invocations = sub_invocations_xdr
             .into_iter()
             .map(|i| AuthorizedInvocation::from_xdr(i))
             .collect::<Result<Vec<_>, _>>()?;
```
jayz22, line 145: On this line, the `collect` call allocates memory  for the result container. But because here the structure is nested, you don't know the exact size to preallocate and and meter for it.

dmkozh: This PR is purely an optimization. In terms of metering, we do know the sizes beforehand and can come up with something.

jayz22 (approved):

PR #794, opened 2023-05-03
Bump dep on bytes-lit to 0.0.5 to trim dep on syn 1

graydon:
More syn-1 dependency trimming

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,14 +141,14 @@ checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
 name = "bytes-lit"
-version = "0.0.4"
+version = "0.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c181bca161a49348b90fa75f9a54fe11b7138098eed90d841a1055d574b4250"
+checksum = "0adabf37211a5276e46335feabcbb1530c95eb3fdf85f324c7db942770aa025d"
 dependencies = [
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -141,14 +141,14 @@ checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
...
-version = "0.0.4"
+version = "0.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c181bca161a49348b90fa75f9a54fe11b7138098eed90d841a1055d574b4250"
+checksum = "0adabf37211a5276e46335feabcbb1530c95eb3fdf85f324c7db942770aa025d"
 dependencies = [
  "num-bigint",
  "proc-macro2",
  "quote",
- "syn 1.0.104",
```
leighmcculloch, line 151: It doesn't look like syn 1 is being trimmed from the entire env. I the syn 1 block isn't being removed, something else is dependent on it.

leighmcculloch: It looks like there's quite a few dependencies that still use syn 1, such as cxxbridge, darling, etc.

leighmcculloch (approved):
Upgrading bytes-lit looks good, but doing so doesn't appear to remove our dependence on syn 1. See comments below.

Cargo.lock
graydon: Some, yeah. I've pushed another change that adds a mix of upgrades and exclusions (some were being excluded because of a misunderstanding of how skip-tree works).

PR #802, opened 2023-05-06
Implement PRNG subsystem

graydon:
This is a second attempt at implementing a minimal PRNG subsystem in the soroban host. It incorporates lessons learned discussing the first attempt in https://github.com/stellar/rs-soroban-env/pull/544, but starts fresh, both because that one had bitrotted a fair amount (there was a lot of churn to rebase around) and also I felt it'd be easier to review if we could have the two side by side rather than overwriting the first with a completely different implementation.

There's no "finalize" mechanism in here since, as @MonsieurNicolas pointed out, it's either useless (you can abort and retry at a transaction level) or, if made persistent across a txset, represents a DoS risk. Instead I added a comment explaining how to use a random-commit scheme to make random choices, as he described in the previous PR (I extended the example to include a ledger-number guard so that the commit-tx and subsequent action-tx could not themselves be embedded in a single composite multi-invocation tx that could be aborted if it didn't suit the attacker). 

Making good use of this random-commit approach means being able to reseed the PRNG in the second transaction (you can also ship your own guest PRNG or make only trivial decisions based on the committed randomness, but reseeding the host is more general) so I included a reseed operation. This also makes the PRNG building block more generally useful for more exotic use-cases so I figured we might as well.

The association between PRNGs and contexts is made more explicit here, and simplified. There's no hashtable keyed by contract IDs, each `Frame` (expanded to be a `Context`) just gets its own `Prng` seeded from the "base" `Prng` held in common in the `Host`. The only thing the "base" `Prng` is used for is seeding the `Context` `Prng`s, and its state can't otherwise be observed, reset by users or influenced. I _think_ this is as good as the previous scheme.

I added the vector-shuffle host function (because it's tricky and there's no cheap way to do it with a sequence of pointwise calls from the guest) and also changed the "next integer" function to take an inclusive range of `u64`s and do a proper rejection-sampling based uniform sample from the range (because it's too easy for users to do modulus themselves and accidentally get a biased sample).

For generating "random bytes" I decided to return a `BytesObject` rather than filling a portion of guest linear memory. You can copy a `BytesObject` to linear memory with a subsequent call, as you could have copied linear memory into a new `BytesObject` in a call subsequent to the linear-memory-fill of the previous approach. It's a bit of a judgment call to guess whether users will want one or the other, and my _guess_ is someone asking for a blob wants a blob to feed to something else (a seed, a hash function, etc.) more often than they want a guest-side array of bytes. I also think that for _large_ random blobs, or random blobs with runtime-variable size, this arrangement is cheaper since it doesn't involve the guest vm code zero-filling a bunch of linear memory before asking the host to fill it with bytes.


soroban-env-host/src/host/prng.rs
```diff
@@ -0,0 +1,151 @@
...
+///     financial integrity. In other words: reusing it here doesn't make
+///     anything worse, we're already obliged to make transaction apply-order
+///     hard to guess or control.
+///
+///   - Each frame (which is to say: each contract invocation or sub-invocation)
+///     will get a new PRNG instance separately seeded from the host's "base"
+///     PRNG, and guest code can only access the frame's PRNG, not the "base"
+///     PRNG or those of any other frame. This doesn't eliminate _all_ attack
+///     vectors or mechanisms for misuse, but it's the best we can give the user
+///     for buiding on. In particular it means that a "random" contract will not
```
Shaptic, line 40: nit: typo
```suggestion
///     for building on. In particular, it means that a "random" contract will not
```

soroban-env-host/src/host/prng.rs
```diff
@@ -0,0 +1,151 @@
...
+        let mut v2 = v.metered_clone(budget)?;
+        // We charge for both the PRNG draws and the swaps here (as "memcpys").
+        self.charge_prng_bytes(budget, 16u64.saturating_mul(v.len() as u64))?;
+        budget.charge(ContractCostType::HostMemCpy, Some(v.len() as u64))?;
+        v2.as_mut_slice().shuffle(&mut self.0);
+        Ok(v2)
+    }
+
+    pub(crate) fn bytes_new(&mut self, size: u32, budget: &Budget) -> Result<ScBytes, HostError> {
+        budget.charge(ContractCostType::HostMemAlloc, Some(size as u64))?;
```
Shaptic, line 137: passing thought: it'd be really cool if the **only** way to get host memory was via charging, rather than relying on ourselves to ensure we always charge the budget when appropriate. but maybe that's too restrictive? dunno, just a drive-by review musing

soroban-env-host/src/host/prng.rs
```diff
@@ -0,0 +1,151 @@
...
+use rand::{distributions::Uniform, prelude::Distribution, seq::SliceRandom};
+use rand_chacha::{
+    rand_core::{RngCore, SeedableRng},
+    ChaCha20Rng,
+};
+use std::ops::RangeInclusive;
+
+/// PRNG subsystem in the host, which provides best-effort pseudo-randomness to
+/// guest contracts using a combination of features that guests cannot easily
+/// implement themselves:
```
Shaptic, line 17: just to emphasize: we aren't (shouldn't!) recommend this for cryptographic usage in any way, shape, or form, right?

Shaptic (commented):
quick drive-by review due to passive interest in how this was going to be implemented! feel free to ignore or respond to any comments :v:

soroban-env-host/src/host/prng.rs
graydon: It depends what you mean by "cryptographic uses". ChaCha20 is a CSPRNG (in terms of its mixing function) and I don't think we need to shy away from committing to providing one. On the other hand it has to operate (a) in consensus among all peers and (b) using strictly public input data, so the main thing people _usually_ want out of a CSPRNG and usually mean when they say "cryptographic usage" -- non-public, secret state that's usable to derive secret keys or unpredictable nonces from, that are then not-shared -- is by definition not available. Every other node in the system (including watchers) can immediately calculate / observe the exact internal state of the ChaCha20 instance by running their own copy on the same input that everyone has, and everything you derive from the state during execution is by necessity public and identical to the copies everyone else calculates in consensus.

The point of this is to allow contracts to "make unpredictable choices" in their actions that are as _close_ to unpredictable-in-advance as we can get them, given this by-defnition public-and-predictable context. An example might be a lottery contract that's going to pay out to a single participant. In theory this should be done by having randomness provided by an off-chain oracle (or set of oracles) that are trusted not to be participating in the lottery, and that first submit commitments and then later submit signed preimages. But that's a lot of work to set up and still relies on a trusted third party, and so people often (especially in the early days of a given contract)  don't do it. Instead they often build "bad randomness sources" like the ledger sequence number, which is _very_ predictable in advance. Someone wanting to win a lottery would be able to guess the ledger sequence number of the payout ledger in advance (and/or as soon as they see that a ledger is closing with a "do the payout" transaction), and could use that knowledge to situate themselves as the winner.

In the design here (and in the corresponding patch we'll land on stellar-core) the seed is intended to be derived from the txset that won nomination, and the tx's hash-order numerical position within it, which is both hard to predict and by definition changes if you observe it and then inject your own transaction. In general it's just hard to influence unless the attacker is themselves a trusted validator, since nominations from anyone else are ignored. And the network generally already affords validators this level of trust (they can already reorder or drop transactions during consensus -- or inject their own -- when picking a txset to nominate, so we're not really increasing the MEV attack surface here).

soroban-env-host/src/host/prng.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl Prng {
+    fn charge_prng_bytes(&self, budget: &Budget, count: u64) -> Result<(), HostError> {
+        // TODO: add a ContractCostType for drawing PRNG bytes
+        Ok(())
+    }
+
+    pub(crate) fn reseed(&mut self, seed: Seed) {
+        self.0 = ChaCha20Rng::from_seed(seed);
```
MonsieurNicolas, line 99: Do we need to do anything wrt requirements wrt `from_seed`  (I think not because we're using a cryptographic hash function, but checking)

```
PRNG implementations are allowed to assume that bits in the seed are well distributed. That means usually that the number of one and zero bits are roughly equal, and values like 0, 1 and (size - 1) are unlikely.
```

soroban-env-host/src/host/prng.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.charge_prng_bytes(budget, size as u64)?;
+        let mut vec = vec![0u8; size as usize];
+        self.0.fill_bytes(&mut vec);
+        Ok(ScBytes::try_from(vec)?)
+    }
+
+    pub(crate) fn sub_prng(&mut self, budget: &Budget) -> Result<Prng, HostError> {
+        let mut new_seed: Seed = [0; SEED_BYTES];
+        self.charge_prng_bytes(budget, SEED_BYTES as u64)?;
+        self.0.fill_bytes(&mut new_seed);
```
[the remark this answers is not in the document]
MonsieurNicolas: I think that as long as the seed used to initialize `base_prng` is secret, there is no way to predict the output.
As a consequence, we have to be really careful to never expose information that could allow somebody to brute force that seed -- it's a bit of a footgun, so we should add comments next to `LedgerInfo` that points to `base_prng`.

MonsieurNicolas (changes requested):
looks pretty good to me. Only a couple questions

soroban-env-host/src/host/prng.rs
graydon: From what I can find on the internet, proper function of ChaCha20 is not sensitive to bit-bias in its seed -- the seed just has to be (obviously) secret and not predictable. ChaCha20 is a minor variant of Salsa20, which was built to demonstrate that one can construct a viable stream cipher (and everything else cryptographic) out of a cryptographic hash, and as such its design is just a CHF of simple set of state-words: some fixed constants, the "key" ("seed" in our case) words, a stream identifier number, and a nonce that increments on each byte emitted. Changing one bit in any of these should do the same thing it does in any CHF: unpredictably flip half the bits. The stream and nonce words are actually _assumed_ to be highly biased numbers (mostly zeroes) and the mixing function is designed to integrate bits uniformly from all words in its input, as far as I can tell. See [section 4 of the design docs](https://cr.yp.to/snuffle/design.pdf) for some elaboration.

That said, I am not a cryptographer and I would be happy to add a documentation qualifier here that advises users to only ever use seeds that they got from the CSPRNG in the first place, since other seeds are likely _predictable_, which defeats the whole point. But I don't think we need to do anything extra here.

And yes, the seed _we_ supply (the one driving the outer `base_prng`) come from a CHF, so they should also satisfy this criterion already.

soroban-env-host/src/host/frame.rs
```diff
@@ -100,6 +109,7 @@ impl Host {
     /// operation fails, it can be used to roll the [`Host`] back to the state
     /// it had before its associated [`Frame`] was pushed.
     pub(super) fn push_frame(&self, frame: Frame) -> Result<RollbackPoint, HostError> {
+        let prng = self.0.base_prng.borrow_mut().sub_prng(self.as_budget())?;
```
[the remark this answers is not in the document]
graydon: I suppose. I don't feel strongly, figured fewer options would be better. I can make it lazy though.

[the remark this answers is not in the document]
MonsieurNicolas: yes that's what I mean, make it lazy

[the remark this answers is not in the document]
graydon: Done

graydon:
@MonsieurNicolas I've addressed review comments, rebased, and added unit tests (which caught some subtle initialization/borrowing issues, variants of https://github.com/stellar/rs-soroban-env/issues/817)

MonsieurNicolas (approved):
Cool - LGTM

PR #803, opened 2023-05-08
Add conversion from Status to ScStatus where the type is HostAuthError

brson:
### What

Add missing case in `TryFrom<Status> for ScStatus` where the type is `HostAuthError`. Use an exhaustive match to prevent similar bugs in future.

### Why

ScStatus values with the HostAuthError type cannot be created from their host equivalents.

### Known limitations

No tests, but I have proptests covering this that I will submit to the SDK in the near future.


dmkozh (approved):

brson:
This is probably obsoleted by https://github.com/stellar/rs-soroban-env/pull/803. Closing.

Issue #772, opened 2023-04-22
Debug events are ambiguous as to which contract generated them


leighmcculloch:
Debug events when output do not include a clear indication of which contract ID generated the event. This can get ambiguous with even small traces involving only a couple contracts. For example:

```
HostError
Value: Status(ContractError(5))

Debug events (newest first):
   0: "Debug escalating error '' to panic"
   1: "Debug contract call invocation resulted in error Status(ContractError(5))"
   2: "Debug caught panic from contract function 'Symbol(add)', propagating escalated error 'Status(ContractError(5))'"
   3: "Debug escalating error '' to panic"
   4: "Debug incompatible signature format"
```

(This log is from #771.)

Could we include contract IDs, or truncated contract IDs in debug events?

cc @graydon

graydon:
Agreed. This goes together with https://github.com/stellar/rs-soroban-env/issues/681 IMO, we should just have objects kicking around for any contract ID that we can tag anything with that is contract-ID-specific.

(Possibly the object handles for contract IDs can/should be related to the ticket numbers in https://github.com/stellar/rs-soroban-env/pull/132 -- if we're never rolling back object handles it might make sense to assign ticket numbers as "the next free object handle" when pushing a frame)

graydon:
implemented in https://github.com/stellar/stellar-xdr/pull/92 -- should land soon

graydon:
Done in https://github.com/stellar/rs-soroban-env/pull/806

Issue #689, opened 2023-02-16
A bunch of simple renamings


graydon:
This is a tracking issue for a bunch of simple renamings to do.

```[tasklist]
### Tasks
- [x] `Status` => `Error`
- [ ] `env_val.rs` => merge content into `convert.rs`
- [ ] `RawVal` => `Val`
- [ ] `RawValConvertable` => `ValConvert` (and possibly make crate-private)
-  [ ] `TryFromVal` `TryIntoVal` => `EnvConvertFrom` / `EnvConvertTo`
- [ ] `Convert` => `EnvConvert`
```

The last three I'm not as certain about, but I think it'd help (a) emphasize the interrelated-ness of the traits and (b) emphasize that what's called `Convert` today is always actually implemented by an `Env` and (c) remove one weird part of `TryFromVal` which is that it's often not converting _from_ a `RawVal` at all, it might even be converting _to_ a `RawVal` (eg. `impl TryFromVal<ScVal> for RawVal` which would I think read better `impl EnvConvertFrom<ScVal> for Val`

graydon:
@leighmcculloch @paulbellamy @sisuresh @jayz22 @dmkozh during value overhaul we postponed a few renamings as "too systematic to bother with at the time and easy to do later" -- I'd be curious to hear your preferences on these names

dmkozh:
I really like the idea of separating the `try_from_val` into 2 cases - the current state doesn't make much sense to me when we convert from env values to, say, `ScVal` (which is outside of env). As for the specific naming, I don't have a strong opinion. Given that `RawVal` is currently usable and useful in the SDK, maybe something like `AnyVal` could be a better fit. But `Val` seems ok to me as well.

jayz22:
`Status` => `Error`, what do we do with `Status::OK` and how do we plan to handle status-returning host functions where the success or failure status is indicated by `Status`?  Will such functions just be `void`-returning where the status will be captured by the `HostError`?
`RawVal` => `Val`, I think this makes sense which resolves the confusion of `RawVal::from_u32(u: u32) -> U32Val`.
`TryFromVal`: One other choice of name proposed before was `TryFromViaEnv` which I also liked.

graydon:
@dmkozh can you say a little more about "separating into two cases"? If you mean conversions that use the env vs. those that don't, the latter are already supported (where possible) through From/Into or TryFrom/TryInto.

graydon:
[...]

Issue #687, opened 2023-02-14
Detect/avoid double-faulting in the host runtime


jayz22:
We've recently encountered a case where host function faults, tries to generate a `DebugError` with the error status. While in the middle of creating a `DebugEvent`, it runs out of budget and faults for the second time with an "out of budget" status. The original status was overridden. 

This is a potential footgun. We need to avoid such double/multiple-faulting cases, or at the very least, be able to detect it. One potential solution @graydon suggested was to add a static counter to track the number of times the host have faulted, and asserting that number cannot exceed 1. 

_How to reproduce it:_
Revert changes mentioned in [this comment](https://github.com/stellar/rs-soroban-env/pull/662/files#r1096318055), and run the `metered_xdr_out_of_budget` test under `budget_metering`.

graydon:
Done in https://github.com/stellar/rs-soroban-env/pull/806

PR #806, opened 2023-05-12
Err reform

graydon:
This is a WIP branch that's quite repetitive but ultimately fairly simple, that performs a variety of fixes related to events and errors. I _think_ it's all stuff we've already agreed "needs to happen" and I figured I'd just try to get it all done this week while everyone else was away from keyboard.

Hopefully I'll be done soon but it's going to probably take another day or two of just grinding through errors. Here's basically what it covers:

1. Renames `Status` to `Error` (the most-agreeable part of #689)
2. Reassigns all error codes to a smaller, consolidated set, fixing #225 (i.e. paired with change https://github.com/stellar/stellar-xdr/pull/92 where the set is explained, and a corresponding WIP rs-stellar-xdr change)
3. Deletes `DebugEvent` entirely, moving all diagnostics to structured-event form, answering at least _some_ of #731
4. As a result, assigns a causal contract ID to every debug event, fixing #772 (and #438 which is I think an earlier dupe of the same issue)
5. Also as a result, externalizes full objects when diagnostic mode is turned on, causing them to render fully, fixing #442
6. Fixes the double-faulting issue in #687
7. Adds a bunch more diagnostic context in cases where we're generating errors (needed since we're simplifying the codes), fixing #673

I thought I'd post it for feedback early, in case anyone _really_ hates the direction. I'm pretty keen on how it's turning out so far -- it seems like it really addresses a lot of the shortcomings we've identified so far and most of the churn is just obvious cleanup after the main structural changes -- but I'm interested to know if it looks to anyone like it's going the wrong direction.


soroban-env-host/src/auth.rs
```diff
@@ -736,7 +758,12 @@ impl AuthorizationTracker {
                 Some(curr_invocation.sub_invocations.len() - 1);
         } else {
             // This would be a bug
-            return Err(host.err_general("unexpected missing authorized invocation"));
+            return Err(host.err(
+                ScErrorType::Auth,
+                ScErrorCode::MissingValue,
```
dmkozh, line 763: nit: this should be `InternalError` probably (given that this is an 'implementation bug' state).

The same actually applies to all the errors below besides 'nonce is incorrect' one. Basically anything starting with `unexpected` is an internal error and shouldn't ever happen (unless we have bugs).

soroban-env-host/src/events/mod.rs
```diff
@@ -23,12 +31,110 @@ pub struct HostEvent {
...
+            u256_from_pieces(v.hi_hi, v.hi_lo, v.lo_hi, v.lo_lo)
+        ),
+        ScVal::I256(v) => write!(
+            f,
+            "{}",
+            i256_from_pieces(v.hi_hi, v.hi_lo, v.lo_hi, v.lo_lo)
+        ),
+        ScVal::Bytes(v) => write!(f, "Bytes({})", v.0),
+        ScVal::String(v) => write!(f, "\"{}\"", v.0),
+        ScVal::Symbol(v) => write!(f, "{}", v.0),
```
dmkozh, line 72: nit: Shall we do `Symbol({})` format for clarity? Othewise this seems a bit too easy to confuse with string

soroban-env-host/src/host/metered_map.rs
```diff
@@ -123,8 +120,8 @@ where
                 ctx: Default::default(),
             })
         } else {
-            // TODO use a better error code for "unbounded input itertors"
-            Err(ScHostFnErrorCode::UnknownError.into())
+            // This is a logic error, we shoudl never get here.
```
dmkozh, line 123: s/shoudl/should

soroban-env-host/src/host/metered_vector.rs
```diff
@@ -82,27 +87,22 @@ where
             vec.charge_deep_clone(budget)?;
             Self::from_vec(vec)
         } else {
-            // TODO use a better error code for "unbounded input iterators"
-            Err(ScHostFnErrorCode::UnknownError.into())
+            // This is a logic error, we shoudl never get here.
```
dmkozh, line 90: s/shoudl/should

soroban-env-host/src/events/mod.rs
graydon: I tried putting some kind of marker on it and it just reads badly. strings are quoted, so in practice it seems not too bad (and this is just for diagnostic display anyway, when the user gets an error dumped out on the console)

soroban-env-common/src/error.rs
```diff
@@ -0,0 +1,217 @@
...
+#[derive(Copy, Clone)]
+pub struct Error(RawVal);
+
+impl_wrapper_tag_based_rawvalconvertible!(Error);
+impl_wrapper_tag_based_constructors!(Error);
+impl_wrapper_as_and_to_rawval!(Error);
+impl_wrapper_wasmi_conversions!(Error);
+
+impl Error {
+    pub const UNKNOWN: Error =
```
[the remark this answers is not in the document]
graydon: Between unknown and internal? Fair point. I'll see how it looks if I remove unknown entirely.

soroban-env-host/src/auth.rs
graydon: ok I'll fix these

soroban-env-common/src/error.rs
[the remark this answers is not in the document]
graydon: Yeah, this works fine with just "internal", changing.

soroban-env-host/src/auth.rs
graydon: fixed (all except the ones about nonces in the ledger having the wrong type -- that seems like a user error)

soroban-env-host/src/host/metered_vector.rs
graydon: fixed

soroban-env-host/src/host/metered_map.rs
graydon: fixed

soroban-env-host/src/auth.rs
dmkozh: These would actually be internal as well as the host manages nonces exclusively currently. But nonces also will hopefully change the implementation after temp storage introduction, so also feel free to skip these for now, it doesn't really matter that much.

soroban-env-common/src/error.rs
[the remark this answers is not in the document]
dmkozh: Nice, thanks!

soroban-env-common/src/env.rs
```diff
@@ -137,65 +137,10 @@ pub trait EnvBase: Sized + Clone {
...
-    fn log_static_fmt_general(
-        &self,
-        fmt: &'static str,
-        vals: &[RawVal],
-        strs: &[&'static str],
-    ) -> Result<(), Self::Error>;
+    /// Log a string and set of values as a diagnostic event, if diagnostic
+    /// events are enabled. When running on host, logs directly; when running on
+    /// guest, redirects through log_from_linear_memory.
+    fn log_from_slice(&self, msg: &str, vals: &[RawVal]) -> Result<Void, Self::Error>;
```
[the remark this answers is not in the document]
graydon: Yeah, the thing we were doing before where guest.rs turned the log calls into no-ops when compiling for wasm we're no longer doing. I removed basically all the special static cases around diagnostic logging -- even static-string optimization in events -- and just went with a 100% dynamic-strings and dynamic-vecs solution. It's costly memory-wise but you don't get to incur those costs in any context where it'll do harm: we'll prohibit turning on dynamic logging on validators entirely, and set it default-off on all production nodes. Users can turn it on on production nodes tx-by-tx if they want, but not validators. And any local case, obviously -- in the SDK change I figured you'd just turn it on by default, but we can discuss that when reviewing the SDK change (I expect you'll have opinions there!)

PR #819, opened 2023-05-25
Remove incorrect debug_assert from ed25519_pub_key_from_bytes

brson:
### What

Remove an assertion that does not hold.

`Host::ed25519_pub_key_from_bytes` contains an assertion about the length of the provided pubkey:

```
    pub(crate) fn ed25519_pub_key_from_bytes(&self, bytes: &[u8]) -> Result<PublicKey, HostError> {
        debug_assert!(bytes.len() == 32);
        self.charge_budget(ContractCostType::ComputeEd25519PubKey, None)?;
        PublicKey::from_bytes(bytes).map_err(|_| {
            err!(
                self,
                (ScErrorType::Crypto, ScErrorCode::InvalidInput),
                "invalid ed25519 public key",
                bytes
            )
        })
    }

    pub fn ed25519_pub_key_from_bytesobj_input(
        &self,
        k: BytesObject,
    ) -> Result<PublicKey, HostError> {
        self.visit_obj(k, |bytes: &ScBytes| {
            self.ed25519_pub_key_from_bytes(bytes.as_slice())
        })
    }
```

### Why

A guest can trigger this assertion in the host by providing a BytesObject with incorrect length to the `Host::verify_sig_ed25519` syscall:

```
    fn verify_sig_ed25519(
        &self,
        _vmcaller: &mut VmCaller<Host>,
        k: BytesObject,
        x: BytesObject,
        s: BytesObject,
        let public_key = self.ed25519_pub_key_from_bytesobj_input(k)?;
        let sig = self.signature_from_bytesobj_input("sig", s)?;
        let res = self.visit_obj(x, |payload: &ScBytes| {
            self.verify_sig_ed25519_internal(payload.as_slice(), &public_key, &sig)
        });
        Ok(res?.into())
    }
```

By removing the assertion, the function will return an error that can be correctly propagated instead of panicking.

### Known limitations

No tests. I attempted to add a test to soroban-test-wasms but could not get the wasms to build and pass existing tests - there seem to be some transitional changes to AuthorizationContext and BytesN/Address that need to be implemented, and which I could not figure out.


leighmcculloch (approved):

graydon (approved):

graydon:
Thanks, I generally don't feel great about debug assertions in the code at all .. I see we have a few, but I think they might be worth removing. Especially in cases where it's possible for user input to violate them, "what do we actually want to do here if this is wrong" should have an answer better than "oh well in a non-debug build we just ignore it". If it's an important invariant we should probably return an error!

Issue #821, opened 2023-05-30
Improve `VmInstantiation` calibration


jayz22:
This comment https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/budget.rs#L650C1-L652 needs to be eliminated. 
If any calibration result cannot be taken as is, need to add proper heuristics to override it.

Issue #824, opened 2023-05-31
Consider using fixed-point integer for cost model parameters


jayz22:
The metering cost model is a const or linear model w.r.t to some input. I.e  `y = a + b * x`.
Because `a` and `b` are currently [defined as integers](https://github.com/stellar/stellar-xdr/blob/077fe707303edb20358c3d6484ccfe2fe587ad4e/Stellar-contract-config-setting.x#L131-L136), a great deal of model precision is lost in the rounding. Consider the following calibrated output of a cost model:

`cpu model params: FPCostModel { const_param: 40.798164912282914, lin_param: 0.5000807017543858, r_squared: 0.9999351263662553 }`

This model has highly linear characteristic (`R^2` is almost 1), however, due to integer rounding, the linear term is either rounded up to 1 or rounded down to 0, both sub-optimal. In the latter case, it would render the linear relation a constant, which is plain wrong.

It would be better if the model params can be represented as fixed point integers and model calculation is performed on fixed-point integers and rounding only happen at the very end. 

This would require introducing a fixed-point integer library in the host, although It will be internal use only. The `ContractCostParamEntry` xdr need to be updated to represent fixed-point integer of its const and linear terms.

jayz22:
Instead of using fixed-point integer, we've added a simple scaling to the linear coefficient which is much simpler and works reasonably well. See https://github.com/stellar/rs-soroban-env/pull/982 for details.

Issue #208, opened 2022-07-13
Cost models might be better with more than one input dimension


graydon:
Some cost models make more sense when viewed in terms of more than one independent variable (such as different sizes of separate inputs to a function that uses them both in a subtle way). Currently we funnel all inputs to cost models into a single input variable, and in many cases that might work (sum all input sizes, take the max of all input sizes, multiply one input size by another, etc.) but in other cases it might make more sense to treat the cost model as a function of separate input dimensions.

jayz22:
Closing this as we have decided that all cost components should only dependent on single input.

PR #825, opened 2023-05-31
Improve `VmInstantiation` calibration

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/821:
- Use synthesized Wasm at different sizes to fit the linear model parameter for `VmInstantiation`.
- Rerun calibration and update the model parameters.  See full [output](https://github.com/stellar/rs-soroban-env/files/11619147/output15.txt)

The main diff is in the `VmInstantiation` costs
```diff
// cpu cost
ContractCostType::VmInstantiation => {
-    cpu.const_term = 1_000_000;
-    cpu.linear_term = 0;
+    cpu.const_term = 736049;
+    cpu.linear_term = 684;
}

// mem cost
  ContractCostType::VmInstantiation => {
-      mem.const_term = 1100000;
-      mem.linear_term = 0;
+     mem.const_term = 107854;
+     mem.linear_term = 49;
  }
```

### Why

`VmInstantiation` cost depends linearly w.r.t contract size, and has been calibrated as a linear model.
Previously we were using the `soroban-test-wasm` contracts which were few, as a result, the linear parameter was poorly calibrated (R^2 around 0.7). As a result we manually overrode the calibrated parameters with a constant ones.

This PR uses synthesized Wasm so that the size of the contracts can be varied and controlled. The resulting R^2 score is > 0.99. 

### Known limitations
During calibration, it was realized the precision of metering is compromised by the integer representation of the model params (especially the linear one that are close to 0). They should probably be represented as fixed-point integers. Created this followup issue: https://github.com/stellar/rs-soroban-env/issues/824


anupsdf:
[Plotting a chart](https://docs.google.com/spreadsheets/d/1foFMWO9crArNfmdMiS5SRJpB-RiYzLHcQmCS01BFsmQ/edit?usp=sharing) from my local run for VMInstantiation shows that CPU instruction is linear but memory seems like a constant.  
No strong objection though if you feel it should be constant for both.

graydon:
Hm yeah I thought this was fairly .. linear actually! I mean, concretely it definitely _does_ have to do work that's at least linear in contract size (in the sense that it has to parse the wasm). Can you not use the synth-wasm subcrate to get a clearly varying set of samples?

jayz22:
Thanks for the plot @anupsdf.
@graydon yeah I can try it with synth-wasm. My guess is it is only weakly linear to the size of the contract function if since it has to parse all the wasm sections, validate them, and link the host functions etc (thus the the high floor of instructions). But it's worth exploring.

jayz22:
Updated the PR and the comments. Ready for review again.

mootz12:
Hello!

Wanted to chime in as I've just started benchmarking the Blend protocol to start looking for optimizations.

I have some concerns about how much weight the `VmInitialization` cost function has on the total cost of both the CpuInst and MemBytes. Currently, this is taking up ~70-80% of total CPU instructions and ~99% of Mem Bytes, leaving the only optimizations available as reducing the number of cross contract calls.

The main culprit is a custom token (tracks collateral or liabilities for an underlying token, protocol calls `balance` on these a lot) that has a fairly small WASM size (8500 bytes).

From what I can quickly grasp from this PR - it looks like I will expect costs to increase substantially in this department, from `1m` CpuInst per cross contract call to `1.78m` (8500 * 119 + 772061).  This is not including the significantly increased access cost of large contracts, either (8400 bytes is the breakeven point to the old fixed cost).

Happy to share more information if it would be useful.

graydon:
@jayz22 The second, larger bound should be used -- we're aiming for an upper bound. Users can submit a malicious contract.

@mootz12 Yes, cross-contract calls are for the time being the largest cost centre, by a fairly wide margin, and they're likely to always be so. We're likely to bring down the magnitude of difference in the future by caching a certain amount of material in the VM, making instantiation avoid re-parsing wasms from one call to the next, but we haven't put any time into this yet, and will only be able to do so much. The rest of the system is fairly efficient, so by comparison instantiation of each VM is high-cost.

jayz22:
@graydon I've updated it to use the second, larger bound. 

@mootz12 Thanks for your feedback. It is unfortunately VM instantiation is as high as it is now, due to the work necessary to set up the VM and parsing the wasm contracts. The metered cost parameters are just reflection of the calibrated costs (and we have to take the upper bound to prevent malicious contracts). We may have improvements and optimizations for it in the future (created https://github.com/stellar/rs-soroban-env/issues/827 to track it).

jayz22:
> Do we have any understanding on what actually drives these numbers up so much and why do we have so much variance? E.g. in Anup's data we have linear param at around 230 and in 2 Jay's experiments we have values from 100 to 600. Either way the number is much higher than the cost of interpreting a Wasm instruction, which seems really weird to me. Does VM do some heavy pre-processing? Is something off about our measurement methodology?

Please see the comment above https://github.com/stellar/rs-soroban-env/pull/825#issuecomment-1570631266
These are two different setups.

dmkozh:
> Please see the comment above [#825 (comment)](https://github.com/stellar/rs-soroban-env/pull/825#issuecomment-1570631266) These are two different setups.

Sure, I understand that these are two or even three different setups. What I don't understand is why do we observe such a drastic difference and whether we are using the proper value as the input size (it seems to me like we are not).

jayz22:
>  Sure, I understand that these are two or even three different setups. What I don't understand is why do we observe such a drastic difference and whether we are using the proper value as the input size (it seems to me like we are not).

Well, in this case (and most cases) there is no perfect single input to capture all the degree of variations.  Instantiating an VM requires much work including parsing and validating the Wasm file, initializing linear memory, linking host functions etc. 
Using Wasm bytes size is the best we can do in terms of capturing all the variations in one degree-of-freedom. But we have to (or try our best to) set it up such that the resulting one-dimensional function represents an upper bound across all other factors. 
In this case, one of the simplest case is having one exported contract function containing all the bytes, the it will just have to parse and link that one function . That was the first experiment (` lin_param: 117.59111280561848`). On the other end, the bytes can be made up of many small functions, each of which consist of a simple argument that is trivial to parse, but the overall cost of parsing that file of the same length as the first one will be much larger (`lin_param: 687.9128326191914`) per byte. We have to pick the second case for meter charging. 
The other stages of VM instantiation (initing linear memory, host functions) should be more or less constant, thus explains the high (but similar for both) floor (~700k).

dmkozh:
> But we have to (or try our best to) set it up such that the resulting one-dimensional function represents an upper bound across all other factors.

That's not necessarily the only answer - we have other options, such as changing what the input value is or breaking down the operation into multiple linear components dependent on different values.

> On the other end, the bytes can be made up of many small functions, each of which consist of a simple argument that is trivial to parse, but the overall cost of parsing that file of the same length as the first one will be much larger (lin_param: 687.9128326191914) per byte. We have to pick the second case for meter charging.

If we believe that the number of functions is what drives the cost, then why don't we use it as the input parameter to metering? This of course creates a 'loop' for the current model (because you need to load the contract to count the functions), but there are ways around that, such as:

- caching the number of functions as part of wasm upload operation
- break the vm instantiation down in two stages (if feasible)

> We have to pick the second case for meter charging.

I'm not sure that's the best solution:

- on one hand, we have an unknown number of contracts (likely majority?) overpaying for instructions at 3-4x rate
- on the other hand, since we still don't have a good understanding on what _actually_ is the cost function, we can't quite claim that this is the worst case

One potential way to exercise this would be to create two sets of benchmarks:

- Contracts with a single function and growing size (e.g. calling into a bunch of host fns)
- Contracts with a growing number of no-op functions 

I think both can be generated with a bit of proc macros.

Another thing that potential might matter is the number of _imported_ host functions. I'm not sure if we're covering that now - it might or might not have an impact (again, even the 687 coefficient might not necessarily be enough). Would be nice to also cover contracts that import as many host fns as possible.

jayz22:
Multi-dimensional inputs were considered (see https://github.com/stellar/rs-soroban-env/issues/208) but we decided it was not worth the complexity at the time. And we will be facing the same challenges of picking the number of degrees of freedom, as well as deciding which inputs are worthy. Not to mention the calibration across multi-dimensional inputs will be challenging in terms of computation and accuracy. In this case one can simply argue the contract can do other wild things to make it expensive to parse but we need to make an assumption and draw a line, and I think the many-simple-local-functions approach is a decent one. 

> we can't quite claim that this is the worst case

If we think there is another clear-cut worse case we would consider using that.

> One potential way to exercise this would be to create two sets of benchmarks

This might help but again not sure this is worth the extra complexity and not sure this solves the problem, it doesn't eliminate the heuristic of deciding how many local functions is "too many" that needs to fall into one bucket vs another. 

> Another thing that potential might matter is the number of imported host functions.

All of the host functions are linked into Wasmi for any contract. That's is necessary for Wasmi to resolve a particular host function for a contract to call. 

As @graydon mentioned above and in the thread. We have other more systematic ways to make the VM instantiation cheaper, e.g. sharing modules, caching parsed contracts which may reduce the cost fundamentally. But in any case the cost formula will have to exhibit the worst case estimate in some way.

dmkozh:
> Multi-dimensional inputs were considered (see https://github.com/stellar/rs-soroban-env/issues/208) but we decided it was not worth the complexity at the time.

I'm not necessarily suggesting to use multi-dimensional inputs. We could just have two linear functions (in case if the runtime is a linear function of the inputs of course), e.g. 'vm parse (contract size)', 'vm instantiate (function count)' or something like that. This may or may not be the right model, but we won't know if we don't try.

> This might help but again not sure this is worth the extra complexity and not sure this solves the problem, it doesn't eliminate the heuristic of deciding how many local functions is "too many" that needs to fall into one bucket vs another.

It does have a potential to solve up 5x discrepancy between the cases.

> If we think there is another clear-cut worse case we would consider using that.

What I'm proposing is to do more rigorous benchmarking to figure that out.

> All of the host functions are linked into Wasmi for any contract. That's is necessary for Wasmi to resolve a particular host function for a contract to call.

That's not exactly true; the contract itself has to declare the host fns it's using. I don't know if this does or doesn't have impact on the performance, which is why that's just one more idea of what could be measured.

>  We're not going to capture all of them and we've no strong reason to believe that "large number of empty functions" is the worst case use of bytes in terms of incurring wasmi costs. Merely that it's worse than other better cases.

Sure, so shouldn't we think more as to what might be the worse case? But in any case, if there is a straightforward parameter of the model (like the function number), shouldn't we just use it? If linear combination of contract size and function number gives us a model that has, say just <2x fluctuation between best and worst cases we find (instead of 3-5x), then I'd say that would be huge win for the ecosystem.

jayz22:
I've done a few experiments and did not find a way to further improve the current metering model. 
The most contract-size dependent component in Wasmi instantiation is `Module::new`, which requires parsing the entire Wasm file, section by section and all the bytes in it. There is no way around that.
The number of the bytes is the best proxy to the single input for a linear model for metering. 
The only way to improve the metering model, without significantly altering the current framework and assumptions (i.e. no multi-dimensional inputs), is to do https://github.com/stellar/rs-soroban-env/issues/838. This requires breaking down `Module::new` into sections and meter them individually. 
I also tried component recycling in https://github.com/stellar/rs-soroban-env/issues/827 and didn't find any quick wins. 

Given our current constrains on time and resources, we should proceed with merging this PR, and adjust the resource limits at the network config level.

soroban-env-host/src/host/declared_size.rs
```diff
@@ -259,6 +264,9 @@ mod test {
         expect!["8"].assert_eq(size_of::<SymbolSmallIter>().to_string().as_str());
         expect!["32"].assert_eq(size_of::<U256>().to_string().as_str());
         expect!["32"].assert_eq(size_of::<I256>().to_string().as_str());
+        #[cfg(target_arch = "x86_64")]
```
anupsdf, line 267: Interesting to see this! I guess its because [Github actions](https://github.com/stellar/rs-soroban-env/blob/main/.github/workflows/rust.yml#LL57C43-L57C43) is configured to run tests using x86_64 architecture.

jayz22: Yeah, this test is disabled currently. 
It is intended purely as a helper for finding out the proper size to use for the `declared_size` of a new type. It is platform-dependent, the actual `declared_size::<T>` has to be the upper bound of the sizes across all platforms. 
This was causing confusion when someone runs it on a different platform than where the hardcoded value was generated. Thus comes this fix.

graydon (approved):

Issue #538, opened 2022-10-13
Add host functions that calculate contract IDs from inputs


leighmcculloch:
We need a set of host functions that will return contract ID (as Bytes) given a set of inputs.

We have the following functions that can create contracts:
https://github.com/stellar/rs-soroban-env/blob/4f2f3a2da58dd869b559f41311815b20d1552ce5/soroban-env-common/src/env.rs#L307-L321

In total we need to add `12` new host functions:

- `7` new host functions that are counter-parts to the host functions above, that will return the contract ID that will be used if the deployment was to occur by calling the respective function.
   - `get_contract_id_contract_from_ed25519(salt: Object, key: Object) -> Object`
   - `get_contract_id_contract_from_contract(salt: Object) -> Object`
   - `get_contract_id_contract_from_source_account(salt: Object) -> Object`
   - `get_contract_id_token_from_ed25519(salt: Object, key: Object) -> Object`
   - `get_contract_id_token_from_contract(salt: Object) -> Object`
   - `get_contract_id_token_from_source_account(salt: Object) -> Object`
   - `get_contract_id_token_from_asset(asset: Object) -> Object`

- `2` new host functions for getting the contract ID for a deployment from contract that isn't the currently executing contract.
   - `get_contract_id_contract_from_other_contract(contract_id: RawVal, salt: Object) -> Object`
   - `get_contract_id_token_from_other_contract(contract_id: RawVal, salt: Object) -> Object`

We can omit any of these host functions that would always return the same value given the same inputs. For example, I expect we can omit all the token functions except the one that accepts an asset, 

Dependent on https://github.com/stellar/rs-soroban-env/issues/537
Blocking https://github.com/stellar/rs-soroban-sdk/issues/730

jayz22:
@leighmcculloch Is this issue still relevant? Most of the mentioned host functions have been removed and the mentioned issue has been closed.

leighmcculloch:
Yeah it should be possible for us to write an SDK function that helps a developer predetermine a contract ID that would be the contract ID of a contract when deployed.

leighmcculloch:
The infrastructure around deployments has changed a lot, so the examples in the original issue are out of date.

jayz22:
We currently have two host functions for creating a contract:

```
                    "name": "create_contract",
                    "args": [
                        {
                            "name": "deployer",
                            "type": "AddressObject"
                        },
                        {
                            "name": "wasm_hash",
                            "type": "BytesObject"
[...]

Issue #818, opened 2023-05-24
Change all Bytes for contract IDs to Address


leighmcculloch:
We need to change the host interface so that contract IDs are always communicated as an `AddressObject` instead of a `BytesObject`.

In https://github.com/stellar/rs-soroban-sdk/pull/947 we changed the way that the SDK externalizes contract IDs, so that they are wherever possible externalized as an `Address`. The SDK does a conversion from bytes to address and vice-versa to make that happen.

The conversions aren't expensive, but ideally the host would just always provide contract IDs as addresses since that is how the SDK externalizes it to contracts.

This change will require changing the types of values passed to or from many host functions. The SDK will need to be updated after this change is made.

jayz22:
Resolved by https://github.com/stellar/rs-soroban-env/pull/853. Only found one residual function `get_invoking_contract` and cleaned it up.

leighmcculloch:
@jayz22 There are other places I think, such as call stacks, and anywhere in auth a contract ID is specified. Maybe these were already cleaned up though. cc @dmkozh

jayz22:
The `get_current_call_stack` returns a Vector of `(RawVal, RawVal)`, with the contract id is stored as `Bytes` in the first RawVal. So I thought it was pretty much internal detail. But it's more consistent to change to Address too. I will clean this up.
All the external host functions I checked have contract ID specified as `AddressObject` already, internally it might still be referring to the `Hash` part (such as `vm.contract_id`) which should be okay. 
Reopening it to fix `get_current_call_stack` and any other ones that pop up.

jayz22:
https://github.com/stellar/rs-soroban-env/pull/866 fixed the `get_current_call_stack`. 
As far as I can tell this completes makes all the host interface work with `AddressObject`. If I missed any thing, feel free to reopen this.

PR #853, opened 2023-06-14
Add host functions for returning contract ids

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/538 

### Why

### Known limitations

[TODO or N/A]


soroban-env-common/env.json
```diff
@@ -1075,6 +1075,34 @@
                     ],
                     "return": "Void",
                     "docs": "Bumps the expiration ledger of the key specified so the entry will live for `min` ledgers from now. If the current expiration ledger is already large enough to live at least `min` more ledgers, then nothing happens."
+                },
+                {
+                    "export": "8",
+                    "name": "get_contract_id",
```
dmkozh, line 1081: nit: Maybe these should be called `compute` instead of `get`? `get` usually assumes retrieving something that already exists.

dmkozh (approved):

leighmcculloch (approved):

soroban-env-common/env.json
jayz22: Sounds good, changed it to `compute`. I agree that makes more sense.

leighmcculloch: Fwiw how the function derives its value is an implementation detail, not part of its interface.

leighmcculloch (approved):

leighmcculloch:
Looks like this PR has some merge conflicts that need fixing before it can be merged.

soroban-env-common/env.json
dmkozh: To me the difference is that `get` implies that the contract id exists somewhere, which isn't true. Maybe that's just my comprehension though.

jayz22:
Merged with latest env changes and fixed conflicts.

Aldo did a drive-by clean up on one host function returning contract id to use `AddressObject`, which should close https://github.com/stellar/rs-soroban-env/issues/818 too. 

@leighmcculloch I think it's ready to be merged.

PR #866, opened 2023-06-20
Support `Timepoint`, `Duration`

### LINKED ISSUE #724: Add host functions to transport objects between host and guest memory for `String`, `Timepoint`, `Duration`

jayz22:
### What

Resolves #724 
- Implement {`Timepoint`, `Duration`} <-> `u64` conversions
- A drive-by clean up for `get_current_call_stack` to return `Address`
- Clean up some trailing whitespaces in `env.json`

### Why

### Known limitations

[TODO or N/A]


leighmcculloch:
What's the win out of having these two disjoint types when they're just u64 values?

I think I missed the conversation when we added them.

jayz22:
> What's the win out of having these two disjoint types when they're just u64 values?
> 
> I think I missed the conversation when we added them.

They are special case u64 subtypes that can have different semantics and operations. They can represent for example ledger timestamp and can specify its own display format (instead of a plain `u64`).  Similar to how `String`, `Symbol` are related to `Bytes`.
For now we just support them by converting them to and from u64, which is the bare minimum so that the guest can implement the semantics in ways they want. We might enrich them post V1.

soroban-env-common/env.json
```diff
@@ -137,7 +137,7 @@
                         }
                     ],
                     "return": "U64Object",
-                    "docs": "Convert a u64 to an object containing a u64."
+                    "docs": "Convert an `u64` to an object containing an `u64`."
```
leighmcculloch, line 140: As far as I'm aware `Convert an u64` should be `Convert a u64`. My understanding is if the word following a/an starts with a `u`, it depends what sound the `u` makes for whether it's an `a` or a `an`. In this case the `u` sounds like `you`, and so it should be `a`.

leighmcculloch (approved):

soroban-env-common/env.json
jayz22: You are right :) Fixed them.

PR #873, opened 2023-06-22
Change StorageType to Just Be An Enum

graydon:
This changes `StorageType` to just be a plain Rust enum (rather than a `Val` wrapper) in preparation for removal of `StorageType` from `ScVal`.

It also includes a change expanding `StorageType` (the type at the env-interface layer) from 2 cases to 3 -- including `::Instance` -- deviating a bit from `ContractDataType` (the type that controls the storage layer). It includes a projection down to that lower layer type though.

This PR isn't 100% complete, it depends on some more changes incoming at the XDR level -- and it appears it's going to have to integrate a bunch of pending XDR changes from other work -- but I thought I'd post it for examination. I think it's not too bad a solution to the problem. We still get to use `StorageType` in `env.json` and we get a strong enum type on both sides of the interface. We just ignore / bypass `Val` entirely.


graydon:
@dmkozh and/or @sisuresh I ran out of time this evening updating this PR to the changes in the XDR (and the last bunch probably need additional eyes from someone who knows more of what the intended semantic changes are, I was somewhat guessing). Feel free to pick up and extend this with the remainder if you get to it before me, otherwise I'll pick up tomorrow.

jayz22 (approved):

sisuresh (approved):

PR #891, opened 2023-06-26
Change [IU]256 conversion host functions to take `Val`

jayz22:
### What

Follow up on https://github.com/stellar/rs-soroban-env/pull/889.
Changes the [IU]256 conversion functions from `Object` to `Val`.
Fixes some broken calibration tests.

### Why

https://github.com/stellar/rs-soroban-env/pull/889#issue-1772336785 explains the reasons for exposing Val instead of Object in the arithmetic host functions for the SDK.  It also added helper functions to allow constructing from u32. 

The reasons same apply for the conversion functions, which allows the guest to construct a general [IU]256 from 256 bits (either via a bytes objects or four [IU]64), and also taking advantage of the small-number optimization from the host side. 

### Known limitations

[TODO or N/A]


jayz22:
Discussed with @graydon , not a totally necessary change. Closing.

PR #899, opened 2023-06-27
Expiration-related fixes

dmkozh:
### What

- Make signatures unusable if nonce might expire
- Make nonce entries to live for at least the min allowed ledger number
- Make expiration computations consistent with the lifetime expectations

### Why

Security/consistency enhancements

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -2031,8 +2031,12 @@ impl VmCallerEnv for Host {
             .borrow_mut()
             .touch_key(&key, self.as_budget())?;
 
-        let min_expiration =
-            self.with_ledger_info(|li| Ok(li.sequence_number.saturating_add(min.into())))?;
+        let min_expiration = self.with_ledger_info(|li| {
+            Ok(li
+                .sequence_number
+                .saturating_sub(1)
+                .saturating_add(min.into()))
+        })?;
         self.0.expiration_bumps.borrow_mut().metered_push(
             self,
             LedgerBump {
@@ -2048,8 +2052,12 @@ impl VmCallerEnv for Host {
         _vmcaller: &mut VmCaller<Host>,
         min: U32Val,
     ) -> Result<Void, HostError> {
-        let min_expiration =
-            self.with_ledger_info(|li| Ok(li.sequence_number.saturating_add(min.into())))?;
+        let min_expiration = self.with_ledger_info(|li| {
+            Ok(li
+                .sequence_number
+                .saturating_sub(1)
+                .saturating_add(min.into()))
+        })?;
 
         let contract_id = self.get_current_contract_id_internal()?;
         let key = self.contract_instance_ledger_key(&contract_id)?;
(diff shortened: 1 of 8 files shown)


soroban-env-host/src/host.rs
```diff
@@ -2031,8 +2031,12 @@ impl VmCallerEnv for Host {
             .borrow_mut()
             .touch_key(&key, self.as_budget())?;
 
-        let min_expiration =
-            self.with_ledger_info(|li| Ok(li.sequence_number.saturating_add(min.into())))?;
+        let min_expiration = self.with_ledger_info(|li| {
+            Ok(li
+                .sequence_number
+                .saturating_sub(1)
```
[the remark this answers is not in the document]
dmkozh: Sounds good - that doesn't need to block this PR, right?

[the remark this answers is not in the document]
sisuresh: No it doesn't

sisuresh, line 2037: Does allowance.rs need to be updated as well? I think `(expiration - ledger_seq)` should become `(expiration - ledger_seq + 1)` right?

dmkozh: Good catch, updated.

soroban-env-host/src/native_contract/token/allowance.rs
```diff
@@ -35,7 +35,11 @@ pub fn write_allowance(
     // validates the expiration and then returns the ledger seq
     // The expiration can be less than ledger seq if clearing an allowance
     let ledger_seq = e.with_ledger_info(|li| {
-        if expiration > li.sequence_number.saturating_add(li.max_entry_expiration) {
+        if expiration
+            > li.sequence_number
+                .saturating_sub(1)
```
sisuresh, line 40: It's unfortunate that this logic is going to creep into contracts as well. I think we're going to see quite a few off by one errors (with or without your change).

sisuresh (approved):
It's unfortunate that seq - 1 logic is going to creep into contracts as well. I think we're going to see quite a few off by one errors (with or without your change).

dmkozh:
> It's unfortunate that seq - 1 logic is going to creep into contracts as well. I think we're going to see quite a few off by one errors (with or without your change).

That's why I'm suggesting to provide a helper method that bumps the entry to a given ledger. I don't think there are any other useful scenarios where the error is possible.

sisuresh:
> > It's unfortunate that seq - 1 logic is going to creep into contracts as well. I think we're going to see quite a few off by one errors (with or without your change).
> 
> That's why I'm suggesting to provide a helper method that bumps the entry to a given ledger. I don't think there are any other useful scenarios where the error is possible.

Oh yeah for the sdk right?

dmkozh:
> Oh yeah for the sdk right?

Yes, we can do that at the SDK level. Doing that in host fn is slightly safer for other SDKs, but probably that's not a big deal (these probably should borrow the implementation details from the Rust SDK)

PR #926, opened 2023-07-04
Misc int32 issues

### LINKED ISSUE #895: Limit map and vec host sizes to u32::MAX
In practice the budget will already limit this from ever happening; but an additional explicit check when building a map or vec to cap at u32::MAX size is worth having just in case.

graydon:
This just fixes a couple issues that can cause 32 bit hosts to diverge from 64 bit hosts in extreme cases (not likely one could ever hit them due to metering, but whatever). Also explicitly caps size of vectors and maps to u32::max.

Fixes #895

### DIFF
--- a/soroban-env-host/src/host/metered_map.rs
+++ b/soroban-env-host/src/host/metered_map.rs
@@ -84,6 +84,9 @@ where
     }
 
     pub fn from_map(map: Vec<(K, V)>, ctx: &Ctx) -> Result<Self, HostError> {
+        if u32::try_from(map.len()).is_err() {
+            return Err((ScErrorType::Object, ScErrorCode::ExceededLimit).into());
+        }
         // Allocation cost already paid for by caller, here just checks that input
         // has sorted and unique keys.
         let m = MeteredOrdMap {
--- a/soroban-env-host/src/host/metered_vector.rs
+++ b/soroban-env-host/src/host/metered_vector.rs
@@ -50,7 +50,7 @@ where
 {
     // Constructs a empty new `MeteredVector`.
     pub fn new() -> Self {
-        Self::from_vec(Vec::new())
+        Self { vec: Vec::new() }
     }
 
     // Constructs a new, empty `MeteredVector` with at least the specified capacity.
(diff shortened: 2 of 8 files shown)


soroban-env-host/src/host/metered_vector.rs
```diff
@@ -49,7 +49,7 @@ where
     A: MeteredClone,
 {
     // Constructs a empty new `MeteredVector`.
-    pub fn new() -> Self {
+    pub fn new() -> Result<Self, HostError> {
```
jayz22, line 52: Is the reason of changing this to a Result to be consistent with the other Vec construction functions? If so, maybe change `MeteredOrdMap::new()` to return Result as well?

graydon: No, it's because it delegates to `Self::from_vec` which itself had to change here. Which is silly, I should just rewrite it to return `Self { vec: Vec::new() }`. I'll push a fix, thanks for pointing it out.

(Kinda wrote this one on autopilot)

soroban-env-host/src/host/metered_map.rs
```diff
@@ -84,6 +84,9 @@ where
     }
 
     pub fn from_map(map: Vec<(K, V)>, ctx: &Ctx) -> Result<Self, HostError> {
+        if u32::try_from(map.len()).is_err() {
```
dmkozh, line 87: No strong opinion here, just a thought.: while I understand this check in theory, it seems to only introduce some additional confusion. There shouldn't ever be objects that are even close to 100 million elements, not to mention billions. So overflow check seems somewhat arbitrary and confusing to me here as it implies that it's possible to create a big enough vector somehow.

dmkozh (approved):

soroban-env-host/src/host/metered_map.rs
graydon: Agreed, it's strictly a theoretical condition. The testcase submitted to demonstrate it had the budget turned off (and when I ran it it consumed 45GiB RAM on my workstation). Also it hinges on usize not being u64, which in practice is never will be either (nobody runs 32-bit hosts anymore). This is a dotting "i"s, crossing "t"s issue. A formality.

PR #928, opened 2023-07-05
Enforce object handle integrity when inserting into containers, fix #569

### LINKED ISSUE #569: object handle integrity
object handles entering the system from SCObject -> HostObj -> guest Object handle conversion have their type tags set correctly, but a cross-contract call can forge an object handle with an invalid tag. This can't corrupt the host -- the host ignores those tags, they're supposed to just be informative to save round trips inquiring about types -- but it can lead to some potential confusion or latent corruption in guest code that acts on the basis of forged tags.

@leighmcculloch suggested a nice fix to this which is to enforce integrity while we're plumbing RawVals from one contract to another (either as calls or returns) such that a forged handle causes the forger to trap. This is a good idea and we should do it.

graydon:
This isn't so much a "safety" issue as a confusion-of-users issue. We want to detect any malformed object handles (bad index, bad tags) as early as possible, so users don't have to diagnose late errors when a lookup finally traps.

Anyway, this makes it impossible to insert a malformed one into a container, and since containers are how you form argument lists to making calls, it implicitly confines malformed object handles to a single contract instance.

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -1000,6 +1000,9 @@ impl EnvBase for Host {
         for k in keys.iter() {
             key_syms.push(Symbol::try_from_val(self, k)?);
         }
+        for v in vals.iter() {
+            self.check_val_integrity(*v)?;
+        }
         let pair_iter = key_syms
             .iter()
             .map(|s| s.to_val())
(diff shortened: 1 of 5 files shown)


dmkozh (approved):

soroban-env-host/src/host.rs
```diff
@@ -1974,6 +1992,8 @@ impl VmCallerEnv for Host {
         t: StorageType,
         f: Val,
```
anupsdf, line 1993: Should we check value integrity for the input flag f as well?

PR #930, opened 2023-07-05
Revive read-tickets WIP

graydon:
Taking over from #132 which had bitrotted. Still WIP.

### DIFF
--- a/soroban-env-host/src/host_object.rs
+++ b/soroban-env-host/src/host_object.rs
@@ -1,6 +1,8 @@
+use bs::Bs;
 use soroban_env_common::{
-    xdr::ContractCostType, Compare, DurationSmall, I128Small, I256Small, I64Small, SymbolSmall,
-    SymbolStr, Tag, TimepointSmall, U128Small, U256Small, U64Small,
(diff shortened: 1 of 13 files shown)


soroban-env-host/src/host_object.rs
```diff
@@ -18,11 +20,74 @@ use super::{
     U64Object, Val, VecObject,
 };
 
+/// A ticket is a number that represents permission to read a set of host
```
MonsieurNicolas, line 23: Maybe I am missing something... but because objects have to be passed in as argument in order to be visible, this seems maybe over complicated? I *think* the only thing to keep track of is how deep in the stack something was shared.

graydon: No, it's unfortunately much more complicated than that. As a minimal example speaking to your question: if A calls B(x), then x should be readable by B, sure; but if B returns and A then calls C(y), x should _not_ be readable by C, even though C is "at the same call depth" as B was.

graydon:
closing in favor of https://github.com/stellar/rs-soroban-env/pull/947

PR #938, opened 2023-07-10
Refactor authorization manager to only maintain mutable borrow on minimal amount of fields

dmkozh:
### What

Refactor authorization manager to only maintain mutable borrow on minimal amount of fields.

### Why

This avoids auth manager re-borrow hack and allows calling `require_auth` from within `__check_auth`. The latter might not be the most important feature, but it makes things more consistent.

### Known limitations

N/A


soroban-env-host/src/test/auth.rs
```diff
@@ -1336,7 +1336,52 @@ fn test_invoker_subcontract_with_gaps() {
         host_vec![&test.host, test.convert_setup_tree(&mid_gap_setup)],
         vec![],
         false,
-    )
+    );
+}
+
+#[test]
+fn test_nvoker_subcontract_auth_without_subcontract_calls() {
```
anupsdf, line 1343: Typo invoker?

graydon (approved):
Currently over my head but I am generally on board with "finer grained borrows" and willing to defer careful review of this code until we're reviewing the whole subsystem.

PR #946, opened 2023-07-13
Adapt to `wasmi::ResourceLimiter`, replacing mem_fuel metering

### LINKED ISSUE #896: [test] Ensure data segment and user-defined memories can't exceed memory budget limit
Add a "hostile wasm" test that makes a very large memory and a very large data segment (declaratively, in the wasm declaration tables rather than a grow-memory op) and check that these are trapped by memory budget limits.

(And obviously if not trapped, make it so)

jayz22:
### What

This PR adapts to the new `wasmi::ResourceLimiter` api, replacing our current metering-based guest memory limiting mechanism. The main benefits are:
- More resources (table, instances) are limited, not just memory
- It also limits memory allocation on startup, preventing guest contract from declaring large memory in in the wasm file. Resolves #896 

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/vm.rs
+++ b/soroban-env-host/src/vm.rs
@@ -36,7 +36,7 @@ use wasmi::{Caller, StoreContextMut};
 
 impl wasmi::core::HostError for HostError {}
 
-pub const WASM_LINEAR_MEMORY_SIZE_LIMIT_BYTES: usize = 0x200000; // 20MB
+pub const WASM_LINEAR_MEMORY_SIZE_LIMIT_BYTES: usize = 0x200000; // 2MB
 pub const WASM_TABLE_ELEMENTS_LIMIT_COUNT: u32 = 1000;
 
 /// A [Vm] is a thin wrapper around an instance of [wasmi::Module]. Multiple
--- a/soroban-env-host/src/vm.rs
+++ b/soroban-env-host/src/vm.rs
@@ -24,7 +24,10 @@ use soroban_env_common::{
     ConversionError, SymbolStr, TryIntoVal, WasmiMarshal,
 };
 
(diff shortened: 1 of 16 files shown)


soroban-env-host/src/vm.rs
```diff
@@ -33,6 +36,9 @@ use wasmi::{Caller, StoreContextMut};
 
 impl wasmi::core::HostError for HostError {}
 
+pub const WASM_LINEAR_MEMORY_SIZE_LIMIT_BYTES: usize = 0x200000; // 2MB
+pub const WASM_TABLE_ELEMENTS_LIMIT_COUNT: u32 = 1000;
```
jayz22, line 40: I've added two new limits here. I think the linear memory size 2MB should be reasonable? Not sure about the table elements count. Would like some feedback.

graydon (changes requested):
While I suspect this all works, I think I might have miscommunicated the approach I had in mind. I was hoping (and please correct me if this is not possible) to make `Host` impl `ResourceLimiter` directly, such that the callbacks would deduct memory usage from the `Budget` directly. I.e. I was thinking we would _not_ be using the `StoreLimits` type from wasmi -- that is a convenience type that implements `ResourceLimiter` but it is not the only impl possible and, in our case, not the one I was expecting us to use.

(Part of the reason I want to do it this way is that I would like the total memory budget to be available "in any combination" to all VMs that get instantiated, such that 1 VM can use the entire 100MB of the budget, if it wants, or the 100MB can be split across multiple VMs as they call one another).

jayz22:
Closing in favor of https://github.com/stellar/rs-soroban-env/pull/950

PR #956, opened 2023-07-18
Add `soroban-bench-utils`, add benchmark tests to measure metering accuracy

### LINKED ISSUE #710: Integration testing for metering coverage
We need a strategy for testing our metering coverage. One way is to enable the host to run in "calibration mode", where we turn on cpu and memory counter (like what we do in benches, but for end-to-end), and compare the actual measured costs to the modeled costs. The calibration model should be accessible from outside of the host (i.e. for integration test) such that we can run any host function or wasm contract in both on and off modes in order to compare their difference.

jayz22:
### What

Resolves #710 
This PR moves the instrumentation helpers from the `bench` to the host (only turned on for linux and macos), and added a new helper class `HostTracker` to get the actual resource usage.

Added two test contracts and run them with instrumentation turned on, here are the results comparing modeled vs actual usage:

```

$RUST_TEST_THREADS=1  cargo test --release --package soroban-env-host --lib -- test::metering_benchmark  --nocapture --ignored

running 2 tests
test test::metering_benchmark::run_add_i32 ... 
model cpu consumed: 1138241, actual cpu insns 1452320 
model mem consumed: 1274466, actual mem bytes 1213168
ok
test test::metering_benchmark::run_complex ... 
model cpu consumed: 1618593, actual cpu insns 1822004 
model mem consumed: 1308496, actual mem bytes 1244696
ok
```
Which looks decent, but the difference is hard to tell due to `VmInstantiation` taking up majority of the share. 

### Why

### Known limitations

1. ~The tracking_allocator dependency has been moved from dev to actual. Because the `HostTracker` is used for "bench" as well (which is under feature `testutils`, but still needs the dependency). Alternatively we make `mod instrumentation` a separate package and use it as a dev-dependency.~ I've moved the tracking-related utils to a separate package `soroban-bench-utils` to avoid the prod dependency.
2. These tests are disabled by default. For two reasons (added a comment in the test file): 
    1. They need to be run with release profile.
    2. They cannot be run with multiple threads due to the `gloabal_tracker` contention. 
3. ~Only supports os="linux" and "macos", due to the cpu instruction counting library being only available there.~ The "other" os now has a stub `InstructionCounter` that returns 0 always.


graydon (approved):
LGTM, thanks!

PR #960, opened 2023-07-21
Fix panic-string-logging code path broken by recent dynamic-borrow fix.

graydon:
In https://github.com/stellar/rs-soroban-env/pull/925 I accidentally changed the control flow in panic-recovery to a double-nested `if let`. Unfortunately there's an `else` after the outer block, so if the second `if let` fails we skip the `else`.

This results in an SDK test failure (`test_invoke_expect_string`) which is blocking completion of an SDK update to the current env.

This PR fixes it (and tightens up the comments).


soroban-env-host/src/host/frame.rs
```diff
@@ -601,25 +601,32 @@ impl Host {
...
                             // panicked. If if was a panic generated by a
-                            // Env-upgraded HostError, it had its status
-                            // captured by VmCallerEnv::escalate_error_to_panic:
-                            // fish the Error stored in the frame back out and
+                            // Env-upgraded HostError, it had its `Error`
+                            // captured by `VmCallerEnv::escalate_error_to_panic`:
+                            // fish the `Error` stored in the frame back out and
                             // propagate it.
                             let func: Val = func.into();
                             let mut error: Error = Error::from_type_and_code(
```
dmkozh, line 609: What is the condition to have the internal error here after this fix?

dmkozh (approved):

soroban-env-host/src/host/frame.rs
graydon: First of all this whole path only occurs in native-testing mode (it's cfg(feature="testutils)) -- it has to do with catching _panics_ which we do not do in production, the facility is only here to emulate the behaviour of VM contracts ("halt on error") when compiled in a non-VM mode (natively) where they panic.

A panic in contract code can happen two ways:

  1. Some host function generates a `HostError` and the guest "rejects" it (using `env.escalate_error_to_panic`) and we store the rejected error in the frame, panic, unwind, catch the panic and retrieve the error from the frame here. In that case we get a non-`InternalError`
  2. Some user code calls `panic!("string")` or `assert!(...)` or something, generating a panic. In a VM build that will execute the WASM `unreachable` instruction which traps the VM, so it would hit the [code path that does the conversion from that to an XDR `Error` code](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-common/src/error.rs#L164) which I see you recently modified from `InternalError` to `InvalidAction` (and for fidelity with the Vm case I guess we'd still put it under `ScErrorType::VmError`).

I'm not really interested in re-debating the question of what `InternalError` should mean, you have a strong belief about it and I only have a weak one, so I'm happy to change the default output from this from `InternalError` to `InvalidAction`. I'm not 100% sure it's best to put it under `ScErrorType::Context` or `ScErrorType::WasmVm` though. The argument _for_ `WasmVm` is that the panic-catch here is "emulating the WASM VM" so it's reasonable to call the error that. The argument _against_ is that it'll confuse people to tell them there was a WASM VM error when they're running in native testing mode. I could see it going either way.

Issue #522, opened 2022-10-04
A contract invocation that fails with a trap or error results in a successful operation / transaction result


leighmcculloch:
Something that @paulbellamy noticed while testing against stellar-core is that when a invocation fails with an error, the operation and transaction result are success.

This is unintuitive. It's very easy to write tools and applications that think everything worked as expected.

cc @graydon

anupsdf:
Setting target milestone to `Testnet Phase 1` after discussing with @graydon

graydon:
@paulbellamy the error-propagation path in core has changed quite a bit since this was filed, and I'm not 100% sure if it's been fixed or not. Can you elaborate a bit here? Like, specifically, what you mean by "invocation fails". Do you mean when a contract successfully returns a value that happens to be a Rust `Val::Error` / XDR `SCError`? Or do you mean like if the host fails execution due to running out of gas or an illegal operation or something like that?

graydon:
I've made some end-to-end tests that "returning `Val::Error` causes the transaction to fail" and it all looks like it works, so I _think_ this was solved some time ago. Please reopen if you see anything fishy here in the future.

PR #967, opened 2023-07-27
Add new tests for error escalation from contract calls.

graydon:
Test that contract calls that return various `Error` all escalate to `Err(Error)` at the host / frame level.

This is part of ensuring #522 but we also need to run the same tests up at the core level and ensure we route these all the way to transaction failures there.


soroban-test-wasms/wasm-workspace/errors/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+use soroban_sdk::{contract, contractimpl, contracterror, Val, Error};
+
+#[contract]
+pub struct Contract;
+
+#[contracterror]
+#[derive(Copy,Clone)]
+pub enum Eek
+{
+    BADNESS = 12345
```
dmkozh, line 11: Somewhat unrelated, but will this work with u32::Max?

Also a bit weird that there is no `repr(u32)` here, I thought it's required...

dmkozh (approved):

soroban-test-wasms/wasm-workspace/errors/src/lib.rs
graydon: `u32::MAX` as an expression doesn't work -- it wants a literal. If I put `0xffff_ffff` in, it doesn't like it as the default repr choice is `isize` which is `i32` on `wasm32-unknown-unknown` and this value doesn't quite fit. If I put in `repr(u32)` it works.

In general `repr(u32)` isn't required for `#[contracterror]`, it just generates conversions to and from `Error` and `Val` that switch on the enumerated values -- so long as the enumerated values fit in the repr in question, `#[contracterror]` doesn't care (and if you provide enumerated values that _don't_ fit in the repr chosen it simply won't compile). So like .. if you _want_ to push right up against against the maximum range by using `0xffff_ffff` then yeah you'd need `repr(u32)`. But only if you want to do that.

PR #975, opened 2023-07-31
Add helpers for container bulk init; applies to auth metering

### LINKED ISSUE #668: Auth Next: Audit the metering

jayz22:
### What

Resolves #668 
- Close metering gaps in `Auth`-related code. Mostly around unmetered container usages i.e. `Vec<T>`. 
- Added a couple of helper traits `MeteredContainer`, `MeteredCollect` to make the bulk collection with metering a little bit more seamless. 

### Why

### Known limitations

[TODO or N/A]


soroban-env-host/src/auth.rs
```diff
@@ -404,6 +411,7 @@ impl AuthorizedFunction {
                             Ok(addr.clone())
                         })?,
                     function_name: contract_fn.function_name.try_into_val(host)?,
+                    // why is this intentionally non-metered? the visit_obj above is metered.
```
dmkozh, line 414: This whole function is not meant to be metered. I'm not sure if we have a non-metered visit_obj version. 
I think we can just get rid of it (or the current implementation) and replace with `host.with_free_budget(self.to_xdr())`.

soroban-env-host/src/auth.rs
```diff
@@ -1135,10 +1203,14 @@ impl InvocationTracker {
         None
     }
 
-    fn push_frame(&mut self) {
+    // metering: covered
+    fn push_frame(&mut self, budget: &Budget) -> Result<(), HostError> {
```
dmkozh, line 1207: I _think_ we already account for this as a part of the `GuardFrame`, so this is technically charged twice.

dmkozh (commented):
This looks nice, thanks! I think that's much better than just using metered vector internally at cost of n^2 memory and allocations.

soroban-env-host/src/auth.rs
jayz22: You are right.
In this case I think `GuardFrame` is no longer necessary. It merely does a context frame push + auth frame push, each of which can (and should) be metered individually. I will clean this part up, and remove the `xdr::GuardFrame`.

dmkozh: Yeah, that sounds good, thanks.

jayz22: Okay, I've cleaned up metering of `GuardFrame`. Will do the xdr removal in a follow up.

soroban-env-host/src/auth.rs
jayz22: There is a bit of divergence between the `to_xdr` and `to_xdr_non_metered` with the `exhausted_sub_invocations_only` check in the non-metered version only. I'll leave this piece along and perhaps you can clean it up later.

dmkozh (approved):

soroban-env-host/src/auth.rs
jayz22: Tracking issue for followup https://github.com/stellar/rs-soroban-env/issues/977

Issue #963, opened 2023-07-25
Upgrade everything to dalek 4.0


graydon:
Dalek finally released a new major version. We should upgrade to it.

PR #980, opened 2023-08-02
Upgrade dalek to at least next-pre-release versions.

graydon:
This is related to #963 -- it turns out dalek only released a stable (4.0) of the _curve25519_ crate not the _ed25519_ crate -- they're still working on the latter. This PR does the work of upgrading to a prerelease state of both (4.0-rc.3 and 2.0-rc.3 respectively) which is .. _probably_ identical API-wise to what we'll see on 4.0 and 2.0. But it's not stable _yet_.

I'm not 100% sure what to do here. I think it's _probably_ better to move to their prereleases assuming an imminent and uneventful final (especially given how old and unmaintained the previous release had got). But that could also be taking a risk of immature or incompletely-tested new code we don't really want to take. Open to input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,15 +138,6 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "block-buffer"
-version = "0.9.0"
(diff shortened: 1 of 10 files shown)


graydon:
Dalek has now fully released v2.0.0 of the ed25519 crate too. I'll open a new PR about that.

PR #982, opened 2023-08-03
Scale the linear cost model coefficient; improve model fitting

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/824

Add a scaling factor `COST_MODEL_LIN_TERM_SCALE_BITS` by which the linear cost model coefficients is scaled. 

Minor improvements in the measurement and fitting code, such as choosing a large enough step size (wrote a big comment for it) and using a proper "ceiling" on the coefficients.

Some minor clean ups and renaming.

### Why

All the cost coefficients are u64. Some of the linear coefficients are small, e.g. for `HostMemCmp`, `HostMemCpy`,  and previously we rounded them off and make them a const model. 
A large portion of the code base is doing memory operation, thus a small round off may lead to significant diff in actual metering accuracy. 
Multiplying a constant scaling factor by the linear coefficient is an easy way (compared to using say fixed-point integers) to solving this problem. 

Here is a rough impact of this change in metering benchmark results:

```diff
running 2 tests
test test::metering_benchmark::run_add_i32 ... 
-model cpu consumed: 1147842, actual cpu insns 1460911 
+model cpu consumed: 1568349, actual cpu insns 1462032
-model mem consumed: 1274594, actual mem bytes 1213688
+model mem consumed: 1287091, actual mem bytes 1213688
ok
test test::metering_benchmark::run_complex ... 
-model cpu consumed: 1630980, actual cpu insns 1827787 
+model cpu consumed: 2061258, actual cpu insns 1816229
-model mem consumed: 1308664, actual mem bytes 1245216
+model mem consumed: 1321108, actual mem bytes 1245216
ok
```
Previously we were pretty heavily under-underestimating the cpu cost, and now we are slightly overestimating (which is what our model is supposed to do). 

### Known limitations

[TODO or N/A]

Here is the model fitting [output](https://github.com/stellar/rs-soroban-env/files/12254978/output20.txt)


graydon (approved):
LGTM! One thing you might want to do in a followup -- which we discussed in meeting today -- is to make a special `struct ScaledU64(u64);` that does nothing more than indicate at the type-system level when a u64 is this scaled type vs. a non-scaled value. Then users could call a constructor like `impl ScaledU64 { fn from_unscaled(x:u64) -> Self { Self(x << 7) } }` or such, just to make it very clear where scaled values vs. unscaled ones are required.

jayz22:
> 

That's a good suggestion! I will make the follow up change in https://github.com/stellar/rs-soroban-env/pull/984

Issue #981, opened 2023-08-02
Re-evaluate the necessity for all the contract cost types before testnet


jayz22:
The contract cost types were created since the beginning of Soroban to represent metering components of the Soroban host. 
The cost definitions has undergone many iterations and currently these 31 [cost types](https://github.com/stellar/stellar-xdr/blob/9125ce3e4e00f168c1d3d41890795c7d534b12f3/Stellar-contract-config-setting.x#L92-L158) encompasses all the metering components. This issue is to reevaluate all the definitions to make them "testnet finalized". 

Here are the list of criteria for the existence of a cost type (in accordance with the [metering CAP](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-10.md)):

1. Significance: 
    a. has either linear or large constant cost in either of {cpu, mem} dimensions
    b. or if the cost is a small constant, it can happen in high frequency
    c. the existence threshold for a constant cost is the cost of `ChargeBudget` (~100 cpu insns)
2. Standalone: no double counting, i.e. no charging (or a different type) inside a charge
3. Recursion prevention: if there is a recursion (loop) path of un-predetermined depth (iterations) in the host without encountering budget charging, we should add some charge. 
4. Ubiquity: this is for future proofing. Some code paths are ubiquitous, such as visiting a host object, or invoking a host function. Adding a metering charge at these "choke points" can prevent 1. someone finding a crack in the metering system to do infinite amount of work for free 2. we add extend the host functionality in the future but forget to add metering. These provide a minimal base coverage.

jayz22:
Questionable types based on above criteria:
- [ ] `VisitObject`
- [ ] `ValXdrConv`
- [ ] `GuardFrame`
- [ ] `InvokeHostFunction`
- [ ] `InvokeVmFunction`
- [ ] `HostMemCpy`/`HostMemCmp` - these are more sublinear, need to be evaluated after https://github.com/stellar/rs-soroban-env/issues/824

jayz22:
**ChargeBudget** Should be removed. 

### Why: 
It is a small constant call that cannot happen recursively or in bulk -- fails the significance check. 

It was originally included mainly because of metering of wasm instructions. Original comment:
```
        // NB: charging a cost-amount to the budgeting machinery itself seems to
        // cost a similar amount as a single WASM instruction; so it's quite
        // important to buffer WASM step counts before flushing to budgeting,
        // and we add a constant charge here for "the cost of budget-counting"
        // itself.
```
[...]

PR #984, opened 2023-08-04
Clean up budget cost types

jayz22:
### What

- Resolves https://github.com/stellar/rs-soroban-env/issues/981
- Also does follow up to https://github.com/stellar/rs-soroban-env/pull/982, adding a `ScaledU64` type and use it explicit for the linear cost parameter, which turned out to be a bigger-than-expected change that required changes to some existing types and interfaces in `budget`. 
- Fixed a bug in cpu->fuel conversion and added a test.
- Did various budget metering related cleanups along the way.
- Recalibration. Here is the [output](https://github.com/stellar/rs-soroban-env/files/12297034/output22.txt)

### Why

### Known limitations

[TODO or N/A]


soroban-env-host/src/budget.rs
```diff
@@ -703,36 +1056,53 @@ impl Budget {
...
     }
 
-    pub(crate) fn get_cpu_insns_remaining_as_fuel(&self) -> Result<u64, HostError> {
-        let cpu_remaining = self.get_cpu_insns_remaining()?;
-        let cpu_per_fuel = self
-            .0
-            .try_borrow_or_err()?
-            .cpu_insns
-            .get_cost_model(ContractCostType::WasmInsnExec)
-            .linear_term;
```
jayz22, line 719: This is the fuel conversion bug I mentioned. This should have be `const_term` instead of `linear_term`. It is basically undercharging cpu for wasm instructions by a factor of 7.

graydon: Interesting! Why wasn't this noticed numerically?

graydon (approved):

soroban-env-host/src/budget.rs
jayz22: I'm glad you asked! My above statement about undercharging wasn't accurate.
This code is called when we calculate how much fuel (converted from the cpu budget) should be provide to the VM. So it affects the "fuel limit" perceived by the VM. The bug here means the VM will think it has a fuel budget that's 7x larger than what it is.
When the VM execution comes back, we charge the number of fuel units back to the cpu budget, which goes through the normal model evaluation flow, with the correct model, and the amount of cpu consumed and charged to the budget is correct. 
So the implication of this bug is: it only affects over-the-cpu-limit contracts, where they will be allowed to 7x run longer before trapping. All contracts will be charged the correct amount, the violating ones might be charged 7x the budget limit. 
I have enhanced the `test_vm_fuel_metering` test to cover this scenario.

PR #996, opened 2023-08-15
Charge write fees for read-only rent entry bumps.

dmkozh:
### What

Charge write fees for read-only rent entry bumps.

Write entry bumps are already charged separately as a part of the total write fees.

### Why

RO entry bumps need to be written in ledger, so there is always some 'flat' fee component due to the write (so bumping the entry by 1 ledger is not just 1 stroop).

### Known limitations

N/A


graydon (changes requested):
Mostly looks good! One change request: can you change the arithmetic here to be saturating_mul / saturating_add based?

dmkozh:
> One change request: can you change the arithmetic here to be saturating_mul / saturating_add based?

FWIW this is supposed to only be called on sanitized small inputs (everything is controlled by the protocol-imposed limits), so I wanted to save some typing/cycles on this. But I guess it's not really significant, so might as well just have safer ops everywhere.

soroban-env-host/src/fees.rs
```diff
@@ -234,9 +243,30 @@ pub fn compute_rent_fee(
     current_ledger_seq: u32,
 ) -> i64 {
     let mut fee = 0;
+    let mut bumped_read_only_entries: i64 = 0;
+    let mut bumped_read_only_key_size_bytes: u32 = 0;
     for e in changed_entries {
         fee += rent_fee_per_entry_change(e, fee_config, current_ledger_seq);
+        if e.read_only && e.old_expiration_ledger < e.new_expiration_ledger {
+            bumped_read_only_entries += 1;
```
graydon, line 251:
```suggestion
            bumped_read_only_entries = bumped_read_only_entries.saturating_add(1);
```

soroban-env-host/src/fees.rs
```diff
@@ -234,9 +243,30 @@ pub fn compute_rent_fee(
     current_ledger_seq: u32,
 ) -> i64 {
     let mut fee = 0;
+    let mut bumped_read_only_entries: i64 = 0;
+    let mut bumped_read_only_key_size_bytes: u32 = 0;
     for e in changed_entries {
         fee += rent_fee_per_entry_change(e, fee_config, current_ledger_seq);
```
graydon, line 249:
```suggestion
        fee = fee.saturating_add(rent_fee_per_entry_change(e, fee_config, current_ledger_seq));
```

soroban-env-host/src/fees.rs
```diff
@@ -258,20 +288,20 @@ fn rent_fee_per_entry_change(
             entry_change.new_expiration_ledger
                 - entry_change.old_expiration_ledger.max(current_ledger - 1),
```
graydon, line 289: I _think_ this can underflow if current_ledger > new_expiration_ledger > old_expiration_ledger

graydon: (I'm not sure if a saturating sub is better here or hoisting the max up to the `if` condition that guards this block

graydon (changes requested):
Couple more minor ones..

soroban-env-host/src/fees.rs
dmkozh: `current_ledger > new_expiration_ledger > old_expiration_ledger` should never happen - again, this is not a library to use for arbitrary inputs, but a helper to use with well-behaved state managed by core or preflight. These shouldn't ever pass the expired entries to the host.

dmkozh:
Given that we might be changing how the rent works, I'll close this for the time being as it probably will become irrelevant.

graydon (approved):

PR #1011, opened 2023-08-21
Fix comparison in Tag::is_object

brson:
### What

Fix `Tag::is_object` and delegate `Val::is_object` to it.

### Why

The comparison in `Tag::is_object` is incorrect, causing the function to always return `true`.

This code is duplicated from the (correct) `Val::is_object` function.

### Known limitations

None.


brson:
I'm going to do this differently.

PR #1012, opened 2023-08-22
Fix comparison in Tag::is_object

brson:
### What

Fix comparison in `Tag::is_object`.

### Why

This function incorrectly always returns true. The more-commonly-used `Val::is_object` is correct. This patch makes both functions agree.

I would prefer `Val::is_object` delegate to `Tag::is_object`, but that can't be done without constructing a `Tag`, which has several branches, and I take it `Val::is_object` is a fast path.

### Known limitations

`Tag::is_object` and `Val::is_object` duplicate code.


graydon (approved):
Nice catch! LGTM

PR #1027, opened 2023-08-31
Run metering-vs-actual test in CI

jayz22:
### What

### Why

### Known limitations

[TODO or N/A]


jayz22:
This was an experiment to try to run instruction calibration and compare to our metering results within CI's execution environment. 
With many iterations, it appears like this wouldn't work because of the limited access to the instruction counter in the CI's environment. For ubuntu-x86 job it failed to run and for apple-x86 it runs successfully but doesn't return the instruction count. 
Closing this for now. We might have a better way in the future. 
I will cherry-pick the test output improvements into another PR.

PR #1034, opened 2023-09-04
host: fix EXPIRATION_ENTRY_SIZE constant

2opremio:
It didn't account for the size of the ledger entry type and` modifiedLedgerSequence` field

See https://github.com/stellar/soroban-tools/pull/922#issuecomment-1705721905 for more details


2opremio:
@sisuresh can you take a look and also update the SDK when merged?

sisuresh (approved):
Ah good catch. Thanks for the fix and the test.

Issue #1006, opened 2023-08-18
[Last item before stable release] Set pre_release_version to 0 before going to Testnet


anupsdf:
### What?
Set pre_release_version to 0 before going to Testnet in [meta.rs](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-common/src/meta.rs#L45)

### Why?
When a release goes out, it should have pre-release version zero. During development of a new release, or before the initial release of soroban, a nonzero pre-release number can be used to force recompiles on interface changes.

PR #1039, opened 2023-09-06
Setting pre_release_version to zero in meta.rs in preparation of Soroban launch

anupsdf:
### What
Setting pre-release version to zero in preparation for Soroban launch
To resolve https://github.com/stellar/rs-soroban-env/issues/1006

### Why
See the comments in the file for more [information](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-common/src/meta.rs#L36-L39)


graydon (approved):

graydon:
We decided to merge this now -- nobody's taking any further snapshots before final that might exhibit version skew, and the only possible perturbations we might make to the env interface are those in local wasms we can just regenerate ourselves immediately (as we would anytime we "change the prerelease number"). IOW it only existed to help _other_ parties downstream version their wasms, and none should be looking at this interface in the (small) window between now and final.

PR #1041, opened 2023-09-06
Fix asset-code rendering in native contract.

### LINKED ISSUE #1032: Null characters rendered in assets codes of built-in Stellar Asset Contract

- stellar/soroban-tools@b1519a4d90502ce3155960541b2a98c88f3dd043 (rpc and cli)
- stellar/stellar-core@22b9bb38416e5ce058eb9c1fadfdd414a4661833
- stellar/rs-soroban-env@eb5a9ba053a7b64a8eff605db625525378f7bea0

### What did you do?

```console
❯ soroban lab token wrap --network local --source me --asset USD:GDP7GGFMNMWABGWTF6BUB45MN6TPOCFY43KJVK5LDBCSJ2QTMQYYU3DA
CAW7SH5JG45U6YLCFB6YD2KVSOZQHMNEJQPN6KDZOCGSXWR72ZIFBBMP

❯ soroban contract invoke --network local --source me --id CAW7SH5JG45U6YLCFB6YD2KVSOZQHMNEJQPN6KDZOCGSXWR72ZIFBBMP -- name
```

The `local` network is a local quickstart running the commits above.

The `me` source is the networks root account.

```console
❯ soroban contract invoke --network local --source me --id CAW7SH5JG45U6YLCFB6YD2KVSOZQHMNEJQPN6KDZOCGSXWR72ZIFBBMP -- name
"USD:GDP7GGFMNMWABGWTF6BUB45MN6TPOCFY43KJVK5LDBCSJ2QTMQYYU3DA"
```

### What did you see instead?

```console
❯ soroban contract invoke --network local --source me --id CAW7SH5JG45U6YLCFB6YD2KVSOZQHMNEJQPN6KDZOCGSXWR72ZIFBBMP -- name
"USD\u0000:GDP7GGFMNMWABGWTF6BUB45MN6TPOCFY43KJVK5LDBCSJ2QTMQYYU3DA"
```

### Discussion

Stellar assets have two types of asset codes, a 4 byte and 12 byte code. The specification of less than a 4 byte asset code when wrapping is selecting the 4 byte asset type, and the remaining characters are null (`\u0000`).

When rendering the asset to a string, it should be rendered without the null bytes. The null bytes are indicative of the string ending early, and not of characters that are meaningfully part of the code itself.

See [SEP-11](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#alphanum4-alphanum12) for the correct rendering of the asset code.

graydon:
I thought this was just a UI error but the native contract stores this rendering in the database, so it's really a protocol-level correctness issue.

Fixes #1032

### DIFF
--- a/soroban-env-host/src/native_contract/token/metadata.rs
+++ b/soroban-env-host/src/native_contract/token/metadata.rs
@@ -1,8 +1,12 @@
+use std::fmt::Write;
+
 use soroban_native_sdk_macros::contracttype;
 use stellar_strkey::ed25519;
 
-use crate::{host::Host, HostError};
-use soroban_env_common::{Env, EnvBase, StorageType, SymbolSmall, TryFromVal, TryIntoVal};
+use crate::{host::Host, native_contract::base_types::BytesN, HostError};
+use soroban_env_common::{
+    ConversionError, Env, EnvBase, StorageType, SymbolSmall, TryFromVal, TryIntoVal,
+};
 
 use crate::native_contract::base_types::String;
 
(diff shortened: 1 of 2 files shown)


soroban-env-host/src/native_contract/token/metadata.rs
```diff
@@ -20,33 +24,149 @@ pub struct TokenMetadata {
...
+        match *x {
+            // When dealing with a 4-byte asset code we stop at the first NUL.
+            0 if buf.len() == 4 => break,
+            // When dealing with a 12-byte asset code we continue escaping NULs
+            // as \x00 until past the 5th byte, so that the result is
+            // unambiguously different than a 4-byte code.
+            0 if buf.len() == 12 && i > 4 => break,
+            b':' | b'\\' | 0..=0x20 | 0x7f..=0xff => {
+                write!(out, r"\x{:02x}", x).map_err(|_| ConversionError)?
+            }
```
leighmcculloch, line 46: Am I correct in thinking that this code will never get triggered because stellar-core never allows for the use of an asset that is a 12-byte with less than 5 non-null characters?

If I am incorrect in thinking that, then we probably have a second bug in that the token wrap method is allowing instantiation of contracts that wrap invalid assets.

leighmcculloch (approved):

soroban-env-host/src/native_contract/token/metadata.rs
graydon: Line 43? It stops the loop at the first null after char 4 (the 5th char). I.e. if you have a 12-char code with 5 leading non-null chars like "ABCDE\0\0\0\0\0\0\0", the 5 chars from buf[0] through buf[4] will get converted, and then buf[5] which is null will be seen by line 43 of the loop and terminate the loop with just "ABCDE" in the result.

I'm not sure if that answers your question. It's not dead code as far as I know!

graydon: (The reason it's conditionalized on 5 is so that a _shorter_ code like "ABC\0\0\0\0\0\0\0\0\0" will get converted to "ABC\x00\x00" rather than "ABC". This is how txrep appears to be differentiating literal 12-char codes from literal 4-char codes)

leighmcculloch: My question is does the contract allow creating a token with "ABC\0\0" because my understanding is it shouldn't be possible to create a AssetCode12 asset with null bytes on Stellar.

leighmcculloch: i.e. Does Soroban let us create contract instances for assets that can't exist on Stellar Classic.

graydon: Oh, I have no idea. I'm just saying even if "ABC\0\0" isn't a legal asset, that line of code will get triggered when rendering "ABCDE" which is legal. The code's not dead. Or were you asking about rendering escaped nulls specifically, that they can never happen on a 12-char code? That might be true. The escape-rendering line is executed for lots of reasons though: trailing nulls in 4-char codes as well as any non-nulls outside the printable range allowed.

Going to your question about allowed assets: it looks vaguely to me like [a 12-char asset "ABC\0\0" is allowed in soroban](https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/native_contract/token/contract.rs#L113-L182) but I don't really understand the setup path -- who can call that and how it gets validated before or after.  I guess you're looking for a check somewhere that mirrors [this code in core](https://github.com/stellar/stellar-core/blob/master/src/util/types.cpp#L184-L210)? I'm not sure if we have any in soroban, it's a few layers up from code I'm familiar with.

leighmcculloch: I'll run a test of this to verify if we have an inconsistency in Soroban and open an issue if so. Thanks!

PR #1044, opened 2023-09-07
Graydon code review

graydon:
Steps in code review so far:

- Rename some variables
- Move code around for ease of review, add review markers
- Equip `Env::Error` with `: Into<crate::Error>` so that we can pass `HostError::error` through from Env conversions
- Do so in bytes.rs
- Equip `Env::Error` with `: From<crate::Error>` so that we can project `Error` into it in certain cases
- Do so in preference to an unconditional `unreachable!()` in convert, which can panic
- Fix some code in convert that does two compares back to back when walking a tuple instead of scrutinizing one compare against two cases
- Change a bunch more uses of ConversionError to Error (some need minor SDK changes, see https://github.com/stellar/rs-soroban-sdk/pull/1083)

### DIFF
--- a/soroban-env-common/src/bytes.rs
+++ b/soroban-env-common/src/bytes.rs
@@ -1,5 +1,7 @@
 use crate::{
-    declare_tag_based_object_wrapper, ConversionError, Env, Error, TryFromVal, TryIntoVal, Val,
+    declare_tag_based_object_wrapper,
+    xdr::{ScErrorCode, ScErrorType},
+    Env, Error, TryFromVal, TryIntoVal, Val,
 };
 
 declare_tag_based_object_wrapper!(BytesObject);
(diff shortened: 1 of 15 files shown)


soroban-env-common/src/bytes.rs
```diff
@@ -8,14 +10,17 @@ impl<E: Env, const N: usize> TryFromVal<E, BytesObject> for [u8; N] {
     type Error = Error;
 
     fn try_from_val(env: &E, val: &BytesObject) -> Result<Self, Self::Error> {
-        let len: u32 = env.bytes_len(*val).map_err(|_| ConversionError)?.into();
+        let len: u32 = env.bytes_len(*val).map_err(Into::into)?.into();
```
leighmcculloch, line 13: Is this changing the error being mapped to? I can't tell what `Into::into` is converting to.

soroban-env-common/src/error.rs
```diff
@@ -163,6 +163,14 @@ impl From<stellar_xdr::Error> for Error {
     }
 }
 
+// This never happens, but it's needed for some impls of TryFromVal downstream
+// in the SDK that use the xdr::Error type.
+impl From<Error> for stellar_xdr::Error {
+    fn from(_value: Error) -> Self {
+        stellar_xdr::Error::Unsupported
```
leighmcculloch, line 170: If this never happens should it be `unreachable!()`?

leighmcculloch (approved):

soroban-env-common/src/bytes.rs
graydon: yes, it's converting to `Error` such that if we get a `HostError` saying something like "you ran out of gas" we're going to .into() it into the `Error` carried inside that `HostError`, rather than downgrading it to `ConversionError`.

soroban-env-common/src/error.rs
graydon: I _think_ it never happens but I don't think it's logically impossible, and I'm trying to avoid anything that can panic when there's a bug somewhere.

PR #1053, opened 2023-09-11
Tighten wasm interface version checks, and do on upload. Fix #1052.

### LINKED ISSUE #1052: do pre_release number check on upload / RPC
We should reject contracts with non-matching pre-release numbers earlier than just invocation -- also on upload or submission-to-the-RPC

graydon:
This fixes #1052 by rejecting contracts with unsupported versions during upload (they are already rejected during instantiation/execution). It also slightly tightens the set of allowed contracts, rejecting old-protocol ones with nonzero prerelease and rejecting all future protocol versions.

### DIFF
--- a/soroban-env-host/src/vm.rs
+++ b/soroban-env-host/src/vm.rs
@@ -75,6 +75,71 @@ pub struct VmFunction {
 }
 
 impl Vm {
+    fn check_contract_interface_version(
+        host: &Host,
+        interface_version: u64,
+    ) -> Result<(), HostError> {
+        let want_proto = {
+            let ledger_proto = host.get_ledger_protocol_version()?;
+            let env_proto = get_ledger_protocol_version(meta::INTERFACE_VERSION);
+            if ledger_proto <= env_proto {
+                // ledger proto should be before or equal to env proto
+                ledger_proto
+            } else {
+                return Err(err!(
+                    host,
+                    (ScErrorType::Context, ScErrorCode::InternalError),
+                    "ledger protocol number is ahead of supported env protocol number",
+                    ledger_proto,
+                    env_proto
+                ));
+            }
+        };
+
+        let got_pre = get_pre_release_version(interface_version);
+        let got_proto = get_ledger_protocol_version(interface_version);
+
+        if got_proto < want_proto {
+            // Old protocols are finalized, we only support contracts
+            // with similarly finalized (zero) prerelease numbers.
+            if got_pre != 0 {
+                return Err(err!(
+                    host,
+                    (ScErrorType::WasmVm, ScErrorCode::InvalidInput),
+                    "contract pre-release number for old protocol is nonzero",
+                    got_pre
+                ));
+            }
+        } else if got_proto == want_proto {
+            // Current protocol might have a nonzero prerelease number; we will
+            // allow it only if it matches the current prerelease exactly.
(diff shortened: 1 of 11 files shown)


soroban-env-host/src/vm.rs
```diff
@@ -75,6 +75,71 @@ pub struct VmFunction {
 }
 
 impl Vm {
+    fn check_contract_interface_version(
```
leighmcculloch, line 78: Does this function gate deployment, or only execution, of a contract?

soroban-env-host/src/vm.rs
```diff
@@ -75,6 +75,71 @@ pub struct VmFunction {
...
+                    want_pre
+                ));
+            }
+        } else {
+            // Future protocols we don't allow. It might be nice (in the sense
+            // of "allowing uploads of a future-protocol contract that will go
+            // live as soon as the network upgrades to it") but there's a risk
+            // that the "future" protocol semantics baked in to a contract
+            // differ from the final semantics chosen by the network, so to be
+            // conservative we avoid even allowing this.
```
leighmcculloch, line 132: This sounds not great to me. Are we preventing even upload of a contract that is for a future version? Will that mean a contract can't be deployed until the vote happens? Or until a version of stellar-core is shipped? I think we might be protecting the developer from something that they can easily footgun themselves other ways anyway (such as bundling a compiled wasm into another contract and deploying later).

leighmcculloch (commented):
One concern about gating upload.

soroban-env-host/src/vm.rs
graydon: both

soroban-env-host/src/vm.rs
graydon: Correct. If a protocol version is not an active version in the host, you can't upload a contract with it. This is the conservative result @MonsieurNicolas and I concluded would be safest. There are always more footguns but this excludes the a case that seems .. probably a mistake if you do it, and not especially important to support. The only thing you get out of being able to _upload_ before the vote is to have the contract sit there idle. It already won't run.

leighmcculloch: I think the network is taking on a role that belongs to tooling. We can very easily provide this functionality without making a protocol change by having tooling generate an error or warning if the version being uploaded is incompatible with the network. We already plan to add functionality similar to this to the cli:
- https://github.com/stellar/soroban-tools/issues/941

Will these checks at upload time also make upload more expensive?

leighmcculloch (approved):
RE: https://github.com/stellar/rs-soroban-env/pull/1053#discussion_r1322092436. I don't think we should do the verify the wasm file during upload for the reasons I mention in that thread, because tooling can easily do that and it doesn't have to be something built into the protocol.

But it sounds like that's a decision already agreed on and doesn't need to block this, so approving.

soroban-env-host/src/vm.rs
graydon: Sure, tooling can check too (in fact #1052 suggests RPC check this condition as well). This is just defense in depth.

Yes, it makes the upload use a bit more resources. No more than instantiating a VM, which is something we do on literally every subsequent call. Contract uploads will be far less frequent.

jayz22 (approved):

PR #1074, opened 2023-09-18
Attach the auth in recording mode to any valid tracker when possible.

dmkozh:
### What

Attach the auth in recording mode to any valid tracker when possible.

### Why

Currently only the tracker that matches with exact same object is used and otherwise a new tracker is created. This results in recording auth producing unnecessarily disjoint call trees.

This is kind of best effort and might lead to a bit weird edge-case scenarios in semantic sense (e.g. a tracker for an exact address object might be used by another instance of that address), but the end results should still fulfil the enforcing mode invariants and shouldn't compromise security.

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/auth.rs
+++ b/soroban-env-host/src/auth.rs
@@ -788,6 +788,10 @@ impl AuthorizationManager {
             AuthorizationMode::Enforcing => self.require_auth_enforcing(host, address, &function),
             // metering: free for recording
             AuthorizationMode::Recording(recording_info) => {
+                // First try to find the tracker for this exact address object.
+                // This is a best-effort heuristic to come up with a reasonably
+                // looking recording tree for cases when multiple instances of
+                // the same exact address are used.
                 let address_obj_handle = address.get_handle();
                 let existing_tracker_id = recording_info
                     .try_borrow_tracker_by_address_handle(host)?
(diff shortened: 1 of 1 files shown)


soroban-env-host/src/auth.rs
```diff
@@ -788,6 +788,10 @@ impl AuthorizationManager {
             AuthorizationMode::Enforcing => self.require_auth_enforcing(host, address, &function),
             // metering: free for recording
             AuthorizationMode::Recording(recording_info) => {
+                // First try to find the tracker for this exact address object.
```
leighmcculloch, line 791: What is the meaning of "exact"? What's the distinction between an exactly the same, and a matching address object?

soroban-env-host/src/auth.rs
```diff
@@ -821,6 +825,35 @@ impl AuthorizationManager {
                         ));
                     }
                 }
+                // If there is no active tracker for this exact address object,
+                // try to find any matching active tracker for the address.
+                for tracker in self.try_borrow_account_trackers(host)?.iter() {
+                    if let Ok(mut tracker) = tracker.try_borrow_mut() {
+                        if !host.compare(&tracker.address, &address)?.is_eq() {
```
leighmcculloch, line 832: Is it possible to add a test for the scenario being fixed?

soroban-env-host/src/auth.rs
dmkozh: Exactly the same == exactly the same `Val` (i.e. the same host object). Matching == same value. Same `Val` takes precedence to help with the cases where `Address` is passed around (which should be probably majority of the cases). This only matters for complex cases where multiple addresses are involved and duplicates are possible.

soroban-env-host/src/auth.rs
dmkozh: Yes, but that's more involved and I don't want to block RC2 PR. I'm working on tests for main though.

soroban-env-host/src/auth.rs
leighmcculloch: Got it, exact is same handle, and matching is same address held in a different host object, because the values aren't interned? Got it.

leighmcculloch (approved):

soroban-env-host/src/auth.rs
leighmcculloch: If the tests are soon, we'll have less churn if we make sure this works with them before starting the release process.

PR #1082, opened 2023-09-22
Error improvements

dmkozh:
### What

- Narrow down the `try_call` return value to a single host error kind for the sake of reducing the backwards compatibility risks.
- Convert more wasmi errors to the appropriate SC errors
- Log the exact wasmi trap code
- Cleanup the code a bit to use `error` helper instead of `err_diagnostics`

### Why

More flexibility/better logging

### Known limitations

N/A


leighmcculloch (approved):

soroban-env-host/src/vm.rs
```diff
@@ -347,7 +347,14 @@ impl Vm {
             match e {
                 wasmi::Error::Trap(trap) => {
                     if let Some(code) = trap.trap_code() {
-                        return Err(code.into());
+                        let err = code.into();
+                        return Err(if host.is_debug()? {
+                            // With diagnostics on: log as much detail as we can from wasmi.
+                            let msg = format!("VM call trapped: {:?}", &code);
+                            host.error(err, &msg, &[func_sym.to_val()])
```
graydon, line 354:
```suggestion
                            host.error(err, "VM call trapped", &[func_sym.to_val(), code.to_val()])
```

graydon (changes requested):
Looks great! Thanks. Just one minor change.

soroban-env-host/src/vm.rs
dmkozh: Should be `err_to_val()` instead, updated.

graydon (approved):

Issue #1087, opened 2023-09-27
ComputeEcdsaSecp256k1Key is never used


graydon:
The interface we settled on for Secp256k1 is the "key recovery" interface which doesn't involve a user ever converting a bytes object directly to a public key. This cost is therefore never used (and the codepaths that do it are only used to calibrate the cost, never actually do it).

It should probably be removed from the XDR cost types.

jayz22:
Xdr change has been merged. Env change will be made with https://github.com/stellar/rs-soroban-env/pull/1105.

PR #1088, opened 2023-09-27
More code review and cleanup of host.rs module

graydon:
Today I didn't get as much done -- had some errands and a doctor appointment -- but my goal was to get through the residual bits of host.rs and I did that at least. Along the way I removed the `allows` that enabled un-warned dead code and unused variables, and found some stuff:

  - Two missing error checks on linear-memory "unpack" vector lengths. These don't represent a _major_ vulnerability, but they do allow users to do an unpack to a mismatched vector length, which could hide a bug. A footgun at least. Fixed.
  - A dead cost type, based on a previous iteration of the secp256k1 interface. Filed #1087 to track removing it.
  - Some diagnositcs weren't using the names we so meticulously pass down to them. Just throwing them out. Fixed.
  - There was a bunch of dead code and over-public interfaces. Fixed.


soroban-test-wasms/wasm-workspace/linear_memory/src/lib.rs
```diff
@@ -27,4 +28,28 @@ impl Contract {
...
+        Ok(())
+    }
+
+    // Same but with a length mismatch
+    pub fn vec_mem_bad(e: Env) -> Result<(), Error> {
+        let in_buf: [Val; 3] = [Val::from(1), Val::from(2), Val::from(3)];
+        let mut out_buf: [Val; 2] = [Val::from(0); 2];
+        let vec: VecObject = e.vec_new_from_slice(&in_buf)?;
+        e.vec_unpack_to_slice(vec, &mut out_buf)?;
+        // Should never get to these lines.
```
dmkozh, line 50: This is confusing, why not assert that the call above is error and remove the statements below?

dmkozh (approved):

soroban-test-wasms/wasm-workspace/linear_memory/src/lib.rs
graydon: fair, fixed

Issue #1089, opened 2023-10-03
Calling `update_current_contract_wasm` from contract in test doesn't work


leighmcculloch:
### What version are you using?

`v20.0.0-rc2`

### What did you do?

1. Called `update_current_contract_wasm` from a contract in a Rust test that deployed a new version of the contract that had a new function
2. Tried to call the updated contract's new function.

### What did you expect to see?

I expected to be able to call the updated wasm's new function.

### What did you see instead?

An error indicating that the function didn't exist, as if the update hadn't worked.

leighmcculloch:
I also noticed that if I store the ledger state as a snapshot by calling `Env::to_snapshot_file` before and after calling `update_current_contract_wasm` the executable hash for the contract does not change.

leighmcculloch:
It seems that calling `update_current_contract_wasm` from a contract on a network also doesn't work. An event is outputted in the TransactionMeta indicating that the executable is being updated, but subsequent attempts to invoke the updated contract indicate the function doesn't exist, and using steexp.com show that the old wasm is still active.

Here's an example:

Contract:
https://testnet.steexp.com/contract/CDNG6Y4YQD47VV3377AT2BF3V6ZHY3TYORSQ2CZQKYM43XHKAF7VTGYM

Tx that calls a function on it, that causes update to occur:
https://horizon-testnet.stellar.org/transactions/ccc24ed173c2e1a90d0ace53cdb6915c3d693812bc0ebcfb4cb29ed268e5cc09

Tx's meta:
```console
[...]

PR #1090, opened 2023-10-03
Fix the instance storage update logic.

dmkozh:
### What

Instead of fetching the current executable and overwriting the entry, only the relevant field has to be updated. I've changed the interface to make sure we have the correct logic everywhere.

### Why

Bug fix: https://github.com/stellar/rs-soroban-env/issues/1089

### Known limitations

N/A


soroban-env-host/src/host/frame.rs
```diff
@@ -736,39 +736,17 @@ impl Host {
...
-                            "unexpected storage for host fn",
-                            &[],
-                        ))
-                    }
-                    Frame::Token(_, _, _, instance) => instance.executable.metered_clone(self)?,
-                    // Mock executable for test contract 'instances'. This is
-                    // just a placeholder - it's not used for actually calling
-                    // the test contracts.
-                    #[cfg(any(test, feature = "testutils"))]
-                    Frame::TestContract(t) => t.instance.executable.clone(),
```
leighmcculloch, line 761: Where are the token and test contract cases handled now? I don't see where.

leighmcculloch (approved):
One question otherwise lgtm.

soroban-env-host/src/host/frame.rs
dmkozh: We don't need to handle these cases at all because we don't care what's the *current* executable. We just need to preserve the executable in the instance entry, which we do by passing `None` into `store_contract_instance`.

PR #1091, opened 2023-10-03
Feature: Expose `rem_euclid` functions on 256 bit numbers

mootz12:
### What

Add `rem_euclid` host functions to both the `I256` and `U256` bit numbers. 

### Why

Fixed point math implementations need to compute a remainder to round results properly. Euclidean remainder chosen for consistent positive results, but I can make the standard one work if that would be more fitting.

See ->
OpenZepplin: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol#L207
Soroban verison: https://github.com/mootz12/soroban-fixed-point-math

### Known limitations

At the moment, I can only support "forced panic" fixed point math with `I256` and `U256` (host forces panic if something overflows during a host call to, for example, `i256_add`).

This can be worked around by changing the way errors are handled for bignum host fn calls here: https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/host/num.rs#L28

I don't imagine people will want to manually handle phantom overflows in contracts anyway, but, there might be some fuzzing implications if it's hard to discern there overflows from other errors.


mootz12:
SDK Implementation here: https://github.com/stellar/rs-soroban-sdk/pull/1102

soroban-env-common/env.json
```diff
@@ -711,7 +711,39 @@
                     ],
                     "return": "u64",
                     "docs": "Convert a `Duration` object a `u64`."
-                }                
+                },
+                {
```
dmkozh, line 715: Could you please move these near the respective u256/i256 function declarations? We could even change the exports given that this is going to be a 'breaking' release anyway.

jayz22:
Can you add some tests in https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/test/num.rs. Doesn't have to cover every corner case but just making sure the operation is covered.

graydon:
- Why did example_contract_data.wasm shrink by 80%? Is that a mistake?
- When you say "forced panic", you mean a _guest_ panic (a wasm trap)? I think that's mostly how we want to handle arithmetic errors, yeah, though we _could_ add special `wrapping_`, `checked_` or `saturating_` variants if guests want finer control. Default shouldn't be something that's ignorable.

mootz12:
> * Why did example_contract_data.wasm shrink by 80%? Is that a mistake?

Looks like it lost 800 bytes (from ~4800 to ~4000). I'm guessing the 80% number on the diff is the amount left unchanged?

Inspecting the WAT files doesn't really make it clear (to me) what optimization occurred, but the WASM looks to be well tested here: https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/test/storage.rs#L316

> * When you say "forced panic", you mean a _guest_ panic (a wasm trap)? I think that's mostly how we want to handle arithmetic errors, yeah, though we _could_ add special `wrapping_`, `checked_` or `saturating_` variants if guests want finer control. Default shouldn't be something that's ignorable.

I believe so, yeah. Basically if any overflow occurs during the calculation things would panic. There is a discord thread with a bit more context: https://discord.com/channels/897514728459468821/1158458172181319841

dmkozh (approved):

dmkozh:
This looks good to me, thanks for your contribution!

PR #1095, opened 2023-10-05
Properly dispatch test contracts based on the executable.

dmkozh:
### What

Properly dispatch test contracts based on the executable.

Current implementation just ignores `update_current_contract_wasm` calls because the test contracts are prioritized unconditionally.

### Why

Bug fix

### Known limitations

N/A


soroban-env-host/src/host/data_helper.rs
```diff
@@ -569,4 +570,28 @@ impl Host {
...
+    // that marks contracts created with `register_test_contract`.
+    // "testutils" are not covered by budget metering.
+    pub(crate) fn is_test_contract_executable(
+        &self,
+        contract_id: &Hash,
+    ) -> Result<bool, HostError> {
+        let key = self.contract_instance_ledger_key(contract_id)?;
+        let instance = self.retrieve_contract_instance_from_storage(&key)?;
+        let test_contract_executable = ContractExecutable::Wasm(
+            crypto::sha256_hash_from_bytes(&[], self)?
```
leighmcculloch, line 584: ❗ Am I understanding correctly that the marker for whether a contract is a test contract the hash of empty bytes? It'd be easier to eyeball it in test data if the value was simply `[0u8; 32]`. Is there a reason it needs to be an actual hash of a known preimage?

leighmcculloch: 💡 We could make it a real enum value in the XDR, i.e. `ContractExecutable::External`? That feels like it'd be less of a hack.

```c
enum ContractExecutableType
{,
    CONTRACT_EXECUTABLE_EXTERNAL = -1,
    CONTRACT_EXECUTABLE_WASM = 0,
    CONTRACT_EXECUTABLE_TOKEN = 1
};

union ContractExecutable switch (ContractExecutableType type)
{
case CONTRACT_EXECUTABLE_EXTERNAL:
    void;
case CONTRACT_EXECUTABLE_WASM:
    Hash wasm_hash;
case CONTRACT_EXECUTABLE_TOKEN:
    void;
};
```

There's no op or host function for uploading none, so it'd be not seen in pubnet, but the type system would echo a reality where envs can support arbitrary other types of contract executables that aren't stored in the ledger.

soroban-env-host/src/host/frame.rs
```diff
@@ -525,9 +525,13 @@ impl Host {
...
 
+        // Try dispatching the contract to the compiled-in registred
+        // implmentation. Only the contracts with the special (empty) executable
+        // are dispatched in this way, so that it's possible to switch the
+        // compiled-in implementation back to Wasm via
+        // `update_current_contract_wasm`.
         // "testutils" is not covered by budget metering.
-        #[cfg(any(test, feature = "testutils"))]
-        {
+        if cfg!(any(test, feature = "testutils")) && self.is_test_contract_executable(id)? {
```
leighmcculloch, line 534: ❗ I believe this change, using `cfg!` with an `if` block instead of the attribute on the block, will result in the block being compiled for both test and non-test builds. There might not be a consequence of that, but given that it could be a reasonable assumption that no test code ever end up in a production build, I think it's worth keeping it an attribute on the block so it is never compiled on prod builds.

leighmcculloch (commented):
Some thoughts inline.

soroban-env-host/src/host/frame.rs
dmkozh: reverted

soroban-env-host/src/host/data_helper.rs
leighmcculloch: What happens if someone uploads an empty contract? Will it think it's a test contract? Given the code has to handle this case, why do we even need a marker? Can't the presence of a registered test contract simply override whatever is on ledger? And then if a contract is updated, we can delete the registered test contract?

leighmcculloch (approved):

Issue #1020, opened 2023-08-28
Run cpu instruction calibration on a variety of hardware


jayz22:
### What
Calibrate the cpu instructions on a variety of hardwares that the validators run on. 

### Why
The metering model is deterministic across all nodes. The model is currently calibrated on a single machine (M1) which may vary from actual hardware that validators use. This can make the actual compute time vary for the same amount of CPU instructions, which could affect ledger close time. The network resource limits need to be set conservatively w.r.t the worst case. We need to calibrate them on various hardware architecture in order to figure out the correct bonds.

graydon:
I talked to @anupsdf about this and we concluded two points:

1. The key question isn't how many "model CPU instructions" a contract takes, it's how much _time_, and so we really will want to set our _network limits_ by reasoning backwards from the _observed_ virtual-instructions-per-unit-real-time value we see in the network, empirically, on the nodes we're running (which, as you mention, are likely to be different _enough_ from our workstations to warrant empirical observation). This is fine, but it means that the actual instruction count values are mostly irrelevant. They're just a term in an equation we divide out to get the number we set the limit to. Eg. if the network says it's processing 20 virtual instructions per nanosecond and we want to limit contracts to 1ms, then we set the instruction limit to 20m instructions. But if it says 5 instructions per ns, we set the limit to 5m instructions. The actual "instructions number" doesn't matter to setting a "time target".
2. That said, it's a little confusing to _think_ about and might be _misleading_ to users to see "virtual instruction counts" that are much higher than the (unknown but plausibly estimatable) true instruction count for the machine they're on. So for sake of not-confusing people, it'd be good to calibrate the _model instruction counts_ to their values as measured on x86-64 machines, because we expect most validators to be on that arch.

So .. I'm going to take this and just run calibration on the x86-64 machine I have here. Doesn't matter what its clock frequency is, we're only talking instruction counts of the cost centers.

graydon:
Some investigation and results here (I meant to discuss this with @jayz22 but I'll make a note here for future reference too):

- x64 actually gives very incorrect-seeming numbers when I try to calibrate
- I think _part_ of what's happening is that for memcpy (and mempcy-like cost centers, of which there are a few) we pass large buffers through to rust's slice-copying code which will engage a throughput-oriented fast path for memory copies -- using AVX2 instructions but with a relatively high setup/teardown overhead for the process.
[...]

PR #1098, opened 2023-10-06
Budget fiddling

graydon:
A bunch of exploration of minor issues discovered in budget calibration, ostensibly about #1020 but also involving some tracy markup, some internal cleanup, machinery to allow excluding the overwhelming VM instantiation cost center, and some attempts at setting budget costs more from first principles.

### DIFF
--- a/soroban-env-host/benches/common/measure.rs
+++ b/soroban-env-host/benches/common/measure.rs
@@ -1,11 +1,12 @@
 use rand::{rngs::StdRng, Rng, SeedableRng};
 use soroban_bench_utils::{tracking_allocator::AllocationGroupToken, HostTracker};
+use soroban_env_common::xdr::ContractCostType;
 use soroban_env_host::{
     budget::{AsBudget, COST_MODEL_LIN_TERM_SCALE_BITS},
     cost_runner::CostRunner,
     Host,
 };
-use std::ops::Range;
+use std::{io, ops::Range};
 use tabwriter::{Alignment, TabWriter};
 
 use super::{fit_model, FPCostModel};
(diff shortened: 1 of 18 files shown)


soroban-env-host/benches/common/measure.rs
```diff
@@ -27,6 +28,31 @@ pub struct Measurements {
...
+    ) -> Result<(), io::Error> {
+        let max = self.measurements.iter().map(&f).max().unwrap_or_default();
+        let base = f(&self.baseline);
+        if max < base * 10 {
+            println!("max {meas} measurement for {cost} is {max} which is less than 10x baseline {base}, try higher iteration or step size");
+            Err(io::ErrorKind::InvalidData.into())
+        } else {
+            Ok(())
+        }
+    }
```
jayz22, line 47: This is a nice check!

soroban-env-host/benches/common/measure.rs
```diff
@@ -318,10 +344,12 @@ where
         &mut Vec<<<HCM as HostCostMeasurement>::Runner as CostRunner>::RecycledType>,
     ),
 {
+    assert!(HCM::STEP_SIZE >= (1 << COST_MODEL_LIN_TERM_SCALE_BITS));
```
jayz22, line 347: Nice check here!

soroban-env-host/src/budget.rs
```diff
@@ -520,13 +536,26 @@ impl Default for BudgetImpl {
...
+                // overhead and a small per-byte cost. But large buffers aren't
+                // really how byte-copies usually get used in metered code. Most
+                // calls have to do with small copies of a few tens or hundreds
+                // of bytes. So instead we just "reason it out": we can probably
+                // copy 8 bytes per instruction on a 64-bit machine, and that
+                // therefore a 1-byte copy is considered 1/8th of an
+                // instruction. We also add in a nonzero constant overhead, to
+                // avoid having anything that can be zero cost and approximate
+                // whatever function call, arg-shuffling, spills, reloads or
+                // other flotsam accumulates around a typical memory copy.
```
jayz22, line 551: I've been thinking about consolidating some of those cost types, and put some thoughts in https://github.com/stellar/rs-soroban-env/issues/1020#issuecomment-1749621087

Some of these models have different linear coefficients, e.g. `HostMemCpy` vs. `HostMemCmp` which can be reasoned, and `ValSer` and `ValDeser` which is harder to reason about. 
Also they have different const coefficients, which will probably need to come from measure. 
So I was thinking a combination of measurement and reasoning to determine the proper overhead and scaling factor for the analytical model maybe the right thing to do? 
It can be a followup, don't want to block this PR. 

Another way that could work is in the calibration, copy/compare a small bytes array N times, instead of on one big array. This could avoid the AVX optimization? Personally I still prefer the analytical approach.

soroban-env-host/src/budget.rs
```diff
@@ -250,15 +242,35 @@ impl BudgetDimension {
...
+            let _span = tracy_span!("charge");
+            _span.emit_text(ty.name());
+            _span.emit_value(amount);
+        }
+
+        if _is_cpu {
+            // Budget charging _itself_ has a nonzero overhead, on the
+            // order of 100ns or maybe 250 instructions.
+            const BUDGET_CHARGE_CPU_OVERHEAD: u64 = 250;
+            amount = amount.saturating_add(BUDGET_CHARGE_CPU_OVERHEAD);
```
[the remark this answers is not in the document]
graydon: I thought of this, but then I also thought that part of why we got rid of it is that it's a bit .. recursive? like if you were to start counting the couple-hundred instructions spent, probably a lot of them are in the steps _before_ we get to this line: updating the tracker, fetching the CPU cost model and evaluating it (and then doing that again for memory), and so if we did all that a _second time_ just to "charge the cost of mesuring costs" we'd be incurring those costs again, and would need to "charge the cost of measuring costs" again, so we'd have to actually set that cost to _twice_ the measurement cost. Since this cost is not entirely negligible, I figured it might be better to just inline it inside here, where it's going to be much smaller, probably only 2-5 instructions (a branch, a saturating add, a store) and only apply to the CPU path.

soroban-env-host/src/budget.rs
graydon: Yeah, we can continue to iterate on this. I don't think this PR is the final word on the question; just what came to mind as I was doing review. I wanted to have a look at x86 (done so!) and do some eyeball-checks with tracy (done so!) and once I did those I wanted to shake some of the most obvious quirks out of what I found (memcpy numbers that don't make sense, plus the issue of overcounting VM setup obscuring undercounting the rest).

graydon:
@jayz22 on second thought I reverted the hard-wired constant measurement overhead; it breaks tests and does so in a way that suggests why it's wrong: it breaks the implicit contract that "costs are reflected by the cost schedule only". Maybe we can just give everything in the cost schedule slightly higher constant terms to cover the measurement overhead.

dmkozh (approved):

PR #1109, opened 2023-10-13
Setup bench framework to run experiments

### LINKED ISSUE #1030: confirm worst-case calibration of ed25519 operations
A year ago, in 7367f00801462667a2b695be38a2ab7792d114be, I added a calibration for Ed25519 operations that was based on a range of possible scalars fed into a Curve25519 operation, such as the one that an Ed25519 signature verification does. This is visible in the `EdwardsPointCurve25519ScalarMulRun` calibration routine. Later in March of this year this was removed along with the `EdwardsPointCurve25519ScalarMul` cost type, in 349dab8a42c9db5a3ecfbc5e4323910c26d52565. The idea I think of adding it was to use it to charge more-carefully for the cost of an Ed25519 signature verification, somehow?

Anyway, looking at the git history, it looks like this finer-grained cost type was never actually _used_ when charging for Ed25519 signature verification. This _might_ be ok, or it might not be. It depends on the variability of signature verification. It _is_ a variable-time operation: it varies by the size of the scalar component. Which I think is usually "a randomly distributed 256-bit number" so perhaps 50%-likely to have its high bit set, and _presumably_ that might hit the worst, or at least a dominant-cost case? But I'm not sure. It might be the case that the worst-case scalar multiplication cost is abnormally high -- higher than the average or whatever the calibration code is testing -- and users could cause us to undercharge if they submit this worst case. This should be investigated.

jayz22:
### What

Setup bench framework to run experiments 
- Experimental bench code is located in `experimental` (`benches/common` and `src/cost_runner` dirs)
- Aggregate `ContractCostType`, `WasmInsnType` and a new  **`ExperimentalCostType`** into top level `CostType`

Resolves #1030 
- Bring back `EdwardsPointCurve25519ScalarMul` as a experimental cost type and run variation histogram on it. 

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1376,6 +1376,7 @@ version = "20.0.0-rc2"
 dependencies = [
  "backtrace",
  "bytes-lit",
+ "curve25519-dalek",
  "ed25519-dalek",
  "expect-test",
  "getrandom",
--- a/soroban-env-host/Cargo.toml
+++ b/soroban-env-host/Cargo.toml
@@ -32,6 +32,9 @@ k256 = {version = "0.13.1", features=["ecdsa", "arithmetic"]}
 # is needed to build the host for wasm (a rare but supported config).
 getrandom = { version = "0.2", features=["js"] }
 sha3 = "0.10.8"
+# NB: this must match the same curve25519-dalek version used by ed25519-dalek above
+# used only for calibration
+curve25519-dalek = { version = "4", default-features = false, features = ["digest"]}
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 tracy-client = { version = "=0.15.2", features = ["enable", "timer-fallback"], default-features = false, optional = true }
--- a/soroban-env-host/benches/common/experimental/ed25519_scalar_mul.rs
+++ b/soroban-env-host/benches/common/experimental/ed25519_scalar_mul.rs
@@ -1,0 +1,62 @@
+use crate::common::HostCostMeasurement;
+use curve25519_dalek::{constants, edwards, scalar};
+use rand::{rngs::StdRng, Rng, RngCore};
+use soroban_env_host::{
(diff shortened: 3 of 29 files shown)


jayz22:
Calibration result shows both `EdwardsPointCurve25519ScalarMul` and `VerifyEd25519Sig` have a large variation (`max/min` = 14 and 100 respectively, see screenshot), which confirms @graydon's concern in https://github.com/stellar/rs-soroban-env/issues/1030 

Screen Shot 2023-10-12 at 6 43 45 PM

Will next figure out how to address it.

jayz22:
Upon closer examination, I change my conclusion above. The worst case vs the average case is actually pretty close. 

I realize the above experiment isn't really testing what we are looking for. We are trying to answer the question "is our calibration, which is based on the average case, underestimating the worst case".
 Here are some tweaks to the setup in order to better answer that question:
1. Throw away the "best" case. For `EdwardsPointCurve25519ScalarMul` (doesn't matter too much for `VerifyEd25519Sig `), the best case is significantly faster than the average, but it's rare and doesn't contribute to the problem statement above. 
2. Fix the input size. For `VerifyEd25519Sig` that is the message size. The variation analysis should fix the input size and examine the effect of random bytes in the input. (Longer message will cost more to verify sig but that's not what we are interested in discovering)
3. Increase the iteration count from 100->1000

With the above tweaks, there are the results:
Screen Shot 2023-10-13 at 1 09 32 PM

The ration between worst/best case for `EdwardsPointCurve25519ScalarMul` is 1.34. 
For `VerifyEd25519Sig` there is no trivial way to set the worst case, so everything is random. With 1000 sample points, the ratio between max/min is just 1.008. 

So I conclude the worst case of ed25519 is not significantly worst than the average, thus no need to tweak our calibration setup or numbers.

graydon (approved):
Looks good! Thanks for investigating

PR #1112, opened 2023-10-13
Decorate error for nonce missing from the footprint.

dmkozh:
### What

Decorate error for nonce missing from the footprint.

### Why

Improving error diagnostics

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/auth.rs
+++ b/soroban-env-host/src/auth.rs
@@ -1704,7 +1704,7 @@ impl AccountAuthorizationTracker {
     fn snapshot(&self, budget: &Budget) -> Result<AccountAuthorizationTrackerSnapshot, HostError> {
         Ok(AccountAuthorizationTrackerSnapshot {
             invocation_tracker_root_snapshot: self.invocation_tracker.snapshot(budget)?,
-            // We need to snapshot authenctioncation and nocne-related fields as
+            // We need to snapshot authentication and nonce-related fields as
             // during the rollback the nonce might be 'un-consumed' during
             // the storage rollback. We don't snapshot `is_valid` since this is
             // not recoverable and nonce is never consumed for invalid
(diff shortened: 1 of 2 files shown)


graydon:
I think this is _plausible_ but it gets at something I've been thinking of a few times during review: we have a few subsystems (storage, metered maps and such) that take a `Budget` rather than `Host` (in order to enable working with them before/after `Host` lifecycle is complete, while still metering things) but by doing that we lose the ability to record detailed causes of failure in those context the way we do with `Host::error`. I was thinking it might make sense to turn some of those methods into `impl AsBudget`, and extend the `AsBudget` trait to include some of the error-creation-and-reporting functions in `Host` so that when we _do_ call something from a `Host` context, we _can_ record the extra context. The default methods of `AsBudget` (or definitions in `impl AsBudget for Budget`) could just construct and return the error without decoration. WDYT?

dmkozh:
> WDYT?

As per offline discussion, it would be generally nicer than returning event-less `ScError`s, but decoration is still required in some cases in order to provide more rich and contextual message and meaningful context (e.g. in this particular case there is no way to convert the `LedgerKey` missing from the footprint to `Val`, thus we explicitly state that nonce entry is missing in the error message).

graydon (approved):

leighmcculloch:
I think it's concerning that we can have errors that don't feed into diagnostic events. I understand we might miss decorating errors with additional information in some places, but is there a way to restructure the errors such that regardless of whether they get decorated or not, the most basic information about them still end up in the diagnostic events?

graydon:
@leighmcculloch Errors are generated in contexts where there is no diagnostic event buffer to record them into, so not "in general", but we can do a little better in a few contexts, probably. We'll keep iterating on this, the set of diagnostic events can improve over time without breaking anything.

PR #1114, opened 2023-10-14
Remove bad LEs

graydon:
This comes out of some review on compare.rs: we were supporting a lot of LE types in the code that are not ever supposed to occur. It's better to treat the presence of those cases as internal errors than have dead code "supporting" them possibly incorrectly.

Also eliminates comparison of LEs altogether. They were only being compared because of older constraints on MeteredOrdMap wanting its value type to be comparable. That's not necessary for LEs (in the Storage map), only Vals (in HostMaps) in order to have Map be ordered itself (because Val must be ordered). Turns out we'd already relaxed the requirement on MeteredOrdMap's value type to not have to be comparable. So all the LE-ordering code was dead anyway, which is good because they were all being treated as calls to Ord on fixed size types, but they weren't all fixed size types.

Also fixed a missing case in metered_clone, where we weren't charging for substructure of AccountEntries.

I'm not sure this is all that needs to be done for these files. I'm just out of time for tonight. Will continue reviewing on monday but figured I should get this started now.

### DIFF
--- a/soroban-env-host/src/storage.rs
+++ b/soroban-env-host/src/storage.rs
@@ -164,7 +164,7 @@ impl Storage {
         match le.data {
             Account(_) | Trustline(_) | ContractData(_) | ContractCode(_) => Ok(()),
             Offer(_) | Data(_) | ClaimableBalance(_) | LiquidityPool(_) | ConfigSetting(_)
-            | Ttl(_) => Err((ScErrorType::Storage, ScErrorCode::InvalidInput).into()),
+            | Ttl(_) => Err((ScErrorType::Storage, ScErrorCode::InternalError).into()),
         }
     }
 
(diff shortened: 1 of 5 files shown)


soroban-env-host/src/storage.rs
```diff
@@ -154,6 +154,34 @@ pub struct Storage {
...
+    /// by Soroban: accounts, trustlines, contract code and data. The rest are
+    /// never used by stellar-core when interacting with the Soroban host, nor
+    /// does the Soroban host ever generate any. Therefore the storage system
+    /// will reject them with [ScErrorCode::InternalError] if they ever occur.
+    pub fn check_supported_ledger_entry_type(le: &LedgerEntry) -> Result<(), HostError> {
+        use crate::xdr::LedgerEntryData::*;
+        match le.data {
+            Account(_) | Trustline(_) | ContractData(_) | ContractCode(_) => Ok(()),
+            Offer(_) | Data(_) | ClaimableBalance(_) | LiquidityPool(_) | ConfigSetting(_)
+            | Ttl(_) => Err((ScErrorType::Storage, ScErrorCode::InvalidInput).into()),
```
dmkozh, line 167: We don't expect invalid entries to get into host, do we? So maybe internal error would be more appropriate. I don't have a strong opinion though.

dmkozh (approved):

jayz22 (approved):

PR #1120, opened 2023-10-19
Review fixes for declared_size and metered_clone

graydon:
Fair amount of churn here:

- Removed several dead impls, cleaned up comments, added missing testcases, reordered stuff for consistency
- Reclassified and removed incorrectly-classified shallow types. This is fairly serious: we weren't counting the substructure cloning cost of much of the auth system as well as the wasm code of a contractcode entry, or the signers of an account entry.

One interesting part of this is that, as far as I can tell, everything in the events system doesn't use `metered_clone` anymore at all. I think this is good because it previously _errored_ when trying to `metered_clone` a diagnostic event, which is "just as observable" as `metered_cloning`-something-that-should-be-free would have been. Luckily it looks like we only "accumulate events", don't clone them, and then externalize them when done, charging the cost of emitting them outside and ignoring the cost of externalizing. I think! @jayz22 and @sisuresh I would appreciate a second pair of eyes on this understanding.

### DIFF
--- a/soroban-env-host/src/events/internal.rs
+++ b/soroban-env-host/src/events/internal.rs
@@ -43,9 +43,9 @@ impl InternalContractEvent {
 /// Internal representation of a structured debug message which is logically
 /// similar to an InternalContractEvent (and will project to a ContractEvent the
 /// same way when we externalize it to XDR) but internally is different: it
-/// stores "plain" rust data values -- String and Vec as well as XDR types like
-/// ScVal -- capturing the input to diagnostics, rather than interacting with
-/// the host object system. It does this for two reasons:
+/// stores more "plain" rust data values -- Vec as well as XDR types like ScVal
+/// -- capturing the input to diagnostics, rather than interacting with the host
+/// object system. It does this for two reasons:
 ///
 ///   1. to avoid to avoid perturbing object numbering when debug-mode is
 ///      enabled -- this could potentially cause hosts running on watcher nodes
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -821,7 +821,7 @@ impl VmCallerEnv for Host {
         Ok(self.max_live_until_ledger()?.into())
     }
 
-    // endregion "context" module functions
+    // endregion: "context" module functions
 
     // region: "int" module functions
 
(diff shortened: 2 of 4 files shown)


jayz22 (approved):
> Fair amount of churn here:
> 
> Removed several dead impls, cleaned up comments, added missing testcases, reordered stuff for consistency
> Reclassified and removed incorrectly-classified shallow types. This is fairly serious: we weren't counting the substructure cloning cost of much of the auth system as well as the wasm code of a contractcode entry, or the signers of an account entry.


Thank you for catching the missing deep clone charging, they can indeed be serious! 
These types of bugs are also easy to miss, e.g. every time the xdr updates with a new "deep" field, we have to update the metered_clone of its containing member.  Which makes me wondering, it is worthwhile to put in some systematic check on this invariant. One way I can think of is a fuzz/prop test that go through all the `MeteredClone` types and assert for each case "metered `MemAlloc` types >= actual allocated bytes", this will help detect the missed mem alloc (not the memcpy part which should be fine). I can create an issue for it if you agree, or have any better ideas. 

> One interesting part of this is that, as far as I can tell, everything in the events system doesn't use metered_clone anymore at all. I think this is good because it previously errored when trying to metered_clone a diagnostic event, which is "just as observable" as metered_cloning-something-that-should-be-free would have been. Luckily it looks like we only "accumulate events", don't clone them, and then externalize them when done, charging the cost of emitting them outside and ignoring the cost of externalizing.

Yes, this is my understanding as well. Except, when externalizing, the contract events are metered (not for diagnostic events obviously). For each `ContractEvents` it is charged with `Vec::<HostEvent>::charge_bulk_init_cpy(1, host)?;` , so that it is independent wrt size of the buffer.

graydon:
> it is worthwhile to put in some systematic check on this invariant. One way I can think of is a fuzz/prop test that go through all the `MeteredClone` types and assert for each case "metered `MemAlloc` types >= actual allocated bytes", this will help detect the missed mem alloc (not the memcpy part which should be fine). I can create an issue for it if you agree, or have any better ideas.

Yes please! Anything to help catch these in the future would be great.

> Yes, this is my understanding as well. Except, when externalizing, the contract events are metered (not for diagnostic events obviously). For each `ContractEvents` it is charged with `Vec::<HostEvent>::charge_bulk_init_cpy(1, host)?;` , so that it is independent wrt size of the buffer.

Aha. And the key reason this is ok is that it's not _cloning_ an XDR `ContractEvent` inside the `HostEvent`, rather it's doing `to_xdr` on an `InternalContractEvent` which charges the depth of the data field using `Host::from_host_val(Val)`. Ok!

Issue #1061, opened 2023-09-13
Change how 'free' budget works


dmkozh:
Currently `with_free_budget` just disables the budget metering completely. While this is conceptually correct (as the respective calls shouldn't be included into the final metering result), it makes the live of the users of the debug mode harder due to full lack of metering harness.

A simple fix would be as follows:

- Add a second budget object ('free budget'), with limits set to some customizable ratio from the 'main' budget (e.g. 1 or 1/2.)
- Switch to the second budget when calling `with_free_budget`. Thus the debug calls will remain unobservable for the 'real' metering, while still bounding the contract execution time.
- 
- (Probably the most annoying part) Make sure the callbacks used with `with_free_budget` are metered using more or less the same metering standards we use for all the host codebase (basically we need to stay within O(1) of the actual work performed).

The side effect of this is that functions might fail due to exceeding the 'free' budget while logging. This is an edge case, though, so it shouldn't matter much.

Additionally, we might want to make `with_free_budget` infallible and simply be a no-op when the 'free' budget is exceeded. This doesn't significantly improve the preflight experience for the edge case above, but would ensure that core nodes with debug enabled won't have divergent contract execution results (they might have divergent results currently due to errors during the 'debug-only' operations).

graydon:
(A small suggestion: if we're going to do this, which is fine by me, we should probably call the "free" budget the "debug" budget, for simplicity sake. It'll just confuse people otherwise)

graydon:
Another possible thing to backstop DoS scenarios that either slip through review or happen in diagnostic mode: we could wrap the whole thing in https://docs.rs/context-allocator/latest/context_allocator/allocators/struct.BumpAllocator.html or some similar fixed-size global allocator so panic out on too-much-memory rather than actually using too much malloc memory.

graydon:
Bumping this to a P0 and blocking final on it -- too risky a DoS vector on RPC servers to ship without this fixed.

PR #1123, opened 2023-10-20
Overhaul "free budget" and migrate non-metered code

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/1061

- Overhaul the "free budget":
    - Add two fields `internal_limit` and `internal_total_count` to budget dimension for keeping track of internal work being done, which don't get metered for fee, but need to be limited for DOS prevention.
     - Remove the `with_free_budget` interface and replace it with `with_internal_mode` (and a `with_debug_budget` on the host side that wraps around it and asserts host is in `DEBUG` mode), taking two closures and ensuring 1. the `internal_limit` is not exceeded 2. any error occurring within is suppressed. The error suppression part is to prevent error-based logic existing outside the closure that accidentally causes divergences between debug and no-debug workflows.
     - Adds a new hostile contract case for testing.

- A refactoring to split the `budget.rs` into multiple smaller files. The file was getting big (1200+ lines) and hard to navigate.
- Migrate all existing non-metered code to metered version, this includes diagnostic logging, diagnostic event, auth preflight code.

### Why

### Known limitations

[TODO or N/A]


jayz22:
All diagnostic workflows have been converted to metered code. Marking this back to ready for review.

soroban-env-host/src/budget.rs
```diff
@@ -779,7 +478,22 @@ impl Debug for BudgetImpl {
             )?;
         }
         writeln!(f, "{:=<165}", "")?;
+        writeln!(
+            f,
+            "Internal details (diagnostics info, does not affect fees) "
```
dmkozh, line 483: I'm not sure how I feel about the name 'internal'. I recall we've discussed 'debug' budget. That's not the best term either, but I think we could brainstorm this more, maybe something like 'diagnostics_budget'?
I also noticed that you call this 'debug_budget' in host, so it's even more confusing this way.

soroban-env-host/src/budget/dimension.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.limit.saturating_sub(self.total_count)
+    }
+
+    pub(crate) fn reset(&mut self, limit: u64) {
+        self.limit = limit;
+        self.total_count = 0;
+        for v in &mut self.counts {
+            *v = 0;
+        }
+        self.internal_limit = limit;
```
dmkozh, line 115: I think we might want to expose the internal limit multiplier or maybe just allow setting it to a flat value.

soroban-env-host/src/budget/dimension.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Tracks the sum of _output_ values from the cost model, for purposes
+    /// of comparing to limit.
+    pub(crate) total_count: u64,
+
+    /// The internal limit tracks work done internally that is not exposed to the
+    /// external user -- it does not affect fees or decide the invocation outcome
+    /// in any way (no error due to exceeding the internal limit). It exists solely
+    /// for dos prevention. Such work include diagnostic logging, or work that
+    /// exists only for preflight.
+    pub(crate) internal_limit: u64,
```
dmkozh, line 35: I thought we were going to introduce a separate budget object for this. I don't have a strong opinion on this (especially given that the current approach is already implemented) and I don't think we should change that, but I wonder if there is some fundamental issue that wouldn't allow us to use 2 regular budget structs and swap between them? If not, maybe add a todo/issue to revisit that later?

soroban-env-host/src/host.rs
```diff
@@ -467,6 +467,17 @@ impl Host {
         self.0.budget.charge(ty, input)
     }
 
+    pub(crate) fn with_debug_budget<T, F, E>(&self, f: F, e: E) -> T
```
dmkozh, line 470: I think a more appropriate name here would actually be something like `with_debug_mode` or something like that, as it also bypasses fn when we're not in debug mode (which is really convenient but should probably be reflected in the name).

soroban-env-host/src/events/internal.rs
```diff
@@ -156,30 +170,43 @@ impl InternalEventsBuffer {
     /// Converts the internal events into their external representation. This should only be called
     /// either when the host is finished (via `try_finish`), or when an error occurs.
     pub fn externalize(&self, host: &Host) -> Result<Events, HostError> {
-        let vec: Result<Vec<HostEvent>, HostError> = self
-            .vec
-            .iter()
-            .map(|e| match &e.0 {
+        let mut vec: Vec<HostEvent> = vec![];
+        vec.reserve(self.vec.len());
```
dmkozh, line 174: Shouldn't we used `metered_collect` here?

soroban-env-host/src/budget/util.rs
```diff
@@ -0,0 +1,160 @@
...
+#[cfg(any(test, feature = "recording_auth"))]
+impl Budget {
+    /// Fallible version of `with_internal_mode`, enabled only in testing and
+    /// non-production scenarios. The non-fallible `with_internal_mode` is the
+    /// preferred method and should be used if at all possible.
+    /// However, in testing and non-production workflows, sometimes we need the
+    /// convenience of temporarily "turning off" the budget. This can happen for
+    /// several reasons: we want the some test logic to not affect the production
+    /// budget, or we want to maintain an accurate prediction of production budget
+    /// during preflight. In the ladder case, we want to exclude preflight-only
```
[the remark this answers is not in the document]
jayz22: fixed.

soroban-env-host/src/test/budget_metering.rs
```diff
@@ -379,7 +379,11 @@ fn total_amount_charged_from_random_inputs() -> Result<(), HostError> {
         Int256Shift              1              None           412            119            412                 0                   119                 0                   
         ChaCha20DrawBytes        1              Some(1)        4926           0              4907                2461                0                   0                   
         =====================================================================================================================================================================
+        Internal details (diagnostics info, does not affect fees) 
         Total # times meter was called: 23
+        Internal cpu limit: 100000000; used: 0
```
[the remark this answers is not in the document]
jayz22: That's a good suggestion. I've replicated all the charging calls in internal mode and updated these numbers.

soroban-env-host/src/budget.rs
```diff
@@ -956,24 +670,50 @@ impl Budget {
...
-        let res = f();
+        let should_execute = self.mut_budget(|mut b| {
+            prev = b.is_in_internal_mode;
+            b.is_in_internal_mode = true;
+            b.cpu_insns.check_budget_limit(true)?;
+            b.mem_bytes.check_budget_limit(true)
+        });
+
+        let rt = if should_execute.is_ok() {
+            f().unwrap_or(e())
```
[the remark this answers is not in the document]
jayz22: Good catch! `unwrap_or_else` is lazily evaluated and more efficient. Fixed.

soroban-env-host/src/budget.rs
```diff
@@ -453,28 +134,46 @@ impl BudgetImpl {
         iterations: u64,
         input: Option<u64>,
     ) -> Result<(), HostError> {
-        if !self.enabled {
-            return Ok(());
+        if self.is_in_internal_mode {
+            if matches!(ty, ContractCostType::WasmInsnExec) {
+                // if wasm instruction charging is happening in debug mode, something
```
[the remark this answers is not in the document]
jayz22: On second thought, I've removed this check. This was from a previous iteration where the internal mode and debug mode were the same. Now that internal mode can happen in testing/preflight workflows (where `is_debug=false`), this check is no longer valid.
Thanks for the catch!

soroban-env-host/src/budget.rs
jayz22: Yeah it is definitely debatable. The choice of "internal" was a result of several considerations.
The internal budget is intended to cover two broad categories of work:
1. Debug, diagnostic work. These work happen in the production (non-test, non-preflight) workflow, but must have `is_debug` turned on. This is previously the "with_free_budget" use case for all diagnostic events and logging-related.
2. Testing, preflight. Most of these are intended for preserving production budget number, such as in the recording auth workflow.  But it may also include any other code that we just "don't care" about budgeting, such as for testing. This can only be activated for non-production scenarios, and the host can have `is_debug` either `true` or `false`.

Given these considerations, it appears like "internal" is the most accurate term from `budget`'s perspective. It is just an internal counter + limit that is not exposed to the user, and the budget doesn't care which scenario or workflow it came from. 

Note the host has a wrapping version called `with_debug_budget`, which asserts `is_debug()` is `true`. From the host perspective, this interface is the entry point for category 1 -- for all debug, diagnostic work where the `DEBUG` mode must be turned on.

soroban-env-host/src/host.rs
jayz22: Not so sure about that. This is a wrapping of "asserting we are in debug mode" and "use the internal budget for metering", so feels like `with_debug_budget` is the most accurate term. Or putting it another way, would there be any scenario where `DEBUG` mode and using internal budget diverges (we want former but don't want to do the latter), if no such scenario exists, then I think `with_debug_mode` also makes sense.

soroban-env-host/src/events/internal.rs
jayz22: No. This line is intentionally unmetered, because metering is taken care of inside the loop (right before `vec.push`) to ensure metering is indifferent of the buffer length.

soroban-env-host/src/budget/dimension.rs
jayz22: There isn't a fundamental issue, i.e. we could have two budgets and switch them as you suggested. It's more of a design choice to keep the interface clean and with minimal disruption. 
- The full `Budget` structure contains many other fields (for tracking, debugging), which isn't necessary for the "internal" use case (detailed [above](https://github.com/stellar/rs-soroban-env/pull/1123#discussion_r1371164341)).
- The requirement for this work comes from the fact that we need some way to mitigate DOS from the diagnostic workflow, by specifying some limit (separate from normal budget) which we don't expect normal behaviors to exceed. So exposing that limit externally isn't strictly necessary, as long as the limit is not too large for DOS, and not too small to be easily tripped over. 
- Setting the internal limit same as the regular limit seem to be a good choice. We could also expose a scale factor as an internal host constant, but exposing it to the budget constructor feels like an overkill (in which case we may also consider exposing new network config params for it). 

Overall this is the most lightweight approach that appears sufficient towards achieving the goal, while keeping all external interfaces unchanged.

soroban-env-host/src/budget/dimension.rs
jayz22: Yeah we could have a host constant for the limit multiplier or a flat value. I don't feel too strongly about that. 
@graydon suggested keeping things minimal and when needed it should be easy to switch to an internal constant.

soroban-env-host/src/host/error.rs
```diff
@@ -229,40 +214,48 @@ impl Host {
...
-                return Some(Box::new(DebugInfo { backtrace, events }));
-            }
-        }
-        None
+        self.with_debug_budget(
+            || {
+                // We do the event-borrowing ourselves here rather than calling
+                // self.try_borrow_events because we can/should only be called
+                // with an already-borrowed events buffer (to insulate against
+                // double-faulting).
```
graydon, line 248: This comment is copied from elsewhere and doesn't describe what's happening here; and at this point we also have a fallible helper `Host::get_events` that does the `externalize` call itself, I think we might as well use that and just do an `if let Ok(events) = self.get_events() { ... }` 

(I've also filed https://github.com/stellar/rs-soroban-env/issues/1132 to enrich this path a bit, DebugInfo is a bit too easily lost currently)

soroban-env-host/src/host/error.rs
```diff
@@ -202,21 +202,6 @@ impl<T> TryBorrowOrErr<T> for RefCell<T> {
 }
 
 impl Host {
-    /// Convenience function that only evaluates the auxiliary debug arguments
```
graydon, line 205: If you're going to remove this, remove the doc-comment that references it on the `err!` macro.

soroban-env-host/src/host/error.rs
```diff
@@ -229,40 +214,48 @@ impl Host {
...
+        self.with_debug_budget(
+            || {
+                // We _try_ to take a mutable borrow of the events buffer refcell
+                // while building up the event we're going to emit into the events
+                // log, failing gracefully (just emitting a no-debug-info
+                // `HostError` wrapping the supplied `Error`) if we cannot acquire
+                // the refcell. This is to handle the "double fault" case where we
+                // get an error _while performing_ any of the steps needed to record
+                // an error as an event, below.
+                // We do the event-borrowing ourselves here rather than calling
```
graydon, line 226: This comment got extended with a copy-paste of a comment from elsewhere that's restating what the comment above also says (that we're doing a fallible borrow) but incorrectly since we actually _don't_ want to assume we're in a shared-borrow here, we're assuming we're _not_ in a _mutable_ borrow, but are guarding against it happening by accident.

soroban-env-host/src/events/internal.rs
```diff
@@ -75,34 +75,41 @@ pub enum InternalDiagnosticArg {
...
-        };
-        Ok(xdr::ContractEvent {
-            ext: xdr::ExtensionPoint::V0,
-            contract_id: self.contract_id.clone(),
-            type_: xdr::ContractEventType::Diagnostic,
-            body: xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, data }),
-        })
+    fn to_xdr(&self, host: &Host) -> Result<Option<xdr::ContractEvent>, HostError> {
+        // this exist as an internal closure to prevent it from being accidentally
+        // called from the outside without `with_debug_budget`
```
graydon, line 81: why would it matter if this was called outside `with_debug_budget`?

soroban-env-host/src/events/internal.rs
```diff
@@ -156,30 +170,45 @@ impl InternalEventsBuffer {
     /// Converts the internal events into their external representation. This should only be called
     /// either when the host is finished (via `try_finish`), or when an error occurs.
     pub fn externalize(&self, host: &Host) -> Result<Events, HostError> {
-        let vec: Result<Vec<HostEvent>, HostError> = self
-            .vec
-            .iter()
-            .map(|e| match &e.0 {
+        let mut vec: Vec<HostEvent> = vec![];
+        // This line is intentionally unmetered, because metering is taken care during event push
+        // to ensure metering is indifferent of the buffer length.
```
graydon, line 175: "Indifferent" isn't quite right here. I think (if I understand correctly) the point is to avoid charging for the Diagnostic events at this point, only charging them to the debug budget.
 
```suggestion
        // This line is intentionally unmetered. We want to separate out charging the main budget
        // for `Contract` events (with "observable" costs) from charging the debug budget for `Diagnostic`
        // events (with "non-observable" costs). Both event types are stored in the same input vector
        // so we must not do a bulk charge based on its length, but rather walk through it charging
        // to one budget or the other on an event-by-event basis.
```

soroban-env-host/src/events/internal.rs
```diff
@@ -156,30 +170,45 @@ impl InternalEventsBuffer {
...
+                            Some(event) => Ok(Some(HostEvent {
+                                event,
+                                failed_call: *status == EventError::FromFailedCall,
+                            })),
+                            None => Ok(None),
+                        },
+                        || None,
+                    );
+                    if let Some(de) = de {
+                        vec.push(de)
```
graydon, line 206: This should also do `charge_bulk_init_cpy` like the observable Contract-event case. It's just charging to the debug budget.

soroban-env-host/src/events/internal.rs
```diff
@@ -130,14 +137,21 @@ pub(crate) struct InternalEventsBuffer {
...
+            || Vec::<(InternalEvent, EventError)>::charge_bulk_init_cpy(1, budget);
+
+        match e {
+            InternalEvent::Contract(_) => charge_internal_event_push()?,
+            InternalEvent::Diagnostic(_) => {
+                budget.with_internal_mode(charge_internal_event_push, || ())
+            }
         }
+
         self.vec.push((e, EventError::FromSuccessfulCall));
```
graydon, line 155: I'm not 100% sure this is right. I think we want to guard pushing the Diagnostic event on debug mode too, right? Like don't even record it if we're not in debug mode (usually this is filtered out by a caller I think but this function is marked `pub` so .. theoretically someone could call without). And also if the internal / debug charge fails, I think we should avoid pushing the corresponding diagnostic event.

soroban-env-host/src/host.rs
dmkozh: I think `with_debug_mode` reflects the use case better - I don't think we want to do any debug/diagnostic calls without debug budget (and vice versa). At least for now I can't think of such scenarios (but I could be missing something of course).

soroban-env-host/src/host/frame.rs
```diff
@@ -372,19 +372,32 @@ impl Host {
...
+        self.with_current_frame_opt(|opt_frame| match opt_frame {
+            Some(frame) => self.opt_contract_id_from_frame(frame),
+            None => Ok(None),
         })
     }
 
+    /// Returns [`Hash`] contract ID from the VM frame at the top of the context
+    /// stack, or a [`HostError`] if the context stack is empty or has a non-VM
+    /// frame at its top.
+    pub(crate) fn get_current_contract_id_opt_internal(&self) -> Result<Option<Hash>, HostError> {
```
graydon, line 397: This comment isn't right in its modified form -- now it only returns an error if the context stack is empty, and returns `None` if there's a non-VM frame on top of stack.

In general I think this function and the newly introduced one are sufficiently similar and their names are sufficiently uninformative that it's unlikely callers will know which they want (considering they have the same signature). Could we look at the call sites, figure out which cases are actually important to differentiate, and make functions that are clear about which cases they return and why?

soroban-env-host/src/budget/dimension.rs
graydon: yeah I don't see any need to overthink this: it's _specifically_ an unobservable part of the semantics, so if it has a simple interface now and someone asks us to make it a more-complex interface sometime in the future, it should not effect anything.

soroban-env-host/src/budget.rs
graydon: Yeah, I discussed this with @jayz22 on friday and we agreed that given the intended dual-use in both diagnostics and preflight, something more-general than `debug` is appropriate. Other terms I've considered here: "shadow budget", "ephemeral budget", "secret budget", "fallback budget", "unofficial budget". @dmkozh do you have any preference?

graydon: Yeah, I discussed this with @jayz22 on friday and we agreed that given the intended dual-use in both diagnostics and preflight, something more-general than `debug` is appropriate. Other terms I've considered here: "shadow budget", "ephemeral budget", "secret budget", "fallback budget", "unofficial budget". @dmkozh do you have any preference?

dmkozh: Ok, that makes sense. I'm still not 100% sure if there is a better name, but 'internal' seems good enough for now.

soroban-env-host/src/budget/dimension.rs
dmkozh: I think there is an immediate requirement for preflight (soroban-rpc) to be able to set this up.

soroban-env-host/src/budget.rs
dmkozh: 'Shadow budget' might be fine as well. While the budget itself is conceptually unobservable, its name is very much observable by the user (IIUC whoever prints out budget will see it), so we need something not too confusing.

soroban-env-host/src/budget/dimension.rs
dmkozh: Yeah, I agree this is serviceable for now. I still would feel a bit more comfortable if we had two distinct objects (so there is no concern that the internal logic is broken somewhere), but the usage seems limited enough to not be too concerned.

graydon (changes requested):
Looking good! Handful of changes requested but I think this is fairly close to what we should be merging, thanks!

soroban-env-host/src/budget/dimension.rs
jayz22: @dmkozh I think adding a couple of setter functions in the host should suffice? Or do we need platform team's input on this.

dmkozh: Just a public setter is sufficient for now.

soroban-env-host/src/budget.rs
jayz22: I renamed all the "internal budget" to "shadow budget". I agree the word "internal" might be a bit overloaded and could cause confusion.

soroban-env-host/src/budget/dimension.rs
jayz22: Added pub setters for the shadow budget limits.

soroban-env-host/src/host.rs
jayz22: Renamed to `with_debug_mode`.

soroban-env-host/src/host/error.rs
jayz22: Removed this comment. 
Sorry about the copy-paste, I originally got tripped on double fault here when I tried to do a cleanup and wrote a long comment, but then saw our comment elsewhere which is better written, so I decided to copy and modify. But realizing now it isn't accurate either. 
(Same as below)

soroban-env-host/src/host/error.rs
jayz22: Removed it in the comment.

soroban-env-host/src/host/error.rs
jayz22: Removed it (as well as the one above).

soroban-env-host/src/events/internal.rs
jayz22: Yes, I like your version better. Modified.

soroban-env-host/src/events/internal.rs
jayz22: You are right, this part is rewritten (with this fix and ones mentioned below).

soroban-env-host/src/events/internal.rs
jayz22: This is fixed.

soroban-env-host/src/events/internal.rs
```diff
@@ -75,34 +75,41 @@ pub enum InternalDiagnosticArg {
...
-        } else {
-            ScVal::Void
-        };
-        Ok(xdr::ContractEvent {
-            ext: xdr::ExtensionPoint::V0,
-            contract_id: self.contract_id.clone(),
-            type_: xdr::ContractEventType::Diagnostic,
-            body: xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, data }),
-        })
+    fn to_xdr(&self, host: &Host) -> Result<Option<xdr::ContractEvent>, HostError> {
```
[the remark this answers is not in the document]
jayz22: I was mainly making sure whichever function that is meant to be executed in debug mode is wrapped in `with_debug_budget`. But that makes the interface look a bit weird as you pointed out (here and above).  Fixed.

soroban-env-host/src/host/frame.rs
jayz22: Consolidated these two internal functions into one. 
I think we are using some of them in test workflow which needs to handle empty stack frame or `None` contract_id. 
@dmkozh may know better if we can clean it up further.

soroban-env-host/src/budget.rs
```diff
@@ -956,24 +665,50 @@ impl Budget {
...
-        self.mut_budget(|mut b| {
-            b.enabled = prev;
+        if let Err(_) = self.mut_budget(|mut b| {
+            b.is_in_internal_mode = prev;
             Ok(())
-        })?;
-        res
+        }) {
+            return e();
+        }
```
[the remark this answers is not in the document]
jayz22: Cleaned it up.

soroban-env-host/src/events/internal.rs
jayz22: I mainly wanted to make sure whatever logic that exist as a function that meant to be in debug mode is wrapped around `with_debug_budget`. 
I moved this closure out and added the `is_debug` check on entry.

soroban-env-host/src/host/frame.rs
dmkozh: I don't really know about all the use cases, but it seems like these can probably be consolidate into `Result<Hash, HostError>` (which should never error during normal execution, unless budget is exceeded). Tests can do `unwrap_or` or whatever is appropriate.

graydon: There are two situations here that are "possibly erroneous, possibly things we want to tolerate (not return errors for)":

  1. Empty context stack. Currently `with_current_frame_opt` tolerates this and returns `Ok(None)` for empty context.
  2. Non-contract frame (`Frame::HostFunction(...)`). Currently `get_current_contract_id_opt_internal` tolerates this and returns `Ok(None)` for a non-contract frame.

I think it's reasonable that some callers will want to treat these contexts as tolerable (eg. to emit events before a contract is running, such as the "we're about to call this contract" event or whatever).

I think it's probably _not_ reasonable for the cases that want to tolerate "empty context" to fail to tolerate "non-contract frame" (or vice-versa). I think if a caller tolerates one, it probably wants to tolerate the other.

As evidence for this, I note that the only caller of `get_current_contract_id_opt_internal`  that passes `true` (i.e. indicating that it _wants_ to differentiate those cases) is `get_current_contract_id_internal`, and it immediately treats the `None` case as an error as well, i.e. it collapses the two cases it just asked to differentiate back into a single case: either no-context or non-contract-context both result in an error return from it.

I therefore propose:

  - `get_current_contract_id_opt_internal` gets loses its boolean argument
  - its `None` branch just becomes `None => Ok(None)`, i.e. it collapses both cases into `Ok(None)`

graydon: (applied this fix myself)

graydon (approved):

PR #1126, opened 2023-10-20
Gate recording auth behind the feature.

dmkozh:
### What

Gate recording auth behind the feature. It will also be automatically available for `testutils` builds (necessary for the SDK).

### Why

Recording auth is needed in non-test builds (for preflight), but not for validators. Thus we can compile it out for additional safety.

### Known limitations

N/A


leighmcculloch (approved):
Great idea 👏

PR #1128, opened 2023-10-23
Add conceptual overview comment to auth module

graydon:
To help prepare myself for reviewing the auth module, I decided it'd be best to do a careful read of the CAP and XDR and write down my understanding from a very high level without reading much of the code beyond data structures. This is the result. I'd love to know if (a) it's approximately correct and (b) this is a good place to leave the comment, or if I ought to move it to docs somewhere else (eg. try to incorporate it into https://soroban.stellar.org/docs/fundamentals-and-concepts/authorization).


soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//!
+//!   - What type of "requirements" can the contract require?
+//!   - What are the contents of the authentication context?
+//!
+//! To answer either of these questions we first need to take a slight detour
+//! into the notion of an _address_.
+//!
+//! ## Addresses
+//!
+//! Addresses generalize and abstract over accounts (corresponding to public
```
dmkozh, line 42: This is not technically incorrect, but I think is a bit misleading. The abstractions here are:

- Accounts (as in 'account abstraction') - entities owned by the users external to the chain. Includes, but not limited to, the Stellar accounts and 'custom' accounts backed by contracts.
- Contracts - entities that only act on-chain (e.g. a liquidity pool contract that transfers money to the user).

The key thing here is that the only special thing about the 'classic' Stellar accounts is that they get a special variant in the `Address` enum, which is more of an implementation detail than anything else.

I think this is correctly reflected in the `Authentication` section below, but here I feel like there is some room for incorrect interpretation.

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//!      there are daily transfer limits on an account to minimize risks. Here
+//!      the rules have nothing to do with authority in the sense of a person,
+//!      merely the evaluation of a given situation.
+//!
+//! Some authorization scenarios will combine cases, and sometimes case 2 will
+//! delegate back to case 1 ("the rules" are "check for a signature from X").
+//! The examples above are illustrative of cases with the strictest difference,
+//! but that difference has crucial practical details:
+//!
+//!   - In case 1, the presence of a _valid signature_ (authentic credentials)
```
dmkozh, line 69: The term 'signature' (here and some other places) creates a false impression that multisig is not supported, which is not true.

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//! the auth module "did A authorize me?" in an A-called-B context, the auth
+//! module will just say "yes". We will see later there are other ways A might
+//! authorize B at some further distance than a direct call, but these are
+//! generalizations of the same principle: a contract _is_ at least partly an
+//! authorization policy, and if it says "yes" in a context where it has been
+//! designated an authority for the sake of authorization, then that "yes" is
+//! all the evidence we need.
+//!
+//! To sum up: addresses exist to generalize between these two cases: they allow
+//! a contract to define its authorization requirements in terms of _either_ a
```
dmkozh, line 96: > //! a contract to define its authorization requirements in terms of _either_ a
> //! person (account), or a program (contract), or both.

The way I read this is that a contract can *choose* to define authorization requirements *only* for e.g. an account, which is not true. The requirement always abstracts the way the authorization is actually performed. Maybe I'm just not reading this correctly, but that still means that we probably need to reword this statement.

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//!      because they are assumed to have been checked for authenticity by the
+//!      embedder, such that the host would never have been instantiated if the
+//!      transaction had an invalid signature. Indeed there is no signature to
+//!      check in this case anyways: the enclosing transaction signature is not
+//!      given to the host.
+//!
+//!   3. Account-based authorizations with credentials that delegate to an
+//!      address that is an `AccountID` (`SOROBAN_CREDENTIALS_ADDRESS` with
+//!      `SC_ADDRESS_TYPE_ACCOUNT`) will load the Stellar classic `AccountEntry`
+//!      from the ledger and evaluate the credential's signature as a normal
```
dmkozh, line 180: > credential's signature

This should be 'signatureS', right?

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//!      Ed25519 signature with the signers-and-weighting scheme built in to
+//!      Stellar classic.
+//!
+//!   3. Account-based authorizations with credentials that delegate to an
```
dmkozh, line 184: 3->4

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//!      address that is an contract (`SOROBAN_CREDENTIALS_ADDRESS` with
```
dmkozh, line 185: a contract

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//!      `SC_ADDRESS_TYPE_CONTRACT`) will instantiate the contract and _call a
+//!      function on it_ to allow the contract to perform authentication, for
+//!      example by implementing signature verification in a public key system
+//!      other than Ed25519. The function implementing such custom
+//!      authentication must be called `__check_auth` and it will be passed an
+//!      abstract, uninterpreted byte string as the "signature" provided in the
```
dmkozh, line 191: > uninterpreted byte string as the "signature"

That's not correct - signature is actually an `ScVal` that must resolve to an arbitrary valid `Val`.

dmkozh (commented):
This looks good/correct overall (besides a few minor things) - I think it takes a bit different angle compared to docs I wrote before, but generally conveys the same ideas.

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,197 @@
...
+//! be _either_ of authorization, authentication, or both.
+//!
+//! ## Authorization
+//!
+//! Authorization judgments are explicitly requested by smart contracts, and
+//! relate to an ambient authorization context that is maintained by this
+//! module. Specifically a contract calls the host (using `require_auth` or
+//! `require_auth_for_args`) informing the host that the contract code needs
+//! authorization before it continues. If the requirement matches the context
+//! the call succeeds, otherwise it fails (and the contract typically traps).
```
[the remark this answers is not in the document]
graydon: fixed

soroban-env-host/src/auth.rs
graydon: Hm. I think I see what you're saying: that the two different variants of the `Address` type -- `Account` and `Contract` -- do not correspond to _conceptual_ differences between "abstract accounts" and "abstract contracts", because one kind of "abstract account" is a custom account backed by a contract (and so identified by `Address::Contract`). Is that what you mean? I will try to clarify this.

soroban-env-host/src/auth.rs
graydon: Ok. I'll clarify.

soroban-env-host/src/auth.rs
graydon: fixed.

soroban-env-host/src/auth.rs
graydon: fixed

graydon:
@dmkozh I've condensed and mostly-rewritten the comment to reflect further understanding from re-reading docs on the soroban website and discussion with you. Please let me know if it's .. closer to accurate?

soroban-env-host/src/auth.rs
dmkozh: Yeah, exactly. The variants basically only influence the authentication code routing and shouldn't otherwise influence any logic.

soroban-env-host/src/auth.rs
```diff
@@ -1,3 +1,151 @@
...
+//! And again, in both directions the "auth services" required or provided may
+//! be _either_ of authorization, authentication, or both.
+//!
+//! All auth services reason about invocations and authorizations, so we
+//! next turn our attention to these.
+//!
+//! ## Invocations
+//!
+//! A Soroban transaction can be seen as a tree of _invocations_: usually
+//! invocations of contract functions, though strictly speaking transctions
```
dmkozh, line 29: We actually may have host functions in the middle of the invocation tree now (specifically, 'create contract' host fn). So we may have call stacks like `contract->create contract host fn->__check_auth->other contract` etc. So the generalized behavior is that call tree has a mix of host function and contract calls, but most of the host function calls are transparent for auth and are omitted for simplicity.

dmkozh (approved):
Looks good overall, thanks!

soroban-env-host/src/auth.rs
graydon: TIL! I'll update that to clarify, thanks.

PR #1133, opened 2023-10-25
Stop caching authorization tracker verification status.

dmkozh:
### What
Stop caching authorization tracker verification status.

Also did a bit of passing-by cleanup (removed another unnecessary state variable `need_nonce`, convert all the authentication errors to auth errors for clarity).

### Why

The previous behavior seemed appropriate but is ultimately not correct for some fringe cases. Also added a test covering the scenario where this matters, as well as some additional rollback tests.

### Known limitations

N/A


soroban-env-host/src/auth.rs
```diff
@@ -1760,14 +1750,13 @@ impl AccountAuthorizationTracker {
...
-        self.authenticated = snapshot.authenticated;
-        self.need_nonce = snapshot.need_nonce;
+        self.verified = snapshot.verified;
         Ok(())
     }
 
     // metering: free
     fn is_active(&self) -> bool {
-        self.is_valid && self.invocation_tracker.is_active()
+        self.verified && self.invocation_tracker.is_active()
```
graydon, line 1759: Is this correct? I mean .. if it hasn't been verified _yet_, isn't it still potentially active?

dmkozh: These conditions are mutually exclusive (because we only match invocations to valid trackers), so I guess `verified` check is kind of redundant.

graydon (approved):

PR #1138, opened 2023-10-28
tighten up handling of debug mode

graydon:
This does a few things @jayz22 and I discussed today:

1. Eliminates the second callback and the value-passing on the shadow and debug mode callbacks in favor of mutating a value outside the closure _when observation is required_. This is not necessarily better, just a little more awkward / less likely to be done accidentally without thinking about it.
2. Removes all convenience accessors for the debug flag, again just to make it harder to accidentally observe.
3. Feature-gates DebugInfo on testutils, so it's never live in production. Note that DebugInfo is _not_ the diagnostic event buffer, it's a snapshot of the buffer along with a (native) backtrace attached to errors that get printed to console in a panic catch, it only makes sense for testing and by eliminating it here we can actually compile _out_ all support for backtraces (including the dwarf-reading crate) from production. Which is just one less pile of software to worry about shipping (this is also good because backtrace is going to be stable in the rust stdlib soon).

### DIFF
--- a/soroban-env-host/Cargo.toml
+++ b/soroban-env-host/Cargo.toml
@@ -26,7 +26,7 @@ rand_chacha = "0.3.1"
 num-traits = "0.2.15"
 num-integer = "0.1.45"
 num-derive = "0.4.0"
-backtrace = "0.3"
+backtrace = { version = "0.3", optional = true }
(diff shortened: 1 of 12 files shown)


jayz22 (approved):
This is great! Really nice to see the diagnostic workflows getting safer.

dmkozh (approved):

PR #1139, opened 2023-10-28
Adapt to rs-stellar-xdr `Limits` change

jayz22:
### What

 - Adapts to https://github.com/stellar/rs-stellar-xdr/pull/317 and runs perf analysis.
 - Make `DepthLimiter` internal (the depth limiting mechanism for the host stays the same)
 - Defines the default XDR read/write limits (depth, length). 

### Why

### Known limitations

[TODO or N/A]


soroban-env-host/Cargo.toml
```diff
@@ -13,8 +13,8 @@ rust-version = "1.71"
 [dependencies]
 soroban-builtin-sdk-macros = { workspace = true }
 soroban-env-common = { workspace = true, features = ["std", "wasmi"] }
-stellar-strkey = { workspace = true }
 wasmi = { workspace = true }
+stellar-strkey = "0.0.8"
```
jayz22, line 17: I had to pin the strkey dependency to crates registry instead of commit hash. Because it was failing cargo-deny showing up as two different version (https://github.com/stellar/rs-soroban-env/actions/runs/6793313451/job/18467959394). `rs-stellar-xdr` is depending on strkey using the crates registry.  
I'm not sure why this previously wasn't an issue though. 
@leighmcculloch let me know if this is not the right move.

soroban-env-host/src/budget/limits.rs
```diff
@@ -0,0 +1,99 @@
...
+    Error, HostError,
+};
+
+/// These constants are used to set limits on recursion and data length in the
+/// context of XDR (de)serialization. They serve as safeguards agaist both
+/// exccessive stack allocation, which could cause an unrecoverable `SIGABRT`,
+/// and exccessive heap memory allocation.
+pub const DEFAULT_XDR_RW_LIMITS: Limits = Limits {
+    /// recursion limit for reading and writing XDR structures.
+    depth: 500,
```
dmkozh, line 14: Why is this limit 500? It is not consistent with the host limit. I'm not sure if it's actually possible to reach this limit without hitting `DEFAULT_HOST_DEPTH_LIMIT` first, but in any case this is too high. 
I don't think we need a separate XDR depth limit, everything should be consistently guarded by the same host limit.

soroban-env-host/Cargo.toml
leighmcculloch: We shouldn't ever need to use a hash git ref for the stellar-strkey package. It's low level enough that when we make changes to it we ship them immediately as a release. Their releases are not tied to overall system releases.

soroban-env-host/src/budget/limits.rs
jayz22: This limit is purely for xdr read and write, enforced on rs-stellar-xdr (previously it was defined as a default value in there). 
As for why the values are different from the host. The xdr (de)serialization takes up fewer bytes on each recursion level, compared to host recursions (val conversion, cross-contract calls), that's why the host depth limit is more conservative than the xdr.

dmkozh: This doesn't sound right to me. I'm not sure why would it take fewer bytes in the first place; actually the host calls have constant argument size, while I'm not so sure about the XDR calls. But that's really beyond the point; the main thing is that the limit should be consistent across the host. Otherwise you might end up e.g. with an `ScVal` that can't be converted to `Val` due to the host limit. There is really no good reason for these to be inconsistent.

jayz22: For the Val<->ScVal conversions and comparisons, perhaps yes. The val conversion paths in the host are a bit more involved than xdr conversion.
For the cross contract calls, I don't think it should really share the same depth limit. Making a function call stores more context (e.g. local variables) and can run out of stack more easily (you've seen it first hand).

jayz22: I think it is important to point out. These limits are a safety net, catch-call kind of limit that are not meant to be triggered by normal operations and by good actors. It is intended to be a proxy for the operating system stack limit, which is different for platforms and can be triggered differently. The goal is not to get a consistent experience of how everything fails, the goal is to prevent from reaching such failing points at all.

dmkozh: > For the cross contract calls, I don't think it should really share the same depth limit.

I agree here, which is why I also suggested to have an explicit contract call depth limit. This has nothing to do with XDR write limit though.

> The val conversion paths in the host are a bit more involved than xdr conversion.

It doesn't really matter how involved they are. I'm arguing for consistent behavior, not for per-use case optimization. There is no reason for why I should be able to deserialize a nested `ScVal`, but then fail while trying to convert it to `Val`.

> The goal is not to get a consistent experience of how everything fails, the goal is to prevent from reaching such failing points at all.

Sure, I think we are on the same page here. Which is why I'm suggesting to use more consistent limits on the host side.  We should be able to clearly explain (at least to ourselves) what the depth limits are for every operation, as they are a part of the protocol. I can't really see where the XDR limits would be reflected, besides maybe consuming some input data during the host setup (but never actually processing it fully).

jayz22: Can we discuss this in a follow up so that this PR is not blocked? 
The XDR and host value implementations (as well as depth-limiting implementation) are different. The xdr ScVal has a number of intermediate layers (e.g. `VecM`) for each specific discriminant type (e.g. `Bytes`), which means the actual max recursion will be different even if we set their limits to be same. Again, I believe this is a limit that we shouldn't have to worry about, it is just in there in case something bad happens.

dmkozh: Ok, we can discuss this later as well, but I don't agree that we shouldn't worry about this - there is a difference between 100 and 500 recursive calls and 500 might actually end up being too much.

dmkozh (approved):

graydon (approved):

PR #1142, opened 2023-10-31
Initial cut at bug 872, observe side effects of tests

graydon:
This is still a draft but I'd say it's about 75% done -- the main tricky (or _hacky_) parts are all working about as well as I can make them.

The goal here is to provide something _like_ what we do for the txmeta recording in core: leverage the unit test suite to record traces of observable side effects into a stable file checked in to version control such that we _notice_ when we unintentionally change observable behaviour of anything the tests exercise.

I just did some preliminary observations to get it working (budget values and prng state). Will add more.

This doesn't work _quite_ as seamlessly as the core version: the Rust unit test harness is basically impossible to hook a pre-run / post-run interceptor into directly, so I wound up providing a wrapper around the Host that records its state when destroyed, and then augmenting it with a (macro-captured) name-of-the-function it's created in. Since most of our tests have exactly one host, this works (and it is only a little work to generalize to tests with multiple hosts).

Unfortunately there's no shutdown system in Rust at all, nor for the overall test harness, and the community consensus seems to be that hooking into libc's `atexit` is the best approach. So I did that too. Gross.

There are a few things left to do here which I'll have to pick up tomorrow, but it seems to _mostly_ work:

- [ ] support both curr and next files
- [ ] go through the rest of the testsuite modifying the Host construction sites
- [x] observe a lot more state about the Host
- [ ] maybe warn when there are unused entries in the observations file? maybe not since running single tests is common
- [x] connect to the host lifecycle hooks system to do some _cumulative_ observations (eg. as VMs are set up and torn down)


dmkozh (commented):
I think it's probably good enough to start with (besides the built-in contract issues). But I wonder if it's possible to observe the host function measurements vs the whole test measurements. That would make it much easier to distinguish between changing the test setup  (e.g. adding an assertion with getter) vs changing the actual host implementation. Again, that's not critical for now, I just wonder how hard/feasible that would be.

graydon:
> I think it's probably good enough to start with (besides the built-in contract issues). But I wonder if it's possible to observe the host function measurements vs the whole test measurements. That would make it much easier to distinguish between changing the test setup (e.g. adding an assertion with getter) vs changing the actual host implementation. Again, that's not critical for now, I just wonder how hard/feasible that would be.

@dmkozh do you mean a separate observation every time any host function call happens during each test? or like a summary of host function counts? I think I could make either work.

dmkozh:
> @dmkozh do you mean a separate observation every time any host function call happens during each test? or like a summary of host function counts? I think I could make either work.

The way I'd envision this is that we hook observations to public (env-common) host fns (hopefully there is a way to generate this). Then the observations can be grouped by test, similarly to what we do here. Then we could conceptually get something like 
```
test_foo(): {
  "vec_new": "abc...",
  "vec_push": "def...",
  "vec_new" : "hij...",
}
```
where we could just store the hash of budget+return value+prng etc. (or even granular numbers, depending on how verbose that would be).
Then if I modify the test it would be very easy to correlate diffs to the modifications I've made (e.g. if I've just added a host fn call, then I expect that nothing else has changed besides my new call).

graydon:
@dmkozh I've updated this with much more involved and finer-granularity measurements. let me know if it's to your liking.

soroban-env-host/observations.json
```diff
@@ -0,0 +1,1091 @@
...
+    "  39 call has_contract_data(51035662, 1)": "cpu:14017581, mem:6816636, prngs:-/d8a09d57, objs:0@150c532c/176@9ba855af, vmem:-, evt:3@b3315b9, store:0@58c79e45/9@4e4e207f",
+    "  40 ret has_contract_data -> Ok(\"False\")": "cpu:14030492, mem:6817468, prngs:-/d8a09d57, objs:0@150c532c/176@9ba855af, vmem:-, evt:3@b3315b9, store:0@58c79e45/10@7d7a448f",
+    "  41 call has_contract_data(51035662, 0)": "cpu:14030492, mem:6817468, prngs:-/d8a09d57, objs:0@150c532c/176@9ba855af, vmem:-, evt:3@b3315b9, store:0@58c79e45/10@7d7a448f",
+    "  42 ret has_contract_data -> Ok(\"False\")": "cpu:14044165, mem:6818368, prngs:-/d8a09d57, objs:0@150c532c/176@9ba855af, vmem:-, evt:3@b3315b9, store:0@58c79e45/11@1bea8adc",
+    "  43 leave VM:6d982c6:sym#347 -> Ok(False)": "cpu:14044165, mem:6818368, prngs:-/d8a09d57, objs:0@150c532c/176@9ba855af, vmem:1114112@d3636605, evt:3@b3315b9, store:0@58c79e45/11@1bea8adc",
+    "  44 end": "cpu:14065967, mem:6819900, prngs:-/d8a09d57, objs:-/178@dcadce48, vmem:-, evt:3@b3315b9, store:-/11@1bea8adc"
+  },
+  "test::vec::empty_vec_back": {
```
dmkozh, line 1019: Why do we not have host function call traces here?

soroban-env-host/observations.json
```diff
@@ -0,0 +1,1091 @@
+{
```
dmkozh, line 1: This looks great, thanks! The only UX-side thing I would probably consider doing (probably not in this PR) is to output step diffs where possible (cpu and mem are trivial, storage and objects are possible in theory thanks to immutability). The reason is that diffs are easier to read and also are better for the cases when test is modified in the middle (we could of course try to avoid these, but I suspect we'll still need to do that from time to time).

dmkozh (approved):

soroban-env-host/observations.json
graydon: I'm not sure what you mean by step diffs but I should highlight two facts here:

1. I'm emitting a single JSON map and rendering each K:V entry in it manually to make it compact for reading; this means the structure of each line is lost on rendering and can't be recovered. it's just "a string" after rendering. so I can't diff the fields within that string.
2. I'm not recording anything else (eg. the actual content of the storage or whatever) from previous runs, _just_ this one file. so we can't diff the preimage (storage state, PRNG state, whatever) that went into each field either.

soroban-env-host/observations.json
graydon: They're direct (rust) calls on the host, not mediated through the VM->host dispatch machinery where I inserted hooks. I might be able to insert more hooks in the VmCallerEnv emulation layer, that's a good point. I'll try that too.

soroban-env-host/observations.json
dmkozh: By 'diffs' I mean the difference between the current step and the previous step. At least for CPU/mem that seems straightforward (so that instead of `step 0: cpu: 1000`, `step 1: cpu: 2500` you'd get `step 0: cpu_delta: 1000`, `step 1: cpu_delta: 1500`. E.g. for the host objects it should be enough to hash not the whole array, but only the slice of (`prev_step_object_count..curr_step_object_count`). For storage it should be enough to store the storage reference for every step and then diff it (we're maintaining all the storage states anyway, right?).

I just want to make it clear again, that this suggestion definitely shouldn't block this PR. I realize it might require a bit of refactoring and it's not that high priority. Just something I'd maybe take a look at, maybe some time after release.

soroban-env-host/observations.json
dmkozh: Yeah, it would be really nice to intercept any host call, because we rarely dispatch test host calls through VM.

soroban-env-host/observations.json
graydon: ah, I see. I was able to accomplish a minimal version of this -- omitting fields that don't change from one line to the next -- without much refactoring. I also hooked the vmcallerenv-based (direct) calls, which reveals quite a lot more traffic (also helps explain why so many host objects -- the env macros aren't using bulk object construction I guess).

let me know if the new file reads easily enough. I think I might also change to a separate observations file per test, the unified "observations.json" file is already 1.9MiB and it's going to get a lot bigger. Github has a 100MiB file size limit and diff certainly isn't going to like diffing files that large either.

soroban-env-host/observations.json
graydon: done.

Issue #1136, opened 2023-10-27
Enhance `soroban-synth-wasm` for fuzzing support


jayz22:
Currently `soroban-synth-wasm` does simple things such as individual wasm instructions on i64. It is not usable for real testing because it doesn't understand any of the host concepts, e.g. value layout, host functions etc. We would like to enhance it in such ways that makes it useful for host testing/fuzzing:
- [x] Allow passing the argument with correct types (which could be randomly generated, or some from output of a host function call) to another host function.
- [x] Generate a footprint and pass footprint keys as arguments to fuzzed contract so fuzzer can find them
- [x] Generate addresses and authorized invocations (possibly by recording in preflight) also passing keys as arguments so fuzzer can find them
- [x] Able to generate "bad" examples, e.g. contracts with complex structures or excessive calls to the host
- ~[ ] Add missing sections. Able to generate "bad" sections (e.g. large table or linear memory).~
- ~[ ] Contract to access regions of linear memory certain data types are stored~
- [x] At the fuzzer level, being able to randomly pick host functions, and call it with correct arity and arguments
- [x] Make cross contract calls
- [x] Cover any other semantics of contract

This is a summary of discussion with @graydon, please let me know if I've described the requirements accurately.

graydon:
Skipping two of these -- they're possible but my attempts to "intelligently" guide the fuzzer with them just soak up randomness from it and lower coverage. I think the fuzzer is doing an ok job finding its way around linear memory just using random numbers.

PR #1171, opened 2023-11-01
`Arbitrary`-compatible expr generation and host fuzzing

jayz22:
### What

#1136 

### Why

### Known limitations

[TODO or N/A]


graydon (approved):
Hooray!

PR #1173, opened 2023-11-03
Fix fuzzer-found frame stack corruption in rollback if instance storage fails

graydon:
This used to be a panic but as of #1160 it's just an `InternalError`. Still, not something we want!


dmkozh (approved):
Oh, sorry about that. It seems like other fallible operations can only fail due to borrow errors, which hopefully won't happen (unless there is a bug elsewhere).

PR #1180, opened 2023-11-07
Fix #1174 fuzzer-found deserialization of non-representable ScVal bug.

### LINKED ISSUE #1174: Fuzz bug: InternalError: deserialize a non-representable scval
```
   1: [Diagnostic Event] contract:0000000000000000000000000000000000000000000000000000000000000000, topics:[error, Error(Value, InternalError)], data:["unexpected non-Val-representable ScVal type", 20]
```
inside a call to `deserialize_from_bytes`, from this program:
```
	Expr(
	    deserialize_from_bytes(
	        bytes_push(
	            bytes_pop(
	                serialize_to_bytes(
	                    vec_back(
	                        vec_push_back(
	                            vec_new_or_existing(
	                                0,
	                            ),
	                            sub_Void(
	                                extend_contract_instance_and_code(
	                                    get_current_contract_address,
	                                    get_ledger_version,
	                                    get_ledger_version,
	                                ),
	                            ),
	                        ),
	                    ),
	                ),
	            ),
	            get_ledger_version,
	        ),
	    ),
	)
```

Probably just a little too quick to treat non-representability as InternalError rather than potentially-user-controlled input.

graydon:
Fixes #1174


soroban-env-host/src/host.rs
```diff
@@ -2077,7 +2077,16 @@ impl VmCallerEnv for Host {
...
+        if Val::can_represent_scval(&scv) {
+            self.to_host_val(&scv)
+        } else {
+            Err(self.err(
+                ScErrorType::Value,
+                ScErrorCode::InvalidAction,
+                "Deserialized ScVal type cannot be represented as Val",
+                &[(scv.discriminant() as i32).into()],
+            ))
+        }
```
jayz22, line 2089: Should we instead change the error type inside the `check_val_representable_scval`, instead of copying its code? 
Of the 4 places it's being called, 3 seems to correspond to the same error type. The only one that would be an internal error is `from_host_val`, and we can `map_err` for it.

graydon: I don't think so. I think the other sites are not accessible to user controlled input so it's properly an internal error for a non-val-representatable ScVal to show up. The embedder should have prevented it.

dmkozh: The check as it is implemented here is non-exhaustive; instead it should happen recursively for every produced `ScVal`. A test where we try to deserialize e.g. a vector with `LedgerKeyContractInstance` element would help.

So probably we should do what Jay suggests: update `check_val_representable_scval` to return `Value, InvalidAction` error and convert it to `InternalError` where appropriate.

graydon: Oh hah! I think the lack of recursion in is_val_representable is itself a bug. I'll fix it along with this.

I don't want to start mapping errors _out_ of `InternalError` since it risks hiding or mis-classifying other internal errors. Since 3 sites it's called are self-consistency checks -- not points of exposure to user input -- I really do think they should remain `InternalError`.

(For consistency of reporting I'm actually going to change this to `UnexpectedType`, to match the the low-resolution `ConversionError` projected errors we get from `TryFromVal` on a non-representable case going through conversion)

dmkozh (approved):

PR #1182, opened 2023-11-08
Add fn to find out if Host can be finished

leighmcculloch:
### What
Add fn to find out if Host can be finished.

### Why
So that the SDK can ask the Host if it has no more references left and it is able to be finished without calling try_finish which is destructive.

I plan to use this in:
- https://github.com/stellar/rs-soroban-sdk/pull/1137

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -486,9 +486,20 @@ impl Host {
         self.as_budget().with_shadow_mode(f, e)
     }
 
+    /// Returns whether the Host can be finished by calling
+    /// [`Host::try_finish`].
+    ///
+    /// Returns true if the host reference is unique, refcount = 1.
+    pub fn can_finish(&self) -> bool {
+        Rc::strong_count(&self.0) == 1
+    }
+
     /// Accept a _unique_ (refcount = 1) host reference and destroy the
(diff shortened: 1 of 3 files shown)


dmkozh:
I'm not necessarily opposed to this change, but I wonder if it's needed for your purpose. You can get read-only access to storage and events at any point - there is really no need to destruct host to do that (and IIUC the linked PR doesn't actually call `try_finish`). `try_finish` is there mostly as a minor optimization for e2e flows and it really just ensures that there is no bug that causes something to hold a reference.

leighmcculloch:
> I'm not necessarily opposed to this change, but I wonder if it's needed for your purpose.

The SDK needs to know if the `Host` has a strong ref count greater than `1` because it is the only way the SDK can tell if the Env being dropped is the last holder of the Host. Env's get dropped all the time, but they all hold a shared ref to the Host via the internal Rc

I could add a `ref_count` fn that simply returns that. But given the Host has a concept of being "finishable" that is directly tied to whether the host is a unique ref or not (a ref count of 1), it feels like it works to add that function based on that terminology.

dmkozh:
I see, thanks for the explanation.

dmkozh (approved):

Issue #1153, opened 2023-11-01
[Test] Build the largest host stack possible within the budget limits


dmkozh:
We need to make sure host doesn't overflow the stack. The idea for the setup is as follows: call use a custom account contract that delegates `__check_auth` call to another instance of the same contract via `require_auth`. This will result in a few host stack frames per contract invocation (definitely more than direct `call`). The recursion depth should be configurable. Then find the maximum stack depth within the budget, and call XDR serialization on a deep recursive XDR (which is supposedly cheaper than cross-contract calls).

We also need to make sure to update this test if per-transactions network limits are increased (specifically the maximum number of entries read and maximum number of instructions allowed).

PR #1185, opened 2023-11-08
Add tests that try to build a deep host stack.

dmkozh:
### What

Add tests that try to build a deep host stack.

The tests combine sub-contract calls with XDR serialization (these happen to have their depth limited independently). Sub-contract calls also go through auth subsystem in order to increase the number of stack frames per contract invocation.

Resolves https://github.com/stellar/rs-soroban-env/issues/1153

### Why

Improving test coverage, making sure there is no host stack overflow.

### Known limitations

N/A


Cargo.toml
```diff
@@ -49,5 +49,9 @@ rev = "e6ba45c60c16de28c7522586b80ed0150157df73"
 # soroban-wasmi_core = { path = "../wasmi/crates/core/" }
 
 [profile.release]
-codegen-units = 1
 lto = true
+
+[profile.test-opt]
+inherits = "dev"
```
leighmcculloch, line 55: I think the `test-opt` profile should inherit from `test`. Functionally there will be no difference. Semantically it makes more sense.

soroban-env-host/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    let opt_level = std::env::var("OPT_LEVEL").unwrap_or_else(|_| "0".to_string());
+    if opt_level != "0" {
+        println!("cargo:rustc-cfg=feature=\"opt_build\"");
```
leighmcculloch, line 4: We don't need this to be a feature. Adding it as a feature isn't really using features correctly by cargo's definition, because features are supposed to be unifiable. Not that we always use features in that way, but we actually don't need a feature in this case.

You can just just make it a custom cfg.

See https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg

For example:
```suggestion
        println!("cargo:rustc-cfg=opt_build");
```

Then in the code you can write:

```rust
#[cfg(opt_build)]
```

leighmcculloch (commented):
I reviewed just the toml/build-script part of the change for the moment and suggest a couple changes.

Cargo.toml
dmkozh: Makes sense, updated.

soroban-env-host/build.rs
dmkozh: Cool, I kind of forgot about this possibility. The only downside is that it's not possible to actually force including this feature for debugging, but I guess temporarily removing the condition would work as well.

graydon (approved):

PR #1191, opened 2023-11-09
Adds loadgen test WASM

SirTyson:
### What

Adds a WASM for loadgen testing. This provides two load functions. First, `write` allows us to write an arbitrary number of `ContractData` entries of arbitrary size. `do_work` loops an arbitrary number of times over arithmetic inside the guest and a simple host function. This provides CPU load on both the guest and host. `U256::mul` was chosen because it is one of the least expensive host functions. The more computationally expensive host functions overestimate instruction count, so picking a cheap host function allows us to do the most amount of actual CPU work before hitting the instruction count limit.

I don't know if I need to include all the updated compiled WASM for other tests. I just ran make inside the test folder and included all the updated binaries.


soroban-test-wasms/src/lib.rs
```diff
@@ -57,6 +57,8 @@ mod curr {
         include_bytes!("../wasm-workspace/opt/curr/example_contract_data.wasm").as_slice();
     pub const LINEAR_MEMORY: &[u8] =
         include_bytes!("../wasm-workspace/opt/curr/example_linear_memory.wasm").as_slice();
+    pub const LOADGEN: &[u8] =
```
dmkozh, line 60: nit: not need to name this 'example' as it's not really an example.

soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+
+const EXTEND_LEDGER: u32 = 5_000_000;
+
+#[contractimpl]
+impl Contract {
+
+    // This will write num_entries of size size_kilo_bytes. The initial entry will have a key of
+    // starting_index, the next key starting_index + 1...
+    pub fn write(e: Env, starting_index: u32, num_entries: u32, size_kilo_bytes: u32) {
+        if num_entries == 0 || size_kilo_bytes == 0{
```
dmkozh, line 16: nit: Could you please run `cargo fmt` on this? I suppose test Wasms are currently ignored by the make file formatting...

soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+            e.storage().persistent().extend_ttl(&i, EXTEND_LEDGER, EXTEND_LEDGER);
+        }
+    }
+
+    pub fn do_work(e: Env, guest_cycles: u64, host_cycles: u64)
+    {
+        // We just want to do a lot of math, overflow is fine
+        let mut val = 1.3;
+        for _ in 0..guest_cycles {
+            val = val * 1.15;
```
dmkozh, line 37: Have you actually somehow managed to get this working? Floating point math should be prohibited (you shouldn't even be able to upload this - let me know if you could). Also overflows will cause panics, so that will fail as well. Given that these are guest ops, there shouldn't be difference between addition and multiplication, so you can just add loop counter to the integer (I'd hope compiler isn't smart enough to optimize that).

soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        // The more computationally expensive the host function,
+        // the more we overestimate its instruction cost. To require
+        // the most actual CPU work, we loop over a cheap host function
+        let mut u256_val = U256::from_u32(&e, 3);
+        let multiple= U256::from_u32(&e, 3);
+        for _ in 0..host_cycles {
+            u256_val = u256_val.mul(&multiple);
```
dmkozh, line 46: This might fail due to overflow as well, so better to add some numbers.

soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
SirTyson: Hmm, I've been testing with the `InvokeFostFunctionTests` framework and it seems to work, as in a call to `do_work` succeeds, and if I increase the cycle count if fails due to resource limit being exceeded. Maybe we skip some checks in the unit tests? I haven't tried actually submitting it to testnet.

jayz22 (commented):
Aside from @dmkozh 's points on overflowing and floating point, which are valid. Want to point out we've configured the compiler to aggressively optimize on contract size, so a contract for loadgen purpose that does a bunch of meaningless work will have a high chance of the computation work getting optimized out completely because the result is not being used anywhere. Please double check your compiled contract actually does produces load as you expected. One way to get around that is to return the result of whatever you computed in the loop. 
A better way to do these type of special case contracts is to use `soroban-synth-wasm` and write the wasm instructions explicitly. We have a bunch of examples in `soroban-env-host/test/utils.rs`.

dmkozh:
> A better way to do these type of special case contracts is to use soroban-synth-wasm and write the wasm instructions explicitly. We have a bunch of examples in soroban-env-host/test/utils.rs.

I would much rather prefer a readable/easy to modify contract. Synth wasm is great for low-level tests, but this is much higher level than that. Also, I doubt host fns can be optimized away - that would be pretty disastrous, right? (most of the contracts can be condensed to 'write a value to storage via a host function').

jayz22:
> > A better way to do these type of special case contracts is to use soroban-synth-wasm and write the wasm instructions explicitly. We have a bunch of examples in soroban-env-host/test/utils.rs.
> 
> I would much rather prefer a readable/easy to modify contract. Synth wasm is great for low-level tests, but this is much higher level than that. Also, I doubt host fns can be optimized away - that would be pretty disastrous, right? (most of the contracts can be condensed to 'write a value to storage via a host function').

I was mostly referring to multiplying a local number a number of times and throwing it away. Host function calls should be fine. 
Yeah this approach is fine, I just wanted to point out some potential surprises.

SirTyson:
> > > A better way to do these type of special case contracts is to use soroban-synth-wasm and write the wasm instructions explicitly. We have a bunch of examples in soroban-env-host/test/utils.rs.
> > 
> > 
> > I would much rather prefer a readable/easy to modify contract. Synth wasm is great for low-level tests, but this is much higher level than that. Also, I doubt host fns can be optimized away - that would be pretty disastrous, right? (most of the contracts can be condensed to 'write a value to storage via a host function').
> 
> I was mostly referring to multiplying a local number a number of times and throwing it away. Host function calls should be fine. Yeah this approach is fine, I just wanted to point out some potential surprises.

I've updated the test such that the return has a dependency on both the host function value and the guest value, so neither value should be optimized out. There's also branch inside the the guest loop, so I don't think the compiler will be able to do any fancy loop elision either.

soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+        let mut val: u32 = 0;
+        for _ in 0..guest_cycles {
+            if u32::MAX == val {
+                val = 0;
+            }
+
+            val += 1
+        }
+
+        // The more computationally expensive the host function,
```
dmkozh, line 42: I think we could as well overestimate the small functions and the end result would be similar (what matters in the end is the ratio between the actual and estimated cost. The code is fine, but I would probably drop this comment. I guess the main benefit here is that we can setup the loadgen in a pretty granular fashion.

dmkozh (approved):

jayz22 (approved):
Looks good, thanks!

Issue #1145, opened 2023-10-31
[test] over-large or otherwise invalid wasms


graydon:
A variety of places in wasm allow over-large or otherwise invalid inputs. Many of these will be rejected before they make it to us, and many will be rejected by wasmi itself. But we should have a test that checks we reject as many as we can think of ourselves:

- [x] corrupt metadata custom section
- [x] floating point instructions
- [ ] SIMD instructions, other post-MVP extensions
  - [x] multi-value 
- [x] very large import table that imports the same host functions over and over
- [x] importing nonexistent host functions
- [x] multiple memories
- [x] exporting nonexistent functions
- [ ] invalid table entries
  - [x] FuncRef
  - [ ] ExternRef 
- [x] start functions (these are banned)
- [ ] invalid const-expressions and active table references in elements table
- [ ] invalid const-expressions in active data segment mode
- [x] multiple data-count sections, sections with an invalid count
- [ ] very large versions of every section type:
  - [x] `Custom`
  - [ ] `Type`
  - [x] `Import`
  - [ ] `Function`
  - [ ] `Table`
  - [x] `Memory`
  - [x] `Global`
  - [x] `Export`
  - [x] `Start`
  - [x] `Element`
  - [x] `Code`
  - [x] `Data`
  - [x] `DataCount`

dmkozh:
A general comment for this issue is that if something can't be statically excluded, then we should cover it via a respective contract call.

Also, a few additions:

- [ ] Invalid host function imports (wrong number/type of args )
- [ ] Invalid contract function exports (wrong number/type of args)
- [ ] Incorrect argument types for host function call (i32 instead of i64) - likely runtime only

PR #1193, opened 2023-11-10
Add wasm test for excessive initial memory and table size

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/896

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-common/src/error.rs
+++ b/soroban-env-common/src/error.rs
@@ -216,32 +216,42 @@ impl From<wasmi::errors::FuncError> for Error {
 #[cfg(feature = "wasmi")]
 impl From<wasmi::Error> for Error {
     fn from(e: wasmi::Error) -> Self {
-        const BUDGET_EXCEEDED_LIMIT: Error =
+        const EXCEEDED_LIMIT: Error =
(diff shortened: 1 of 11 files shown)


graydon (changes requested):
Good start! But I think it is missing treatment of large wasm data sections, large data_count values, large element counts in other table types, and large numbers of sections that allow multiple sections.

jayz22:
> Good start! But I think it is missing treatment of large wasm data sections, large data_count values, large element counts in other table types, and large numbers of sections that allow multiple sections.

@graydon I've added tests for large data size, large number of data segments. I think this should be enough to close this issue? 

The other things you mentioned
- large element counts in other table types. The resource limiter should not differentiate between table element type. 
- large numbers of sections that allow multiple sections. Other sections like multiple CustomSections, CodeSections are not limited by the resource limiter (they should just increases wasm size and increase vm instantiation cost).
 
We can explore them further in https://github.com/stellar/rs-soroban-env/issues/1145 . Wdyt?

graydon (approved):

PR #1197, opened 2023-11-11
Hook envbase and vec slices

graydon:
This is a followup to #1142

  - Adds hook calls to EnvBase methods that previous instrumentation missed
  - Moves some macros around to clean things up in the host
  - Changes some builtin-SDK and SAC code to use the more efficient vec_new_from_slice path (this was just an omission)
  - Updates observations: generally they got longer (more hooks to observe at) but produced fewer objects (more efficient)


soroban-env-host/src/builtin_contracts/base_types.rs
```diff
@@ -469,6 +465,12 @@ impl Vec {
     }
 }
 
+macro_rules! vec_new {
```
dmkozh, line 468: I think this should be reconciled with `host_vec![]` macro in test utils. I don't have a strong opinion on the final naming, but I would slightly prefer `[]` syntax to follow the SDK closer (which is for we generally tried to achieve with the builtin contracts). Though probably cleanup can wait for a followup PR to make this one simpler.

dmkozh (approved):

soroban-env-host/src/builtin_contracts/base_types.rs
graydon: done

PR #1201, opened 2023-11-13
Make Host::set_lifecycle_event_hook pub

leighmcculloch:
### What
Make Host::set_lifecycle_event_hook pub.

### Why
For use in the SDK to record host calls for snapshotting and tracking behavior consistency over time in contracts.


sisuresh (approved):

leighmcculloch:
Not going ahead with as I'm not going to use it in the SDK now. Instead going to have the SDK focus on recording observable contract outputs and not all observable behavior.

Issue #1155, opened 2023-11-01
[test] type and argument checking on host function dispatch


graydon:
- [ ] for each host function:
  - [ ] invoke with wrong-typed args should fail
  - [ ] invoke with wrong-number of args should fail
  - [ ] invoke with an unknown object handle should fail

jayz22:
>  invoke with wrong-number of args should fail

I'm not sure how this can be done. The wasm code to call a function will have its type already encoded and will load the number of arguments based on the arity, so if we were to leave an extra (or one less) operand on the stack before `call` it would just make the wasm module invalid.

PR #1207, opened 2023-11-14
`contract_event` reject invalid topic handle

jayz22:
### What

Discovered while doing https://github.com/stellar/rs-soroban-env/issues/1155

### Why

Reject it early such that it does not proceed until trying to externalize the events at the very end, which will fail anyway.

### Known limitations

[TODO or N/A]


dmkozh (approved):

jayz22:
This has bit-rotten, will be continued in https://github.com/stellar/rs-soroban-env/pull/1213

PR #1209, opened 2023-11-14
Fix #1175 error code spoofing

### LINKED ISSUE #1175: Error-code spoofing
When a contract returns an error the ABI doesn't differentiate the guest-level values `Ok(Error)` from `Err(Error)` from `Error`. This is fine and expected, but `with_frame` in the host attempts to "fix" this by funneling `Ok(Error)` into `Err(HostError)` back in host-land, and this makes it possible for user code to "spoof" error codes -- return things other than `ScErrorType::ContractError`.

For most error codes this is at best a minor source of confusion, but we have some paths that care about the difference, eg. `try_call` decides whether an error is recoverable or unrecoverable based on the error type, and also we just generally want `ScErrorCode::InternalError` to only represent "there's a problem in the host's own logic" unambiguously.

Naturally the fuzzer discovered this game immediately: it can generate `InternalError` by writing a contract consisting of the constant value `InternalError`.

The fix for this is probably to get `with_frame` out of the business of massaging `Ok(Error)` returns and have `call`, `try_call` and whatever else emulates calling and error-propagation in the native-contract case look at the `Ok(Error)` cases they specifically want to admit -- `Ok(ScErrorType::ContractError)` alone -- and map that to `Err(...)`, and the rest into different `InvalidAction` codes or something.

https://github.com/stellar/rs-soroban-env/blob/6f69811f85364466fc4151ce68af6351ae78134a/soroban-env-common/src/arbitrary.rs#L14-L21

graydon:
I _think_ this is actually all that we need (or should) do for #1175. It's a little less than I was expecting to change: in particular I kept the adjustment in `with_frame`, and just tightened its condition (and commented why).

There are more call paths that go through `with_frame` than I expected and adjusting all of them is a bit more repetitive and introduces a bit more risk of drift between them than I think warrants the risk of accidentally over-adjusting an error unintentionally here (eg. if someone uses with_frame to pass an error we _don't_ want adjusted this way for some reason).

Nothing is ideal in this story, but I think this is a conservative fix that at least avoids the problem we were concerned with. @dmkozh wdyt? I will tack on a test that exercises this if you concur.


soroban-env-host/src/host/frame.rs
```diff
@@ -361,8 +361,45 @@ impl Host {
 
         let res = f();
         let mut res = if let Ok(v) = res {
+            // If a contract function happens to have signature Result<...,
+            // Code> its WASM ABI encoding will be ambiguous: if it exits with
+            // Err(Code) it'll wind up exiting the WASM VM "successfully" with a
```
dmkozh, line 366: nit: Wasm

dmkozh (commented):
This looks good to me, so just looking forward to tests.

soroban-env-host/src/host/frame.rs
graydon: fixed

graydon:
@dmkozh added test (and regenerated wasms, once to update Cargo.toml to new deps from previous PR, and again separately to make it clear which things changed with the new test)

dmkozh (approved):

PR #1214, opened 2023-11-15
Do not error if no authorizations

leighmcculloch:
### What
When asking the Host for a list of authenticated authorizations do not generate an error if there are none.

### Why
At the moment when asking the host for information about any auths that have been seen it generates and records an error if no auth was setup. This is inconvenient because this function is primarily called by tests, but the generation and recording of the error goes into the diagnostic log and muddies the logs for a developer. It makes it look like the contract had some problem where-as it was simply a test utility called at a time when it had no data.

There's little benefit in returning an error when no auth exists. Instead we can simply return an empty list. No auth was setup, no auth occurred, therefore an empty list showing that no auth occurred can be returned.

I encountered this as a problem when trying to store auth snapshots in the Rust SDK. At the end of a test I ask the host for auth info, but that causes a diagnostic error to be written into the events for every test that does nothing with auth.


soroban-env-host/src/auth.rs
```diff
@@ -2084,15 +2084,10 @@ impl Host {
...
-            .ok_or_else(|| {
-                self.err(
-                    ScErrorType::Auth,
-                    ScErrorCode::InvalidAction,
-                    "previous invocation is missing - no auth data to get",
-                    &[],
-                )
-            })?
-            .get_authenticated_authorizations(self))
+            .map(|am| am.get_authenticated_authorizations(self))
```
dmkozh, line 2087: This is a bit surprising - the intent here was that we only return error when there was no previous call (i.e. if you call this without calling any contracts at all). After the first invocation previous auth manager has to always be present. I don't mind approving this, but we might have a bit deeper issue here if the auth manager has really been missing after some invocations.

dmkozh (approved):

soroban-env-host/src/auth.rs
leighmcculloch: Ah so there folks can write SDK tests that never call a contract, and the Env Drop has no way of really knowing if that occurred or not, so it calls .auth() regardless.

Imo it should be fine to call .auth() without a contract call happening, and for it to return an empty list.

dmkozh: > Ah so there folks can write SDK tests that never call a contract
Ok, that's fine then. I was just worried that we might trigger this in between contract calls.

leighmcculloch: Would it be bad if we calls .auth() between contract calls?

PR #1217, opened 2023-11-15
Updated loadgen Wasm

SirTyson:
Consolidate loadgen wasm to a single function.


soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
```diff
@@ -4,32 +4,32 @@ use soroban_sdk::{contract, contractimpl, Bytes, Env, U256};
...
-        let slice = [0_u8; 1024];
+        let mut slice = [0_u8; 1024];
+        e.prng().fill(&mut slice);
         let mut bytes = Bytes::new(&e);
         for _ in 0..size_kilo_bytes {
             bytes.extend_from_slice(&slice);
         }
 
-        for i in starting_index..starting_index + num_entries {
+        for i in 0..num_write_entries {
```
dmkozh, line 29: I just wonder why did you remove the start index? I guess it doesn't really matter now, but would matter for parallelization eventually (as we'd want to use different ranges of keys)

dmkozh (approved):

soroban-test-wasms/wasm-workspace/loadgen/src/lib.rs
SirTyson: Initially, I had one instance and had TXs use different key ranges like you mentioned. However, now that we have one instance per account, each invocation can use the same keys with no collisions. I didn't think about parallelization when I took it out, but that's still pretty far down the road.

dmkozh: Ah, I see. Well, I guess then for parallelization we will want a bit of write key collisions, but I agree, that definitely can wait.

PR #1224, opened 2023-11-17
Observe more things

graydon:
This adds several new observation types to the test observer. The most important part is that it now observes the state of the auth manager, which is reports with two separate values: a (size,hash) pair for the current auth stack, and a (size,hash) pair for the sum of all the trackers. I _think_ this is probably the most useful way to eyeball changes, but @dmkozh should confirm (and maybe give a look at the recorded observation traces here to see if they make sense to you)


soroban-env-host/src/auth.rs
```diff
@@ -312,6 +313,16 @@ struct RecordingAuthInfo {
     disable_non_root_auth: bool,
 }
 
+#[cfg(test)]
+impl std::hash::Hash for RecordingAuthInfo {
```
dmkozh, line 317: While observing recording auth could have some marginal testing value, I'm not sure it should belong to observations (as almost everything else there concerns only observable protocol changes). So I wouldn't feel bad if we just skipped observing recording auth completely.
FWIW majority of auth tests works in the enforcing mode. We do have test cases that cover recording auth and I think these are good enough to make sure recording auth does hasn't diverged.

soroban-env-host/src/test/observe.rs
```diff
@@ -463,6 +455,67 @@ impl Host {
...
+        }
+    }
+    fn storage_footprint_size(&self) -> usize {
+        if let Ok(storage) = self.try_borrow_storage() {
+            storage.footprint.0.len()
+        } else {
+            0
+        }
+    }
+    fn storage_footprint_hash(&self) -> u64 {
```
dmkozh, line 501: I wonder if we could enforce the footprint immutability invariant instead of logging it into observations. Technically there is no need for us to eyeball the diffs if testing framework does verification on its own.
That's probably not necessary for this PR, just an idea.

soroban-env-host/src/auth.rs
graydon: the test observations are made without recording, it's just to make the hash derives compile

dmkozh: I see. Should it be simplified to a constant then, so that we're sure recording auth can never influence observations? Up to your preference though.

dmkozh (approved):

soroban-env-host/src/test/observe.rs
graydon: Oh, because the tests should never be recording footprint either?

soroban-env-host/src/auth.rs
graydon: I think I'd actually prefer to notice if we ever do!

soroban-env-host/src/test/observe.rs
dmkozh: The tests actually can/do record the footprint, but the recording behavior is not observable by the protocol. So removing this could reduce the noise a bit. In the enforcing mode we just expect the footprint to be immutable, so that would work better as assertion.

graydon: Mm. I think as with recording mode for auth, I'd prefer to _notice_ when we change it. It's not especially hard to update and while it's not divergence-critical, it's unlikely we'll want to have the recording behavior change by accident.

PR #1229, opened 2023-11-17
Pub ContractInvocationEvent

leighmcculloch:
### What
Pub ContractInvocationEvent.

### Why
The contract invocation hook function exposes the type to importers, but because it isn't public they can't reference it to do things like matching.

### DIFF
--- a/soroban-env-host/src/lib.rs
+++ b/soroban-env-host/src/lib.rs
@@ -48,8 +48,8 @@ pub use host::testutils::call_with_suppressed_panic_hook;
 #[cfg(any(test, feature = "testutils"))]
 pub use host::ContractFunctionSet;
 pub use host::{
-    metered_map::MeteredOrdMap, metered_vector::MeteredVector, Host, HostError, LedgerInfo, Seed,
-    SEED_BYTES,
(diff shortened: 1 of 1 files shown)


soroban-env-host/src/lib.rs
```diff
@@ -48,8 +48,8 @@ pub use host::testutils::call_with_suppressed_panic_hook;
 #[cfg(any(test, feature = "testutils"))]
 pub use host::ContractFunctionSet;
 pub use host::{
-    metered_map::MeteredOrdMap, metered_vector::MeteredVector, Host, HostError, LedgerInfo, Seed,
-    SEED_BYTES,
+    metered_map::MeteredOrdMap, metered_vector::MeteredVector, ContractInvocationEvent, Host,
```
dmkozh, line 51: This should be guarded by `#[cfg(any(test, feature = "testutils"))]`

leighmcculloch: [699a4cc](https://github.com/stellar/rs-soroban-env/pull/1229/commits/699a4ccfe276005168b11a00f00c0a58bcca9dd6)

dmkozh (approved):

PR #1230, opened 2023-11-17
Pass ref to host to contract invocation hook

leighmcculloch:
### What
Pass ref to host to contract invocation hook.

### Why
So that hook implementations have access to the host without needing to hold a reference to the host via some other means.

This is the pattern used on other hooks in the Host.

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -96,7 +96,7 @@ pub enum ContractInvocationEvent {
 }
 
 #[cfg(any(test, feature = "testutils"))]
-pub type ContractInvocationHook = Rc<dyn Fn(ContractInvocationEvent) -> ()>;
+pub type ContractInvocationHook = Rc<dyn for<'a> Fn(&'a Host, ContractInvocationEvent) -> ()>;
 
 #[derive(Clone, Default)]
 struct HostImpl {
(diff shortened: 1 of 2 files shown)


dmkozh:
I've intentionally omitted host to keep things simple - we use the lifecycle hooks in different context and don't expose these. You probably would want to capture the whole env with this, so passing the host seems redundant.

leighmcculloch:
If the impl of the hook holds a ref to the Host, won't the Host will live forever and never get dropped because its strong count will never reach zero?

dmkozh:
> If the impl of the hook holds a ref to the Host, won't the Host will live forever and never get dropped because its strong count will never reach zero?

I'm not sure why should the hook hold a strong reference to the host (if you capture it by reference); also it seems like it would be possible to mess up even if the host reference is passed as an argument (if you're not careful with how the env is captured). But I haven't played much with closures and borrow checker, so I could be missing something here.
Anyway, if you feel like this is necessary, you also need to update the hook call-sites to pass host.

dmkozh (approved):

leighmcculloch:
Admittedly I didn't test it using my own auth ref captured by the hook impl, but I don't see how there wouldn't be a cyclic ref here.

Host holds the hook, then hook holds a copy of the host that it captured (cloned). The ref count won't reach 0, unless there's some mechanism to handle that in Rust.

dmkozh:
> Admittedly I didn't test it using my own auth ref captured by the hook impl, but I don't see how there wouldn't be a cyclic ref here.

I've been rather thinking about capturing the reference, but that simply won't compile. So I guess that's the way to go...

PR #1232, opened 2023-11-19
Run user seeds to PRNGs through HMAC-SHA256 to unbias

### LINKED ISSUE #1170: set_base_prng_seed does nothing to prng state as viewed by fn passed to with_test_contract_frame

2674d867d7c6aa4212abab05ff30e5804ff1db90

### What did you do?

Called `set_base_prng_seed` before using the prng inside a function passed to `with_test_contract_frame`.

I expected to see calling `set_base_prng_seed` to change the results I get back from the prng.

### What did you see instead?

No change. Setting the base prng seed appears to have no impact on contract code running inside a `with_test_contract_frame`.

### Example

This is an example written in the SDK. The call to `as_contract` calls `with_test_contract_frame` with the given function.

```rust
#[test]
fn test_prng_seed() {
    let e = Env::default();
    e.host().set_base_prng_seed([0; 32]).unwrap();
    let id = e.register_contract(None, TestPrngContract);
    e.as_contract(&id, || {
        assert_eq!(e.prng().u64_in_range(0..=90), 78);
    });

    let e = Env::default();
    let id = e.register_contract(None, TestPrngContract);
    e.host().set_base_prng_seed([2; 32]).unwrap();
    e.as_contract(&id, || {
        assert_eq!(e.prng().u64_in_range(0..=90), 78);
    });
}
```

graydon:
This addresses #1170 by running user input to the PRNG seed function through HMAC-SHA256. As far as I can tell this is at least .. plausibly recommended practice (it's hard to find discussion of "how to use ChaCha correctly as a PRNG" that is not primarily concerned with applications involving secrets and therefore strong entropy).

I would greatly appreciate input from @kwantam or someone who has better knowledge about CSPRNGs or, specifically, properties-of-ChaCha than myself. 

Fixes #1170


graydon:
(Perhaps @nano-o knows something about this subject?)

soroban-env-host/src/host/prng.rs
```diff
@@ -98,8 +103,64 @@ impl Prng {
...
+    // exactly why this helps, considering ChaCha is itself a hash construct
+    // that's supposed to diffuse all entropy in every input (key & nonce) bit
+    // to every output bit, but it seems to be a common pattern.
+    //
+    // Given all this, we're going to run any user-provided seed through SHA256
+    // before using it as a ChaCha key. In production we should always be
+    // receiving a seed that's an output from SHA256 of ledger and transaction
+    // context information, but this adds a layer of defense against the
+    // possibility that we're given a nonuniform seed (which happens in testing
+    // at very least, and users might also manually reseed with a biased seed).
```
kwantam, line 160: I would not use SHA-256 for this purpose.

The "best possible" property you could ask for here is: given an input with `k` bits of entropy, you want to output a 256-bit string with (negligibly less than) `min(256, k)` bits of entropy, with the additional property that an adversary has (negligibly more than) `2^-k` chance of distinguishing the output from uniform. In other words: you want to conserve the input's entropy, but get an output that's as close as possible to uniform.

This isn't what SHA-2 was designed to do, and to my knowledge there's no formal analysis of SHA-2 that would give this property other than simply by regarding SHA-2 as a random oracle. In other words: the only way to show that SHA-2 is good enough for this task is simply to assume that it is.

But the good news is that there has been a lot of work on ways to implement randomness extractors using functions like SHA-2---and it turns out to be pretty straightforward! If you only read one paper on this topic, read Krawczyk's [analysis of HKDF](https://eprint.iacr.org/2010/264.pdf). It's probably also worthwhile to read Hugo's notes about HKDF in [RFC5869](https://www.rfc-editor.org/rfc/rfc5869).

---

The fix, as you may have guessed, is to use something like HKDF. In fact, you could just use RFC5869 directly:

1. Sample a fixed, public 32-byte random value, `salt`. Hard-code this value in the source code.
2. Fix some domain separator string, `info`. Perhaps something like `soroban::new_prng_from_seed` would work.
3. Instead of computing `sha256(input)`, compute `hkdf-sha256(salt, info, input, 32)`.

The above requires 4 SHA-256 invocations (a.k.a. two HMAC-SHA-256 invocations). It is almost certainly safe to cut this in half:

1. Sample `salt` as above.
2. Instead of computing `sha256(input)`, compute `hmac_sha256(key=salt, message=input)`.

There's nothing cryptographically dangerous about this, but it does mean that you lose the degree of freedom that the `info` string gives you. So if you, say, wanted to derive two independent keys from the same user input, that's slightly harder / less efficient. But if you don't think that's a downside in this context, HMAC is good enough.

---

There may be some political question around selecting the random seed: extremely paranoid users might suspect that you've somehow cooked the seed to your advantage (never mind that this is unlikely to be possible).

So, if necessary, you could select the seed in a "nothing up my sleeve" way by computing, e.g., `hmac_sha256(key=0, message=<first million digits of pi as ascii>)`.

kwantam (commented):
Definitely a good idea to unbias user seeds, but using SHA-256 by itself probably isn't good enough. The fix is pretty easy, though! See below.

soroban-env-host/src/host/prng.rs
graydon: @kwantam Ok, I think I will just go with the simpler/cheaper "just an hmac with a fixed key" option; is it acceptable to make the fixed key itself the output of a process a bit simpler than generating a million digits of pi? i.e. can I set it to `sha256("nothing up our sleeve")` or so (assuming I check to see that that has a fairly good balance of zeroes and ones)? Is there some standard sense of what makes a "public random value" appropriate in this role?

graydon: (Actually I just tried and it's fairly trivial to write a python program that calculates hash-of-pi-digits so if that is in any way more robust to assumptions I'm happy to use that too; whatever will raise fewer eyebrows)

kwantam: This is more a political than a technical decision---I'm sure that `sha256("nothing up our sleeve")` would be perfectly fine to use as a salt.

If someone is seriously intent on sniffing out a nonexistent conspiracy, any choice you make here will be insufficient. But it seems like "we chose a large number of digits from a well-known constant" is pretty hard to find evil intent in. (I guess there's still the "ah but why a million and not a thousand digits" style of gotcha...)

kwantam: Ah, sorry, to answer your first question: there's nothing in particular that makes the choice of one salt better than another, modulo the nothing-up-my-sleeve question and assuming that the choice is reasonably close to uniform.

In an ideal world you'd choose a value that other folks aren't likely to choose (because this makes your KDF independent of their KDF). Who knows whether HMAC of the first million digits of pi with the zero key meets that criterion. Maybe HMAC on "first million digits of pi" as the data with "hash of the [millionth stellar ledger](https://horizon.stellar.org/ledgers/1000000)" as the key does a better job on the uniqueness axis. But now again you've got to deal with people deciding there really is something up your sleeve...

graydon: Thanks @kwantam for all the help! I've updated to a fixed-salt-HMAC. Could you give one last glance at https://github.com/stellar/rs-soroban-env/pull/1232/commits/d8bc65dc91c6d77d57d747b96462cf434898bf46#diff-cc7d8d1f0039fd49ba0a0275f27a4e866fb7e9c800519d2e02020b79f1045c16R244-R287 to double check that I'm not saying or doing something egregiously wrong?

soroban-env-host/src/host/crypto.rs
```diff
@@ -238,3 +241,47 @@ pub(crate) fn chacha20_fill_bytes(
...
+    // reproducible way. In this case we use the Stellar Public Network ID,
+    // `sha256("Public Global Stellar Network ; September 2015")`.
+    //
+    // This number as a bitstring has 137 zeroes and 119 ones, which is within
+    // the range we get when taking 32-byte samples from /dev/random (feel free
+    // to check this yourself).
+
+    const SALT: [u8; 32] = hex!("7ac33997544e3175d266bd022439b22cdb16508c01163f26e5cb2a3e1045a979");
+
+    // Running HMAC will will run SHA256 2 times.
```
dmkozh, line 278: typo: will

soroban-env-host/src/host/crypto.rs
```diff
@@ -238,3 +241,47 @@ pub(crate) fn chacha20_fill_bytes(
...
+    budget
+        .as_budget()
+        .charge(ContractCostType::ComputeSha256Hash, Some(2))?;
```
dmkozh, line 281: You're charging for hashing 2 bytes here. Should probably be `bulk_charge` with two iterations, or just two charges for the expected payload sizes (I'm not sure about HMAC implementation details).

soroban-env-host/src/host/crypto.rs
```diff
@@ -238,3 +241,47 @@ pub(crate) fn chacha20_fill_bytes(
...
+
+    let mut hmac = Hmac::<Sha256>::new_from_slice(&SALT)
+        .map_err(|_| Error::from_type_and_code(ScErrorType::Context, ScErrorCode::InternalError))?;
```
kwantam, line 284: nit: `Hmac::new_from_slice` is infallible. This function returns a `Result` because the trait allows for the possibility that a key could be a bad length, but for HMAC any length is valid. So in principle you could `.expect("valid HMAC key length")` safely here. Not sure if that jibes with the local style, though :)

kwantam (approved):
:+1:

soroban-env-host/src/host/crypto.rs
graydon: oops, fixed

soroban-env-host/src/host/crypto.rs
graydon: nice catch, fixed

soroban-env-host/src/host/crypto.rs
graydon: yeah we're being extra paranoid about panics and turning them all into "internal errors" to minimize the odds of a stray unwind due to a mistake.

dmkozh (approved):

PR #1237, opened 2023-11-20
Add tests for de-serializing deep XDR in deep call stack.

dmkozh:
### What

Since we always convert decoded XDR to host objects, the end result of de-serialization is guarded by the host depth limit. However, we still want to make sure there is no stack overflow when the intermediate XDR object is deep.

### Why

Improving test coverage

### Known limitations

N/A


jayz22:
The meta files are pretty huge, maybe we skip generating them?

dmkozh:
> The meta files are pretty huge, maybe we skip generating them?

This is the only way to observe the depth-related invariants, so I wouldn't be comfortable removing them. These will stay unchanged as long as the Wasm is unchanged, so I don't see the file size as a big issue here.

graydon (approved):

PR #1240, opened 2023-11-21
Rename extend host functions

### LINKED ISSUE #1134: Rename extend to extend_ttl
Following [this Discord thread](https://discord.com/channels/897514728459468821/1159192183870607452), it seems that `extend_ttl` is preferred for host function names instead of just `extend`. We should rename the host functions accordingly. Also, we should consolidate our interface such that all host functions speak in terms of TTL instead of `liveUntilLedgerSeq`. I believe we already do this everywhere except for `get_max_live_until_ledger()` We should change `get_max_live_until_ledger()` to `get_max_ttl()`.

SirTyson:
### What

Awhile back we renamed the extend family of functions in the SDK. This renames the host functions accordingly for consistency.

Resolves #1134.


dmkozh:
You need to bump pre-release version in `soroban-env-common/meta.rs`. That also means that you'll need to regenerate the test wasms, which will also mean that you need to update observations. We've actually planned to switch to pre-release 0, so this check won't be needed anymore, so maybe it would be wiser to just wait a day or two before merging this PR.

dmkozh:
> You need to bump pre-release version in `soroban-env-common/meta.rs`. That also means that you'll need to regenerate the test wasms, which will also mean that you need to update observations. We've actually planned to switch to pre-release 0, so this check won't be needed anymore, so maybe it would be wiser to just wait a day or two before merging this PR.

We've set pre-release to 0, so you shouldn't need the update anymore.

PR #1244, opened 2023-11-22
Add Host::has_frame

leighmcculloch:
### What
Add a method to the host for determining if the host currently has a frame on the stack.

### Why
In several places in the SDK it would be ideal if we could error if a function is used inside or outside of a contract function, because doing so would be invalid.

In some code being added or auth tracking in the SDK I need to reliable know if we are inside a contract function or not.

I tried to do this using top invocation hooks, but the way native contract functions get called in the SDK doesn't pass through anything from the Env that is available outside of the contract. We can sort of know if we're in a contract, but it isn't guaranteed to be correct in edge cases. We might change that some day, but at the moment that limitation exists.

Exposing whether the host currently has a frame is a simple way the host can indicate to SDK test utilities if it is currently in a contract function or not.

Close https://github.com/stellar/rs-soroban-sdk/issues/1084


soroban-env-host/observations/test__frame__in_frame.json
```diff
@@ -0,0 +1,6 @@
+{
```
dmkozh, line 1: Where does this file come from?

dmkozh (approved):

soroban-env-host/observations/test__frame__in_frame.json
leighmcculloch: It came from the has_frame test, when it was named in_frame. I'm curious why the CI build is letting me add a json file for a test that doesn't exist. We should probably have something like this: https://github.com/stellar/rs-soroban-sdk/blob/8a6ea7ede02f1cb06fda27e879c92d411d912fe1/.github/workflows/rust.yml#L97-L113

dmkozh: Yeah, I'm pretty sure we just don't have a check for this yet.

leighmcculloch: I'll fix the json file before merging.

Issue #1162, opened 2023-11-01
[test] bytes and serialization


graydon:
A bunch of tests should get added to tests/bytes.rs:

- [ ] structurally amplify some host values (a busy vec-of-vecs-of-vecs say) and see that it fails to serialize
- [ ] amplify some bytes (doubling a large bytes object) and see that it runs out of memory and CPU budget (each separately)
- [ ] amplify some bytes to be very large input to XDR deserialization, check that it fails to deserialize
- [ ] check that corrupt XDR fail to deserialize
- [ ] check that XDR that lies about the length of fields fails to deserialize
- [ ] do more roundtrip types (connect to XDR `Arbitrary` implementation, run in a loop with a random seed)

dmkozh:
> structurally amplify some host values (a busy vec-of-vecs-of-vecs say) and see that it fails to serialize

Covered by https://github.com/stellar/stellar-core/pull/4041 quite comprehensively.

dmkozh:
>  amplify some bytes (doubling a large bytes object) and see that it runs out of memory and CPU budget (each separately)

Covered by https://github.com/stellar/rs-soroban-env/pull/1247

PR #1247, opened 2023-11-22
Add tests for invalid maps, oversized maps/vecs/bytes from various paths

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/1148 and partially https://github.com/stellar/rs-soroban-env/issues/1162

### Why

### Known limitations

[TODO or N/A]


soroban-env-host/src/test/map.rs
```diff
@@ -372,3 +369,182 @@ fn large_map_exceeds_budget() {
...
+    let scv_bytes = ScVal::Bytes(buf.try_into().unwrap());
+    let bytes_val = host.to_host_val(&scv_bytes)?;
+
+    let vals = vec![bytes_val; 5];
+    let keys = ["a", "b", "c", "d", "e"];
+
+    let map = host.map_new_from_slices(&keys, &vals.as_slice())?;
+    let res = host.from_host_val(map.to_val());
+    // This is actually Budget::ExceededLimit, but ScVal::try_from_val converts
+    // he error to a ConversionError. We have an issue for it https://github.com/stellar/rs-soroban-env/issues/1046
```
dmkozh, line 399: We probably should triage this issue and either fix it now, or close it. I think it's a bit better to have a budget error, but if the fix takes more than updating a couple lines of code, then it's probably not worth it.

soroban-env-host/src/test/map.rs
```diff
@@ -372,3 +369,182 @@ fn large_map_exceeds_budget() {
...
+
+    // large map from slices
+    let buf = vec![0; 7_000_000];
+    let scv_bytes = ScVal::Bytes(buf.try_into().unwrap());
+    let bytes_val = host.to_host_val(&scv_bytes)?;
+
+    let vals = vec![bytes_val; 5];
+    let keys = ["a", "b", "c", "d", "e"];
+
+    let map = host.map_new_from_slices(&keys, &vals.as_slice())?;
```
dmkozh, line 396: Should we also have large values as keys for completeness?

soroban-env-host/src/test/map.rs
graydon: The fix is unfortunately quite complex as it involves iterating on common and sdk code and macro code to find a stable trait arrangement. I tried once and was defeated. It's one of the things I decided we just don't have time or effort available for, but I feel bad about it.

graydon: (I've pinged @leighmcculloch in case he has time for it, I really wanted to find time before final to fix it but at this point I expect we're going to be stuck with this behaviour forever)

soroban-env-host/src/test/map.rs
jayz22: The keys must be symbols. 
But for completeness I've added a case where the key slice is too long.

dmkozh (approved):

Issue #1141, opened 2023-10-30
Saturating math -> checked math?


graydon:
A number of places in soroban (fees, metering, ledger arithmetic) use saturating arithmetic to avoid the risk of over/underflow. This is _often_ correct (or harmless) but not always -- it also opens up the potential for two problems:

  1. Violating intended arithmetic identities (a - b makes c, then c + b should make a, but .. maybe now it doesn't)
  2. Hiding bugs (a - b underflows because we let b get too big .. but now we don't notice)

The next-best alternative is the _checked_ variants of all the currently-saturating arithmetic. I think in most of the places we use saturating, checked would work as well or better. We might want to add a helper trait that provides abbreviations for `checked_{add,sub,mul,div}(...).ok_or(Error::from_type_and_code(ScErrorType::Value, ScErrorCode::ArithDomain))`?

graydon:
cc @MonsieurNicolas and also @dmkozh who no doubt has opinions about this in fees!

dmkozh:
We could do that, but I'm just not sure if that's the most useful thing to do (specifically for fees). It makes the bridge and core side implementation more complex, because now the fee computation becomes fallible and we need some code to handle that.  It also makes the library much more complex. 
Since the maximum valid fee has an upper bound that is way less than `i64::MAX`, it seems fine to me to have imperfect math for the edge cases because we can't include transactions that get to these cases. Underflows also don't seem possible.

MonsieurNicolas:
This seems fine. I guess we need the proper checks anyways outside of fee computation, so if we assume that no account can pay even "as little" as `i64::MAX` we're good

MonsieurNicolas:
so for fees I guess this is fine -- what about other places?

dmkozh:
> so for fees I guess this is fine -- what about other places?

A quick audit shows that we're otherwise using saturating arithmetic for budget arithmetic, which is a very similar case to the fees - we don't want it to be fallible because failure is kind of meaningless (if you reach u64::MAX instructions/memory charged, you should fail anyway due to running out of budget). The only remaining small case is ledger seq maths for the rent bumps, which has to be addressed *somehow* when we get to `u32::MAX` ledger; but we probably don't need to worry about that for now.

jayz22:
[...]

PR #1248, opened 2023-11-23
Re-calibration on x86 and update cost parameters

### LINKED ISSUE #1246: Recalibration before final build
- Recalibrate the cpu and mem cost coefficients on an x86 machine. 
- Use analytical ones for mem-related cost types (MemAlloc, MemCpy, MemCmp). 
- Update core.

jayz22:
### What

Resolves #1246  Also made a minor fix on changing the plain `*` to `saturating_mut` for map charges (mentioned in #1141).

This is the first calibration results on x86, previous results were obtained on M1. Overall most of the cost characteristics stay the same, there are some notable changes:
- The VmInstantiation cpu cost has gone down significantly, which explains cpu numbers in most test observations going down (by more than half). 
- ~The arithmetics related cost centers have gone up, this includes {U,I}{256, 128} arithmetics, cryptography functions.~ This was due to the host observation overhead, it has been fixed.
- The `InvokeVmFunction` is taking maximum arg size (=32).
- `ValSer` `ValSer` has been calibrated using closer-to-worst examples

The mem-related cost parameters are using analytical inputs (not calibrated), same as before.

[raw output](https://github.com/stellar/rs-soroban-env/files/13494985/output_x86_nov_28.txt)

### Why


soroban-env-host/src/budget.rs
```diff
@@ -233,19 +233,19 @@ impl Default for BudgetImpl {
                 // whatever function call, arg-shuffling, spills, reloads or
                 // other flotsam accumulates around a typical memory copy.
                 ContractCostType::MemCpy => {
-                    cpu.const_term = 250;
+                    cpu.const_term = 50;
```
dmkozh, line 236: Why did this change (given that the comments suggests that the value isn't based on calibration results)?

graydon: The linear term isn't calibration-derived, but the constant one is.

soroban-env-host/src/budget.rs
```diff
@@ -233,19 +233,19 @@ impl Default for BudgetImpl {
...
                     cpu.lin_term = ScaledU64((1 << COST_MODEL_LIN_TERM_SCALE_BITS) / 8);
                 }
                 ContractCostType::MemCmp => {
-                    cpu.const_term = 250;
+                    cpu.const_term = 50;
                     cpu.lin_term = ScaledU64((1 << COST_MODEL_LIN_TERM_SCALE_BITS) / 8);
                 }
                 ContractCostType::DispatchHostFunction => {
-                    cpu.const_term = 263;
+                    cpu.const_term = 1368;
```
graydon, line 244: I believe this is a mistake -- or rather a result of having `feature = "testutils"` enabled during testing, which is creating an (expensive) input to the host lifecycle hook. We need to un-measure this because it's a significant overcharge from what dispatching a host function will cost in production without `feature = "testutils"` (and if you notice below it also charges _memory_ which should not happen)

soroban-env-host/src/budget.rs
jayz22: That's right. The const term was previously derived from calibration. I just updated the number to the new calibration.

soroban-env-host/src/budget.rs
jayz22: Thanks for the good catch! I have split all calibration related code to a new feature "bench" such that it's decoupled from "testutils". Now the numbers for `DispatchHostFunction` and `I256` arithmetics have gone down to previous levels.

jayz22:
Thanks @dmkozh and @graydon for the review. I've refreshed the PR again after your comments. Now it should be ready to go again. 

One thing that's been updated since your last review was the calibration of xdr conversions (`ValSer`, `ValDeser`), previously we were was treating it like a mem op (using the same analytical numbers from memcpy). But that's because previously it was calibrated to a not-too-worst case input, which was just (de)serializing a large bytes array. I've since updated the samples to deeply nested structure trying to resemble the worst case, and have updated the parameters accordingly. The cpu cost parameters are directly taking from calibration, the memory cost parameters are still analytical but based on calibration result (~a factor of 3 bytes per input byte). I think this is the best we can do but curious what you guys think.

graydon (approved):
Great work, thanks!

PR #1253, opened 2023-11-27
Added test wasm writes with arbitrary key sizes

SirTyson:
### What

Adds test wasm functions that write `ContractData` entries with arbitrary key sizes.

### Why

This allows us to test network upgrades that reduce `maxContractDataKeySizeBytes`.


dmkozh:
This change is fine, but do we actually need it? This should be possible to cover by just passing the keys to the footprint, they don't have to be actually used.

SirTyson:
> This change is fine, but do we actually need it? This should be possible to cover by just passing the keys to the footprint, they don't have to be actually used.

Sure, I guess it's not strictly required, but I think this is the easiest way. We don't allow writing keys above the limit, but we still allow reading. For the write test we can just put a dummy key in the footprint and make sure it fails, but for the read test we need some value with a large key that actually exists for us to read. Given our current testing framework, it seemed easiest to write this value via a contract call instead of manually injecting it into the ledger.

dmkozh (approved):

dmkozh:
>  Given our current testing framework, it seemed easiest to write this value via a contract call instead of manually injecting it into the ledger.

Yeah, you're right, this probably will give us better coverage.

SirTyson:
Closing, after further discussion we're changing the behavior of ledger entry key size limit reductions and this test is no longer necessary.

PR #1254, opened 2023-11-27
Tighten up `saturating_` math checks

### LINKED ISSUE #1141: Saturating math -> checked math?

jayz22:
### What

Resolves #1141 

Several places benefit from changing `saturating_` to `checked_` math:
- `saturating_add(_mul)` on u32, saturation would be within possibility
- Some `saturating_sub` saturation would mean actual error, e.g index out of bound

In both cases, changing to `checked_` version means overflow errors are caught sooner. 

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/budget/limits.rs
+++ b/soroban-env-host/src/budget/limits.rs
@@ -96,7 +96,9 @@ impl DepthLimiter for BudgetImpl {
     // `leave` should be called in tandem with `enter` such that the depth
     // doesn't exceed the initial depth limit.
     fn leave(&mut self) -> Result<(), HostError> {
-        self.depth_limit = self.depth_limit.saturating_add(1);
+        self.depth_limit = self.depth_limit.checked_add(1).ok_or_else(|| {
+            Error::from_type_and_code(ScErrorType::Value, ScErrorCode::ArithDomain)
+        })?;
         Ok(())
     }
 }
--- a/soroban-env-host/src/budget/limits.rs
+++ b/soroban-env-host/src/budget/limits.rs
@@ -97,7 +97,7 @@ impl DepthLimiter for BudgetImpl {
     // doesn't exceed the initial depth limit.
     fn leave(&mut self) -> Result<(), HostError> {
         self.depth_limit = self.depth_limit.checked_add(1).ok_or_else(|| {
-            Error::from_type_and_code(ScErrorType::Value, ScErrorCode::ArithDomain)
+            Error::from_type_and_code(ScErrorType::Context, ScErrorCode::InternalError)
         })?;
         Ok(())
     }
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -954,10 +954,13 @@ impl VmCallerEnv for Host {
             let msg = String::from_utf8_lossy(&msg);
 
             let VmSlice { vm, pos, len } = self.decode_vmslice(vals_pos, vals_len)?;
-            Vec::<Val>::charge_bulk_init_cpy(len.saturating_add(1) as u64, self)?;
+            Vec::<Val>::charge_bulk_init_cpy((len as u64).saturating_add(1), self)?;
             let mut vals: Vec<Val> = vec![Val::VOID.to_val(); len as usize];
             // charge for conversion from bytes to `Val`s
-            self.charge_budget(ContractCostType::MemCpy, Some(len.saturating_mul(8) as u64))?;
+            self.charge_budget(
+                ContractCostType::MemCpy,
+                Some((len as u64).saturating_mul(8)),
+            )?;
             self.metered_vm_read_vals_from_linear_memory::<8, Val>(
                 vmcaller,
                 &vm,
@@ -1510,7 +1513,10 @@ impl VmCallerEnv for Host {
         Vec::<Val>::charge_bulk_init_cpy(len as u64, self)?;
         let mut vals: Vec<Val> = vec![Val::VOID.into(); len as usize];
         // charge for conversion from bytes to `Val`s
-        self.charge_budget(ContractCostType::MemCpy, Some(len.saturating_mul(8) as u64))?;
+        self.charge_budget(
+            ContractCostType::MemCpy,
+            Some((len as u64).saturating_mul(8)),
+        )?;
         self.metered_vm_read_vals_from_linear_memory::<8, Val>(
             vmcaller,
             &vm,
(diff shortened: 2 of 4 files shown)


soroban-env-host/src/budget/limits.rs
```diff
@@ -96,7 +96,9 @@ impl DepthLimiter for BudgetImpl {
     // `leave` should be called in tandem with `enter` such that the depth
     // doesn't exceed the initial depth limit.
     fn leave(&mut self) -> Result<(), HostError> {
-        self.depth_limit = self.depth_limit.saturating_add(1);
+        self.depth_limit = self.depth_limit.checked_add(1).ok_or_else(|| {
+            Error::from_type_and_code(ScErrorType::Value, ScErrorCode::ArithDomain)
```
dmkozh, line 100: This looks like an internal error to me - this could only ever occur if we call `leave` without `enter`.

soroban-env-host/src/host.rs
```diff
@@ -2416,8 +2425,10 @@ impl VmCallerEnv for Host {
         let vnew = self.visit_obj(b, |hv: &ScBytes| {
             self.validate_index_lt_bound(i, hv.len())?;
             let mut vnew: Vec<u8> = hv.metered_clone(self)?.into();
-            // len > i has been verified above but use saturating_sub just in case
-            let n_elts = (hv.len() as u64).saturating_sub(i as u64);
+            // len > i has been verified above but use checked_sub just in case
+            let n_elts = (hv.len() as u64)
+                .checked_sub(i as u64)
+                .ok_or_else(|| self.err_arith_overflow())?;
```
dmkozh, line 2431: This should be an internal error if we claim to guarantee the invariant in the comment above.

soroban-env-host/src/host/ledger_info_helper.rs
```diff
@@ -16,16 +16,18 @@ impl Host {
                 self.with_ledger_info(|li: &LedgerInfo| Ok(li.min_persistent_entry_ttl))?
             }
         };
-        Ok(ledger_seq.saturating_add(min_live_until.saturating_sub(1)))
+        ledger_seq
+            .checked_add(min_live_until.saturating_sub(1))
+            .ok_or_else(|| self.err_arith_overflow())
```
dmkozh, line 21: This should probably be an internal error as well, as it means that we probably have a bad misconfiguration with huge min TTL.

soroban-env-host/src/host/ledger_info_helper.rs
```diff
@@ -16,16 +16,18 @@ impl Host {
...
 
     pub(crate) fn max_live_until_ledger(&self) -> Result<u32, HostError> {
         self.with_ledger_info(|li| {
-            Ok(li
-                .sequence_number
+            li.sequence_number
                 // Entry can live for at most max_entry_live_until ledgers from
                 // now, counting the current one.
-                .saturating_add(li.max_entry_ttl.saturating_sub(1)))
+                .checked_add(li.max_entry_ttl.saturating_sub(1))
```
dmkozh, line 29: Same here, should be an internal error.

soroban-env-host/src/storage.rs
```diff
@@ -422,8 +422,11 @@ impl Storage {
             ));
         }
 
-        let mut new_live_until =
-            host.with_ledger_info(|li| Ok(li.sequence_number.saturating_add(extend_to)))?;
+        let mut new_live_until = host.with_ledger_info(|li| {
+            li.sequence_number
+                .checked_add(extend_to)
+                .ok_or_else(|| host.err_arith_overflow())
```
dmkozh, line 428: Same, should be an internal error (we verify that the value is sane above, so this would only occur due to mis-configuration/bug).

soroban-env-host/src/host.rs
jayz22: Good point, fixed.

soroban-env-host/src/host/ledger_info_helper.rs
jayz22: That wouldn't be "internal" error though. Like, if someone (say in their own test network) sets up a bad min TTL, it's not the host's internal logic bug. The error comes from "external" to the host and arithmetic overflow is the most accurate code to surface it back.

soroban-env-host/src/storage.rs
jayz22: The `extend_to` is also passed externally, see reply above.

soroban-env-host/src/host/ledger_info_helper.rs
jayz22: See reply above.

soroban-env-host/src/host/ledger_info_helper.rs
dmkozh: Any error caused by misconfiguration is currently an internal error, which makes sense: we don't really want such errors to be recoverable and we do want to catch them as early as possible. Otherwise, the errors represent the conditions caused by the host function user input. I realize that philosophically you may say that network updates are also caused by the 'user' actions, but these are not the same 'user' as the regular Soroban user. In any case, even in your scenario it would be beneficial to return an internal error here - it will quickly identify the misconfiguration instead of hiding it behind a recoverable error. FWIW misconfiguration in this particular case will completely brick the network.

soroban-env-host/src/storage.rs
dmkozh: `extend_to` is guarded by the configuration, and the misconfiguration should be an internal error.

soroban-env-host/src/host/ledger_info_helper.rs
jayz22: Sounds good, updated to `InternalError` and added a comment.

soroban-env-host/src/host/ledger_info_helper.rs
jayz22: Fixed.

soroban-env-host/src/storage.rs
jayz22: Fixed.

graydon (approved):

Issue #1222, opened 2023-11-16
Cargo version pinning improvements


graydon:
We're still not aggressive enough about making Cargo not change stuff underfoot. Plan:

 - [x] pass `--locked` to all cargo commands in every place we do anything in CI
 - [x] set all Cargo.toml version specs of our own to `=` constraints
 - [ ] add a make update-deps target that does `cargo +nightly -Z minimal-dependencies update` or so
 - [x] enable cackle and cargo-deny on core
 - [ ] update core's and env's deps using the same make update-deps and carefully compare solutions match
 - [ ] check that core and env both use cargo resolver v2

graydon:
Not going to do all of this -- we'll do the parts that represent "locking what we've got" (steps 1, 2 and 4) but not adopt minimal-deps or change resolver or anything like that.

leighmcculloch:
Why won't we change the resolver? It's a one line change.

graydon:
Because it'd be weird / misleading to say we're on resolver 2 but not re-resolve with it, and the point here is to not change any more deps at this point.

graydon:
Done all we're going to do here as of https://github.com/stellar/stellar-core/pull/4061 and https://github.com/stellar/rs-soroban-env/pull/1257

PR #1255, opened 2023-11-28
Cargo version pinning

graydon:
This does the env side of #1222 -- there are still core-side things to do but this is a step.

No packages are changed in the main lockfile here, just the one in the fuzzer sub-workspace which was out of sync.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,21 +19,21 @@ exclude = ["soroban-test-wasms/wasm-workspace"]
 version = "20.0.0-rc2"
 
 [workspace.dependencies]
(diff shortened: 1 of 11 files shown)


leighmcculloch (approved):
Looks good. 

Questions that don't need to impede merging:

I don't see us adding locking in the publish commands. Is that necessary?

Do we plan to pin transitive deps too some way? This change is only pinning direct deps.

graydon:
Transitive deps get pinned in the trees baked into core in the core build. We don't have anything doing that at the plain env layer. It might be nice, but the core build is the level that's most worried about minor deviation.

I don't know if locking is necessary for publishing, that's an interesting question. I assumed not. It looks like maybe it does support --locked also. I guess I will open another PR with that!

PR #1256, opened 2023-11-28
Cap number of args to wasm functions

graydon:
This just limits the number of args allowed to wasm functions to something reasonable (32).


leighmcculloch (approved):

leighmcculloch:
It'd be good if we caught limktd like this earlier in the SDK in codegen and CLI when building, but a limit of 32 is so high I don't think it's worth the maintenance cost of the logic to add to those things.

PR #1257, opened 2023-11-28
Pass --locked to publish as well

graydon:
Missed this in previous change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ reobserve-tests:
 	UPDATE_OBSERVATIONS=1 cargo test --locked -p soroban-env-host --features testutils --profile test-opt
 
 publish:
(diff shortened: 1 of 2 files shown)


dmkozh (approved):

leighmcculloch (commented):
Hmm. The files edited in this change don't actually get used to publish.

Publishing is a shared GitHub workflow that is started here, but lives in another repo: https://github.com/stellar/rs-soroban-env/blob/main/.github/workflows/publish.yml.

So I think what we need to do is edit the shared workflow.

And we should probably remove the publish from the Makefile and the script, since they should not be being used.

leighmcculloch:
Opened the following change to address the unused files:
- https://github.com/stellar/rs-soroban-env/pull/1258

PR #1269, opened 2023-11-30
Adjust versions for `proc-macro2`, `syn` and `quote`

jayz22:
### What

### Why

To be compatible with core's `cxx`

### Known limitations

[TODO or N/A]

### DIFF
--- a/cackle.toml
+++ b/cackle.toml
@@ -266,6 +266,7 @@ allow_unsafe = true
 build.allow_apis = [
     "env",
     "process",
+    "terminate",
 ]
 allow_unsafe = true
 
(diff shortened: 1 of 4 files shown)


cackle.toml
```diff
@@ -266,6 +266,7 @@ allow_unsafe = true
 build.allow_apis = [
     "env",
     "process",
+    "terminate",
```
dmkozh, line 269: This is a bit concerning, how come it appeared in the *older* version of the library and we didn't need it before?

jayz22: I didn't look into how this is being used and not sure how concerning this is. But I could roll back to the previous version that doesn't require it.

PR #1282, opened 2023-12-02
Bug #1178 - use Timepoint on get_ledger_timestamp

### LINKED ISSUE #1178: Inconsistent time types

2674d867d7c6aa4212abab05ff30e5804ff1db90

### What did you do?

Used `get_ledger_timestamp` to get timestamp.

https://github.com/stellar/rs-soroban-env/blob/e2f9c1defc261cb9981df258cda03d986fb121a6/soroban-env-common/env.json#L77-L83

A `Timepoint`.

### What did you see instead?

A `u64`.

### Discussion

We seem to have these two types `Timepoint` and `Duration` that get exposed all the way up in the SDK, however they don't seem to get used anywhere. Even the places in the host interface where one might think a time would be exposed like the `get_ledger_timestamp` function, they just use a plain u64.

What are `Timepoint` and `Duration` for? Should we remove them if they aren't in use? Or are there places that aren't using them should be using them?

If they have no use I'd like to remove them from the SDK to reduce the noise.

### History

It looks like timepoint and duration were added in:
- https://github.com/stellar/rs-soroban-env/pull/866

In response to this issue:
- https://github.com/stellar/rs-soroban-env/issues/724

According to this comment it sounds like the intent was to use Timepoint for the ledger timestamp:
> They can represent for example ledger timestamp and can specify its own display format

Ref:  https://github.com/stellar/rs-soroban-env/pull/866#issuecomment-1599010686

@jayz22 @graydon What should we do here?

graydon:
Fixes #1178 

This would be an env-breaking change if we do later. But it's also not _super_ important? We can paper it over in the SDK or ship a different function in the future.

### DIFF
--- a/soroban-env-common/env.json
+++ b/soroban-env-common/env.json
@@ -78,7 +78,7 @@
                     "export": "4",
                     "name": "get_ledger_timestamp",
                     "args": [],
-                    "return": "U64Val",
+                    "return": "TimepointVal",
                     "docs": "Return the timestamp number of the current ledger as a u64."
                 },
                 {
--- a/soroban-env-common/src/env.rs
+++ b/soroban-env-common/src/env.rs
@@ -5,8 +5,8 @@ use crate::Object;
 use super::Symbol;
 use super::{
     AddressObject, Bool, BytesObject, DurationObject, Error, I128Object, I256Object, I256Val,
-    I64Object, MapObject, StorageType, StringObject, SymbolObject, TimepointObject, U128Object,
-    U256Object, U256Val, U32Val, U64Object, U64Val, Val, VecObject, Void,
+    I64Object, MapObject, StorageType, StringObject, SymbolObject, TimepointObject, TimepointVal,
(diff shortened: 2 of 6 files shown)


graydon:
(Discussed with @leighmcculloch and decided this won't make it this time around, so maybe never, who knows)

leighmcculloch (approved):

leighmcculloch:
> (Discussed with @leighmcculloch and decided this won't make it this time around, so maybe never, who knows)

Sorry, I don't remember this. We must have been past the point where we were changing the API.

This relates to:
- https://github.com/stellar/rs-soroban-sdk/issues/1197

Can this PR be repurposed to adding a new host function that returns `Timepoint`? After it is released in the next protocol we can update the SDK to return `Timepoint` too.

Without this change `Timepoint` is hard to adopt since much of the time folks operate on "now" and modifications of now, and ledger close time is the closest thing to now that contract developers have.

PR #1284, opened 2023-12-02
Add fuzz target that runs wasmi on wasm-smith output

graydon:
This adds a new fuzz target that fuzzes wasmi -- including execution -- with wasms generated by wasm-smith.


MonsieurNicolas (approved):
Looking good

PR #1289, opened 2023-12-03
Make out-of-footprint accesses recoverable errors

graydon:
This is something that came up during review: I'm not sure out-of-footprint accesses should be considered unrecoverable since it's quite possible for a caller to try_call a callee that fails with such an out-of-footprint access and then recover and continue (as it would with, say, a logic bug in a contract). The other unrecoverable cases (internal error, out of budget) much more convincingly _can't_ be recovered from in the sense that either the contract has no idea how to or else it is literally impossible because budgets are one-way.

@leighmcculloch argues that perhaps the footprint-error cases should be unrecoverable because someone _invoking_ a transaction can independently control the failure, but I think this may be a stretch? I think invokers have lots of ways of causing pinpointed failures in contracts, both recoverable and non. At least it's no different than auth failures, which we allow recovery from. I would appreciate some discussion from @dmkozh and anyone else with opinions (he came up with the recoverable-vs-non distinction).

In case we _do_ want to make out-of-footprint accesses recoverable, this PR does so. If we want to leave things as-is, with unrecoverable out-of-footprint accesses, we can close this.

### DIFF
--- a/soroban-env-host/src/host/error.rs
+++ b/soroban-env-host/src/host/error.rs
@@ -140,9 +140,6 @@ impl HostError {
     }
 
     /// Identifies whether the error can be meaningfully recovered from.
-    ///
-    /// We consider errors that occur due to broken execution preconditions (
-    /// such as incorrect footprint) non-recoverable.
     pub fn is_recoverable(&self) -> bool {
         // All internal errors that originate from the host can be considered
         // non-recoverable (they should only appear if there is some bug in the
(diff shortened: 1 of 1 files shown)


graydon:
@dmkozh Hmm, that's .. not really true; or at least if you're proposing the notion of an "invalid transaction" as a static judgment that's intrinsically true or false about a given transaction.

The the footprint is one input to the possibility of an out-of-footprint error, but the decision to actually do an out-of-footprint access is a dynamic (and undecidable, as context-sensitive as any dynamic decision) aspect of the contract. You might as well treat every failed array bounds check in linear memory as unrecoverable because a transaction that dynamically chooses to do it must be invalid. The decision of which things to actually access happens dynamically while the contract runs, and might be context sensitive (depend on the exact ledger state it runs on, which preflight can be out of sync with).

Anyway, it seems there's a consensus that this is at least a dimension of a transaction's failure conditions that we don't want attackers to be able to vary independently from paths that tx authors intend to guard against. Which I guess is fair. I'll close this.

PR #1291, opened 2023-12-05
Turn off `wasm_reference_types` in Wasmi

jayz22:
### What

Disable wasmi `reference_types`and make all wasmi feature selections explicit.
Also did a pass-by clean up on the fuel configs code. 

### Why

`bulk_memory` and `reference_types` are both post-MVP features.
We use 1 extensively for object init/copy between host and guest. 
We do not use 2 at all. We are not using `ExternRef` and we are limiting table count to 1 (via resource limiter).

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/src/budget/wasmi_helper.rs
+++ b/soroban-env-host/src/budget/wasmi_helper.rs
@@ -1,55 +1,7 @@
-use super::AsBudget;
-use crate::{xdr::ContractCostType, Host};
-use wasmi::{errors, ResourceLimiter};
-
-/// This is a subset of `wasmi::FuelCosts` which are configurable, because it
-/// doesn't derive all the traits we want. These fields (coarsely) define the
-/// relative costs of different wasm instruction types and are for wasmi internal
-/// fuel metering use only. Units are in "fuels".
-#[derive(Clone)]
-pub(crate) struct FuelConfig {
-    /// The base fuel costs for all instructions.
-    pub base: u64,
-    /// The fuel cost for instruction operating on Wasm entities.
-    ///
-    /// # Note
-    ///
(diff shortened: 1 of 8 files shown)


soroban-env-host/src/budget/wasmi_helper.rs
```diff
@@ -141,3 +93,41 @@ impl ResourceLimiter for Host {
...
+        .wasm_mutable_global(true)
+        .wasm_saturating_float_to_int(false)
+        .wasm_sign_extension(true)
+        .floats(false)
+        .consume_fuel(true)
+        .fuel_consumption_mode(FuelConsumptionMode::Eager)
+        .set_fuel_costs(fuel_costs);
+
+    if protocol > 20 {
+        config.wasm_reference_types(false);
```
dmkozh, line 129: Is it possible to accidentally get this enabled in the build?

jayz22: what do you mean by that? This is a runtime config passed into vm instantiation

dmkozh: Right, but does it invalidate certain Wasms? E.g. `floats` option invalidates Wasms that do floating point operations, right?

jayz22: The main purpose of the reference type proposal is to enable object passing between the host and the guest via a new reference type (`ExternRef`). It also expands the table operations and loosens some requirements such as allowing multiple tables. 
Our host design is not taking advantage of this feature by any means, and we are explicitly limiting (via the resource limiter) the number of tables to 1. So any legit contract built from the SDK should never touch this feature. The only way it can show up is if someone crafts it manually, which we want to disable them to reduce our maintenance surface area.

graydon (approved):

PR #1293, opened 2023-12-06
Improve cost model fitting for better lower-range accuracy

jayz22:
### What

This PR was originated from the fact that previously the `VarDeser` cost's constant term was unrealistically high.  

It improves the numerical approach towards model generation by constraining it to the smallest input in the range. 
Thus it avoids fitting a model that tries to produce a marginally better fitness at the expense of lower-range accuracy. 

It also removes the dependency of the `HostCostMeasurement`'s input size on the linear parameter's scale factor (`COST_MODEL_LIN_TERM_SCALE_BITS`), which previously put a numerical constrain on the smallest input size and minimal gaps.

[raw calibration outputs](https://github.com/stellar/rs-soroban-env/files/13621533/output_x86_dec_8.txt)

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/benches/common/modelfit.rs
+++ b/soroban-env-host/benches/common/modelfit.rs
@@ -2,7 +2,10 @@ use std::collections::HashSet;
 use std::str::FromStr;
 
 use linregress::{FormulaRegressionBuilder, RegressionDataBuilder};
+use na::DVector;
 use num_traits::Pow;
+
+use nalgebra::{self as na, OMatrix, OVector, U2, DMatrix};
 
 #[derive(Debug, Default, Clone, PartialEq, PartialOrd)]
 pub struct FPCostModel {
@@ -79,14 +82,29 @@ pub fn fit_model(x: Vec<u64>, y: Vec<u64>) -> FPCostModel {
         };
     }
 
+    let nrows = x.len();
     let x = x.iter().map(|i| *i as f64).collect::<Vec<_>>();
     let y = y.iter().map(|i| *i as f64).collect::<Vec<_>>();
-    let mut res = fit_linear_regression(x.clone(), y.clone());
-    assert!(res.lin_param > 0f64, "negative slope detected, examine your data, or choose a constant model");
-    if res.const_param < 0f64 {
-        println!("negative intercept detected, will constrain it to 0 and recompte the r-squared");
-        res.const_param = 0f64;
-        res.r_squared = compute_rsquared(x, y, 0f64, res.lin_param);
-    }
-    res
+
+    let mut a = x.clone();
+    a.append(&mut vec![1.0; nrows]);
+    // println!("{}, {}", a.len(), y.len());
+    let a = OMatrix::<f64, na::Dyn, U2>::from_column_slice(&a);
+    let b = OVector::<f64, na::Dyn>::from_row_slice(&y);
+    // println!("{}, {}", a.len(), b.len());
+    let res = lstsq::lstsq(&a, &b, 1e-14).unwrap();
+    let const_param = res.solution[1];
+    let lin_param = res.solution[0];
+    let r_squared = compute_rsquared(x.clone(), y.clone(), const_param, lin_param);
+    FPCostModel{ const_param, lin_param, r_squared }
+
+
+    // let mut res = fit_linear_regression(x.clone(), y.clone());
+    // assert!(res.lin_param > 0f64, "negative slope detected, examine your data, or choose a constant model");
+    // if res.const_param < 0f64 {
+    //     println!("negative intercept detected, will constrain it to 0 and recompte the r-squared");
+    //     res.const_param = 0f64;
+    //     res.r_squared = compute_rsquared(x, y, 0f64, res.lin_param);
+    // }
+    // res
 }
--- a/soroban-env-host/benches/common/modelfit.rs
+++ b/soroban-env-host/benches/common/modelfit.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 use std::str::FromStr;
 
 use linregress::{FormulaRegressionBuilder, RegressionDataBuilder};
+use num_traits::Pow;
 
 #[derive(Debug, Default, Clone, PartialEq, PartialOrd)]
 pub struct FPCostModel {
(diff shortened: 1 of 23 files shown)


soroban-env-host/src/budget/model.rs
```diff
@@ -35,7 +35,7 @@ pub trait HostCostModel {
 /// been scaled by this factor during parameter fitting to retain more significant
 /// digits. Thus to get the cost from the raw input, we need to scale the result
 /// back by the same factor.
-pub const COST_MODEL_LIN_TERM_SCALE_BITS: u32 = 7;
+pub const COST_MODEL_LIN_TERM_SCALE_BITS: u32 = 0;
```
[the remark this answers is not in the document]
jayz22: This change has been reverted.

soroban-env-host/src/budget.rs
```diff
@@ -584,6 +584,67 @@ impl Display for BudgetImpl {
     }
 }
 
+#[allow(unused)]
+#[cfg(test)]
+impl BudgetImpl {
+    fn retrieve_unscaled_param(param: ScaledU64, is_protocol20: bool) -> u64 {
+        const PROTOCOL_20_LIN_TERM_SCALE_BITS: u32 = 7;
+        match is_protocol20 {
+            true => param.unscale() << PROTOCOL_20_LIN_TERM_SCALE_BITS,
```
dmkozh, line 593: Why do we need this check?

soroban-env-host/src/budget.rs
```diff
@@ -294,79 +294,79 @@ impl Default for BudgetImpl {
                     cpu.lin_term = ScaledU64::from_unscaled_u64(1).safe_div(8);
                 }
                 ContractCostType::DispatchHostFunction => {
-                    cpu.const_term = 310;
+                    cpu.const_term = 295;
```
dmkozh, line 297: While this seems _relatively_ safe, I'm not sure if changing the hardcoded values is the way forward. Probably env should just settle on some calibration and then SDK could be updated to match the network settings.

soroban-env-host/src/budget.rs
jayz22: It's not needed now since we didn't change the number of bits. But just wanted to be extra sure we are scaling to the correct bits, in case we did decide to change it (now or in the future).

dmkozh: But is there an practical need to do that? Even then, we do have extensions, so it would be more reasonable to put the scaling bits there (though I'm not sure why would we ever want to do that; 128 seems good enough even for fast operations).

jayz22: Removed it. Realistically we will rarely need to print a current set of parameters in an older protocol. So I removed it for less confussion. The whole things is gated by `test` anyway .

dmkozh: Yeah, the test gating was actually confusing; ideally tests wouldn't need to deal with protocol versions much.

FWIW the parameter set should normally be independent of the protocol. We might introduce the new cost parameters, but otherwise it doesn't really matter from the host test perspective what the exact costs are. So I would try to limit the impact of the calibration changes on the host (see comment below). Maybe it's worth to have just a sanity check test that exercises all the newly calibrated costs and dumps the results (we do have one such test in hostile.rs, according to diffs), and otherwise have the calibration constant.

soroban-env-host/benches/common/modelfit.rs
```diff
@@ -1,13 +1,24 @@
...
+    lin_param: f64,
+    r_squared: f64,
+}
+
+impl From<FPCostModel> for MeteredCostComponent {
+    fn from(model: FPCostModel) -> Self {
+        let (const_term, unscaled_lin_term) = model.params_as_u64();
+        MeteredCostComponent {
+            const_term,
+            lin_term: ScaledU64::from_unscaled_u64(unscaled_lin_term),
```
graydon, line 19: Ah, this is really nice!

graydon, line 19: Wait, I think this is still maybe not quite right -- you're still losing all the fractional precision in this conversion.

To preserve the fractional precision, you want to scale the f64 as an f64 first, then convert to integer:

```
let fp_scale_factor: f64 = (1 << COST_MODEL_LIN_TERM_SCALE_BITS) as f64;
let fp_scaled_lin_param: f64 = model.lin_param * fp_scale_factor;
lin_term: ScaledU64 = ScaledU64(fp_scaled_lin_param as u64);
```

You'll probably want to encapsulate that logic inside `ScaledU64` though; perhaps a clean way would be to `impl From<f64> for ScaledU64`, then you could have `lin_term: model.lin_param.into()` in this code.

jayz22: Ah yes, you are right.

jayz22: Fixed it, thanks for the catch!
I want to point out (in case you missed it) that there was some rounding logic that had to be threaded in, besides what you suggested.

graydon (approved):
Looks good, thanks!

soroban-env-host/src/budget/model.rs
```diff
@@ -72,16 +72,18 @@ impl Display for ScaledU64 {
     }
 }
 
-impl Debug for ScaledU64 {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Scaled({})", self.0)
+impl From<f64> for ScaledU64 {
```
dmkozh, line 75: Shall we guard this with benchmark feature?

graydon: I think it's a fairly harmless / intrinsic aspect of the type. The type represents a fixed-point fraction, it's reasonable to equip it with a (lossy but exactly-as-you-would-expect) conversion from floating-point. Not that anyone's going to _use_ that outside benchmarking, but it seems harmless to me.

dmkozh: We just avoided floats in host code altogether, but I don't have a strong opinion on this either.

jayz22: I've put it behind the feature gate. 
I think it is good to have, just to be extra clear floating points aren't enabled in the regular code flow.

PR #1300, opened 2023-12-14
Migrate preflight computations from soroban-rpc

2opremio:
### What

Migrate preflight computations from https://github.com/stellar/soroban-tools

### Why

https://github.com/stellar/soroban-tools/issues/1098

### Known limitations
N/A


soroban-env-host/Cargo.toml
```diff
@@ -12,6 +12,10 @@ rust-version = "1.74"
 build = "build.rs"
 
 [dependencies]
+# To be removed
+anyhow = { version = "1.0.75", features = [] }
+thiserror = "1.0.46"
```
2opremio, line 17: I understand this needs to go, I want to first get a general approval on the structure and code.

soroban-env-host/src/preflight.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub memory_bytes: u64,
+    pub restore_preamble: Option<RestorePreamble>,
+}
+
+pub fn preflight_invoke_hf_op(
+    ledger_storage: LedgerStorage,
+    bucket_list_size: u64,
+    invoke_hf_op: InvokeHostFunctionOp,
+    source_account: AccountId,
+    ledger_info: LedgerInfo,
```
dmkozh, line 45: I'm not certain what exactly do we achieve with this change (besides changing code ownership, I guess?). If that's just the first step before more refactoring, that's fine, but I think even in this scenario I'm not sure if all the logic here belongs to the host, like `LedgerStorage` wrapper.
Basically, on a high level I see the following components:
- Invoking host fn e2e in recording storage mode. This is similar to what `e2e_invoke` already does and it would be nice to share as much code as possible with it. The interface should probably be more abstract (e.g. use `SnapshotSource` instead of `LedgerStorage`). Budget probably should be set up externally. Restoration preamble also probably doesn't belong here. Basically, the closer we can make this to `e2e_invoke::invoke_host_function`, the better.
- Then we could have a module that computes the resource requirements and fee given the result of e2e invocation
- Restore/extend ops preflight really have nothing to do with the host. I don't have a strong opinion as to where they should belong, the main point is that they belong to a separate module.
- XDR config settings are normally abstracted away from host as well. Again, no matter the final location, these probably should belong to a separate module
- The same goes for transaction XDR manipulation - probably should be a separate module
- The 'glue' layer that puts everything together belongs to soroban-rpc. I can't quickly tell what belongs there, but it seems to me that more bespoke stuff like archived entry set for restoration should be there as well.
Basically, we want to encapsulate the host-related logic, so that it's possible to reasonably test it and compare with our e2e_invoke results. Everything else should be separated and covered separately. Again, we don't have to do all the refactoring immediately, but maybe it's worth to lay some foundations, at least in terms of the interface.

2opremio: In general, I agree with your comments. The initial state of this PR mostly had the intention of starting a conversation.

Some remarks/questions:
* I agree that the restore/extend fee computation is mostly unrelated to the host. However, the code should live in a place:
  * Which makes it easy to update whenever Core's metering is updated
  * Where can be automatically tested against Core's metering to make sure they don't diverge.
  * It can share code with the invoke host function fee computation (this is less important).
  Any suggestions?
  
* Maybe we don't need `LedgerStorage` in its current state, but `SnapshotSource`
  1. doesn't allow querying non-live entries which is needed for:
      * the restore operation preflight
      * tracking ledger entries which require restoring. This could arguably stay in soroban-rpc, though
  2. The `Rc`s make it awkward to use
 * Do you have any suggestions as to how to split the code in separate modules other than host? I don't have any strong opinions and I am not familiar with the existing env modules, so it would make the migration smoother is I was guided with respect to that.

dmkozh: > Which makes it easy to update whenever Core's metering is updated

I don't think there is a good solution for now, just because core code lives in C++ and this lives in Rust. Moving these to Rust is probably infeasible - the fee computation is already on the Rust side, but doing an XDR roundtrip just to compute the entry size seems inefficient. That said, the Core metering update is a protocol change, so it would happen very rarely (if ever). So while I agree it would be nice to share even more code, the changes are unlikely enough now to not worry about that too much. The same goes for comparison with Core - while it would be nice to have an automated way to test this, it seems like *a lot* of setup and maintenance work, all for very little benefit. I think a compromise solution would be to implement unit tests with the same setup in Core and in the Rust library and leave the reference in Core stating that Rust test has to be in sync with it.

 To be clear, I'm not opposed to storing this in `rs-soroban-env`, but it would probably be cleaner if we had a separate crate for non-host-related stuff like this. I think something like `soroban-preflight` or `soroban-simulation`  would be appropriate if we also move more preflight-specific utils there; otherwise something like `soroban-protocol-utils` would work as well.

> Maybe we don't need LedgerStorage in its current state, but SnapshotSource

To be clear, what I'm suggesting is to just take `SnapshotSource` trait in e2e invocation function; it seems like we need only `SnapshotSource` functions there (besides footprint restoration preamble, but as I've mentioned, I think it should happen outside of this function). `LedgerStorage` itself may live either in `soroban-preflight` in env, or in `soroban-rpc`. This will help with setting up the comparison tests between the 'canonical' and preflight host fn invocations.

> Do you have any suggestions as to how to split the code in separate modules other than host?

As I've suggested above, I would probably move most of the new code into a separate crate. You can then make a pass to maybe separate this into a bit more granular modules (I had some ideas above), but really for the first iteration we don't need to spend too much time on this - we can refactor this further later. The end goal would be to have a reasonably small public interface and a good unit test set that covers it, as well as some tests for internal modules where appropriate.

dmkozh: Just a small clarification: I've mentioned 'separate workspace' above, but in fact I meant just a separate workspace member (and respective separate crate) in rs-soroban-env workspace.

2opremio: @dmkozh PTAL, I moved everything to a another workspace member (`soroban-preflight`).

I think it's a good starting point.

2opremio: @dmkozh if you are OK with the structure, I can move on to remove the anyhow/thiserror dependencies.

dmkozh: There is some portion of this that we'll need to move to env, but it's probably ok to do as the next step. I'll need to spend some time refactoring this to reuse the env code more, and maybe move some bits back to the soroban-rpc repo if they seem too specific. But this looks fine for now.

2opremio: OK, great! I will tweak the PR a bit further and let you know once it’s ready to be merged.

soroban-preflight/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "soroban-preflight"
```
leighmcculloch, line 2: Can we rename the crate to `soroban-simulation`? Preflight is a specific use case. Simulation is a broad capability that we'll likely use in more places that preflight. I believe we also discussed the vocab we'd use for this and settled on simulation, which is also what we use in APIs such as soroban-rpc.

leighmcculloch: @2opremio I've reserved the `soroban-simulation` crate on crates.io for us to use: https://crates.io/crates/soroban-simulation

soroban-preflight/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+description = "Soroban transaction preflight computations."
```
leighmcculloch, line 3: Suggest changed description to focus on the capability and less on one of the use cases we're using it in:
```suggestion
description = "Soroban host invocation simulations."
```

soroban-preflight/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+homepage = "https://github.com/stellar/rs-soroban-env"
+repository = "https://github.com/stellar/rs-soroban-env"
+authors = ["Stellar Development Foundation <info@stellar.org>"]
+license = "Apache-2.0"
+version.workspace = true
+readme = "../README.md"
+edition = "2021"
+rust-version = "1.74"
```
leighmcculloch, line 11:
```suggestion
rust-version.workspace = true
```

soroban-preflight/Cargo.toml
2opremio: > Can we rename the crate to `soroban-simulation`? Preflight is a specific use case. Simulation is a broad capability that we'll likely use in more places that preflight. I believe we also discussed the vocab we'd use for this and settled on simulation, which is also what we use in APIs such as soroban-rpc.

Done!

soroban-preflight/Cargo.toml
2opremio: I cannot do this right now since there is no top-level version. I think that belongs to a separate PR

leighmcculloch: There's a top level version in `main` now. Go for it 😄

soroban-env-host/Cargo.toml
2opremio: I tried removing `anyhow` and there wasn't a clean way to provide the same context to errors without replicating much of the machinery anyhow provides. So I am going to leave it for now.

2opremio: What I have done though, is hiding anyhow externally.

2opremio:
@dmkozh PTAL

soroban-simulation/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "soroban-preflight"
```
dmkozh, line 2: soroban-simulation?

soroban-simulation/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+use std::iter::FromIterator;
+use std::rc::Rc;
+
+pub struct RestorePreamble {
+    pub transaction_data: SorobanTransactionData,
+    pub min_fee: i64,
+}
+
+#[derive(Default)]
+pub struct PreflightResult {
```
dmkozh, line 28: With the crate rename it would probably make sense to make a find-and-replace pass to change 'Preflight'  into `Simulation/Simulate` depending on context.

soroban-simulation/src/ledger_storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    NulError(#[from] NulError),
+    #[error("utf8 error: {0}")]
+    Utf8Error(#[from] Utf8Error),
+    #[error("unexpected config ledger entry for setting_id {setting_id}")]
+    UnexpectedConfigLedgerEntry { setting_id: String },
+    #[error("unexpected ledger entry type ({ledger_entry_type}) for ttl ledger key")]
+    UnexpectedLedgerEntryTypeForTtlKey { ledger_entry_type: String },
+}
+
+impl From<Error> for HostError {
```
dmkozh, line 34: This conversion looks suspicious, why do you need to build the host errors somewhere outside of the host?

soroban-simulation/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        &storage,
+        &budget,
+        &diagnostic_events,
+        &result,
+        bucket_list_size,
+        ledger_info.sequence_number,
+    )
+    .context("cannot compute resources and fees")?;
+
+    let restore_preamble = compute_restore_preamble(
```
dmkozh, line 130: I would still prefer to decouple the restore preamble logic from this function in this PR and switch to using the `SnapshotSource` trait instead of `LedgerStorage`. I would do that sooner or later, but if we do that now, we'll be able to preserve the interface of this function. OTOH I might need to change the interface more, so maybe it doesn't matter in the end. Up to you.

dmkozh (commented):
Just a reminder, if you have any unit tests for any components here, I would really appreciate moving them here as well.

soroban-simulation/src/lib.rs
2opremio: Let's do it later if you are OK with it.

soroban-simulation/src/lib.rs
2opremio: Done

soroban-simulation/src/ledger_storage.rs
2opremio: It's needed for `impl SnapshotSource for LedgerStorage`

2opremio:
> Just a reminder, if you have any unit tests for any components here, I would really appreciate moving them here as well.

Unfortunately, there aren't

soroban-simulation/src/ledger_storage.rs
dmkozh: I see, makes sense.

dmkozh (approved):
I think this is fine as a starting point. I'll work on moving the relevant parts of this to the rs-soroban-env. As I've already mentioned this will most likely have some interface changes, so we'll need a soroban-rpc followup eventually, but hopefully it won't be more complex than a few lines.

PR #1302, opened 2023-12-18
Use strkeys for contract IDs and addresses in diagnostic events.

### LINKED ISSUE #1049: Addresses are outputted as hex instead of strkeys

eb5a9ba053a7b64a8eff605db625525378f7bea0

### What did you do?

Saw diagnostic logs from the Env:
```
   1: [Diagnostic Event] contract:60cbc66b12df4bf04b9e5f7cf9051b806deaad52bc9d623673313893039b2a6f, topics:[error, Error(Auth, InvalidAction)], data:["contract call failed", transfer, [Address(Contract(c132544e1bd49d95cba076edcfd9bb175dd55ab21a881d9b7f7bf497cffa8ed8)), Address(Contract(60cbc66b12df4bf04b9e5f7cf9051b806deaad52bc9d623673313893039b2a6f)), 150000000]]
   2: [Failed Diagnostic Event (not emitted)] contract:1e276ddd0b7568275cc561c86704f18f77e3238dc2443b7afff3bc2db63fad06, topics:[log], data:["VM call trapped with HostError", transfer, Error(Auth, InvalidAction)]

```

See https://discord.com/channels/897514728459468821/1149701029387055166/1149701029387055166

Addresses, including Contract IDs, outputted as strkeys ([SEP-23]).

It's confusing for developers if they see values in different formats.

[SEP-23]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md

### What did you see instead?

Addresses outputted as hex strings.

graydon:
What it says on the label.

Fixes #1049

### DIFF
--- a/soroban-env-host/src/events/mod.rs
+++ b/soroban-env-host/src/events/mod.rs
@@ -28,12 +28,12 @@ fn display_address(addr: &ScAddress, f: &mut std::fmt::Formatter<'_>) -> std::fm
         ScAddress::Account(acct) => match &acct.0 {
             PublicKeyTypeEd25519(e) => {
                 let strkey = stellar_strkey::ed25519::PublicKey(e.0.clone());
-                write!(f, "Address(Account({}))", strkey)
+                write!(f, "{}", strkey)
             }
         },
         ScAddress::Contract(hash) => {
             let strkey = stellar_strkey::Contract(hash.0.clone());
-            write!(f, "Address(Contract({}))", strkey)
+            write!(f, "{}", strkey)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


sisuresh (approved):

soroban-env-host/src/events/mod.rs
```diff
@@ -26,9 +26,15 @@ pub struct HostEvent {
 fn display_address(addr: &ScAddress, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
     match addr {
         ScAddress::Account(acct) => match &acct.0 {
-            PublicKeyTypeEd25519(e) => write!(f, "Address(Account({}))", e),
+            PublicKeyTypeEd25519(e) => {
+                let strkey = stellar_strkey::ed25519::PublicKey(e.0.clone());
+                write!(f, "Address(Account({}))", strkey)
```
leighmcculloch, line 31: What is this used for? Is the `Address(Account(` required? Asking because the `Address(Account(` seems redundant, that same information will be obvious in the strkey.

leighmcculloch (approved):

soroban-env-host/src/events/mod.rs
leighmcculloch: At the very least we can drop the `Account(` part I think.

dmkozh: Yeah, I suggest just display addresses as strkeys everywhere - there shouldn't be contexts where we actually want to communicate e.g. AccountId without the corresponding address.

dmkozh: To be clear, the same goes for the contract address.

graydon: Nothing is required, this is purely for display purposes; I put it there originally because they were uninformative 32-byte hex blobs. Maybe if strkeys are more-obvious to users we can remove it. I have no strong feelings.

Issue #779, opened 2023-04-27
Add host functions to support BLS12-381 and the required field arithmetics  


tomerweller:
Pairing friendly elliptic curves are required for advanced zero knowledge proof systems including privacy preserving constructions and zkrollups. The Soroban ecosystem has already expressed interest in these type of systems but can not experiment with them due to the lack of host side support for pairing friendly curves. 

The most widely used pairing friendly curve is BN254 (ake BN256 aka alt-bn128) due to the fact that ethereum has support for. However, BLS12-381 has been proven to be more secure and various projects in the zk space already use it or have signaled support for it and announced that they will migrate to it once available in Ethereum (See [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537)). There are other curves available like BLS12-377 but these appear to be less widely used. 

I propose that we add the relevant host functions to support BLS12-381 and the underlying fields arithmetics.  

Questions
1. does anyone disagree?
2. what is the exact set of required host functions?
3. what implementation should we use? 

cc @kwantam

kwantam:
A couple other things to note: BLS12-381 is the curve that's used by the Eth beacon chain, and there's at least one very high quality implementation ([blst](https://github.com/supranational/blst), which has been thoroughly vetted / formally verified by Galois (not 100% sure the verification conditions, but I joined some of the early calls and it seems like they were proving low-level safety properties and conformation with the spec).

I'd definintely implement all of the EIP-2537 functionality. I would also _consider_ implementing mappings from arbitrary strings to Fp and Fp2 ("hash to field" functions), which together with the BLS12_MAP_FP_TO_G1 and BLS12_MAP_FP2_TO_G2 functions from EIP-2537 give the full hash-to-curve functionality. If you go this way, the hash-to-field mappings should be the ones specified in the IETF hash-to-curve draft (which will soon be an RFC). Happy to give more details.

The only downside of using BLS12-381 instead of BN254 is that existing proof statements will need to be rewritten. Sometimes that's easy, but in general there's real porting effort involved. So, if you think it's important for existing proof systems already running on Eth to port directly over, it would make sense to also support BN254. But it's not obvious to me that's important...

tomerweller:
I'm comfortable with not having BN254 for now given that: 
1. BLS12-381 has been proven to be more secure 
2. Major ZK projects on ethereum have signaled support for BLS12-381 and indicated they will migrate once that's available
3. I don't expect in the short term to see EVM constructions migrate as is to Soroban
[...]

PR #1310, opened 2023-12-27
BL12-381 Integration

PlamenHristov:
### What

Added: 
- BLS12_G1ADD
- BLS12_G1MUL
- BLS12_G1MULTIEXP
- BLS12_G2ADD
- BLS12_G2MUL
- BLS12_G2MULTIEXP

Missing:

- BLS12_PAIRING
- BLS12_MAP_FP_TO_G1
- BLS12_MAP_FP2_TO_G2
- BLS12_HASH_TO_G1
- BLS12_HASH_TO_G2

### Why

Because of issue  #779

### Known limitations

- Missing tests
- Panics if points are invalid
- Implementation is likely to be imperfect


jayz22:
Some general comments. I'm not too familiar with the implementation details yet.

None of these functions can use existing metering cost types, for each of these functions {p1, p2}_{add, mul, serialize}, a new cost types need to be added, calibrated and charged.
(I also remember @graydon mentioned something we need to be careful about when it comes to FFI, but I can't recall exactly)
These functions need to be exposed via the env, via env.json entry and a corresponding implementation in host.rs.
It would be nice to have some tests, not comprehensive coverage but for sanity checks, to make sure they are doing what's expected and there are no obvious issues.

PlamenHristov:
1. Implemented the missing:
- BLS12_PAIRING
- BLS12_MAP_FP_TO_G1
- BLS12_MAP_FP2_TO_G2
- BLS12_HASH_TO_G1
- BLS12_HASH_TO_G2

2.  Added tests 

Ready for review in terms of the implementation. 


Still requires adding budgeting of individual operations and charging them in [stellar-xdr](https://github.com/stellar/rs-stellar-xdr)

PR #1315, opened 2024-01-08
Enable VM execution in a WASM environment by guarding time track behind `time` feature

heytdep:
### What

This PR guards the usage of `Instant` behind a `time` feature, at least for a VM invocation. 

### Why

This enables the soroban vm to execute within a WASM context. For instance, without this modification running the soroban VM in wrangler resulted in

```
✘ [ERROR] panicked at null.<anonymous> (library/std/src/sys/wasm/../unsupported/time.rs:13:9):

  time not implemented on this platform
  
  Stack:
  
  Error
      at nt
  (file:///mnt/storagehdd/sdf-work/soroflare/soroflare-wrangler/build/worker/shim.mjs:2:4496)
      at [object Object]x19354a
      at [object Object]x163c0d
      at [object Object]x1730b8
      at [object Object]x2e74d
      at [object Object]x3f76d
      at [object Object]x1dc90
      at [object Object]x8de53
      at [object Object]x2cee6
      at [object Object]x129a8d


✘ [ERROR] Uncaught (in response) Error: The script will never generate a response.
```

Discord reference: https://discord.com/channels/897514728459468821/1187049026831523940/1187457822930260018

### Known limitations

N/A

### Implementation details

I'm using a closure here but a named fn is probably better.


soroban-env-host/src/vm.rs
```diff
@@ -220,87 +220,98 @@ impl Vm {
...
+            host.as_budget().track_time(
+                ContractCostType::VmInstantiation,
+                now.elapsed().as_nanos() as u64,
+            )?;
+
+            Ok(vm)
         } else {
-            None
-        };
+            let vm = instantiate()?;
```
dmkozh, line 311: nit: This can be replaced with just `instantiate()` .

soroban-env-host/src/vm.rs
```diff
@@ -220,87 +220,98 @@ impl Vm {
...
-                .ensure_no_start(&mut store)
-                .map_err(|ie| wasmi::Error::Instantiation(ie)),
-        )?;
+        if cfg!(feature = "time") {
+            let now = Instant::now();
+            let vm = instantiate()?;
 
-        let memory = if let Some(ext) = instance.get_export(&mut store, "memory") {
-            ext.into_memory()
+            host.as_budget().track_time(
```
dmkozh, line 304: Probably it would make sense to guard `track_time` with the same feature (it's meant to only be used in conjunction with real time).

dmkozh (commented):
Thanks, this looks good overall, just a couple small things.

jayz22:
Thanks for getting started with the fix. I think for now it is okay to proceed with the minimally disruptive fix to unblock your use case. So I wouldn't worry too much about things like returning different errors on feature on vs off, or gating the budget behind it. 
The time measurement is a one-off case to get better metrics for Vm instantiation (for all other types we currently return 0). It will likely go through more changes in the next protocol release, when we improve Vm instantiation cost model. We might either expand it to include other cost types (which will require a more holistic design), or remove it all together. 

> I'm using a closure here but a named fn is probably better.

I prefer a named function for maintainability.

soroban-env-host/Cargo.toml
```diff
@@ -79,6 +79,7 @@ next = ["soroban-env-common/next"]
 tracy = ["dep:tracy-client", "soroban-env-common/tracy"]
 recording_auth = []
 bench = []
+time = []
```
leighmcculloch, line 82: I think we could gate the tracking of time functionality on the target instead of a feature. I would favor this for a couple reasons:
- Every new feature we add increases build times and testing. We don't need to avoid adding features. But we should be intentional about adding them.
- Gating the functionality on the wasm32 family would more clearly communicate what use case the code is gated out of. I doubt there will be other situations we need to disable time.

soroban-env-host/src/vm.rs
```diff
@@ -288,19 +268,53 @@ impl Vm {
...
+    pub fn new(
+        host: &Host,
+        contract_id: Hash,
+        module_wasm_code: &[u8],
+    ) -> Result<Rc<Self>, HostError> {
+        let _span = tracy_span!("Vm::new");
+
+        if cfg!(feature = "time") {
+            let now = Instant::now();
+            let vm = Self::instantiate(host, contract_id, module_wasm_code)?;
```
leighmcculloch, line 307: I think we should add a CI run that builds for wasm32 and executes the environment inside a wasm vm to verify that this capability is maintained. We used to have this capability of running the env inside a vm (I did it for soroban-fiddle) and clearly we broke it at some point without knowing and we should have a test that tells us if we break it again, or tells us where we need more gating like this.

leighmcculloch (changes requested):
Oh thank for identifying this problem. We used to support using the environment in a vm and that's still a goal. Couple asks inline.

soroban-env-host/Cargo.toml
dmkozh: That's actually a great idea, thanks for the suggestion. I agree we should the target gate instead.

jayz22: Yes agreed this is better!

heytdep: Yeah definitely makes sense

heytdep: @leighmcculloch changed this in the latest commit.

soroban-env-host/src/vm.rs
```diff
@@ -196,32 +196,12 @@ impl Vm {
...
-    ///   - Instantiates the module, leaving it ready to accept function
-    ///     invocations.
-    ///   - Looks up and caches its linear memory export named `memory`
-    ///     if it exists.
-    ///
-    /// This method is called automatically as part of [Host::invoke_function]
-    /// and does not usually need to be called from outside the crate.
-    pub fn new(
+    /// Instantiates a VM given the arguments provided in [`Self::new`]
+    pub fn instantiate(
```
dmkozh, line 200: This doesn't need to be public.

heytdep: yep definitely, updated it now

jayz22 (approved):

dmkozh (approved):

PR #1336, opened 2024-01-30
Bug 1283 asset code rendering

### LINKED ISSUE #1283: Asset codes are stored in strings with trailing zeros

4aee9508153688ef984628a43dba7d9e276a0cd1

### What did you do?

- Used the stellar asset contract.
- Looked at the stored ledger entries.

- Asset code's stored in strings as "ABC".

### What did you see instead?

- Asset code's stored in strings as "ABC\0".

### Discussion

The XDR stores asset codes in fixed length arrays of 4 and 12 bytes long, but it's unnecessary for those asset codes to be stored with those trailing bytes when stored as ScVal Strings which are variable length.

Applications will use ledger entries directly, and this pushes the need to know about the oddity of asset codes in their XDR form to downstream systems.

This also causes asset codes to render with those codes in other places exposed to developers, like in JSON for stellar asset contract ledger entries.

Note that asset codes when endered to string/JSON from their AssetCode/AssetCode4/AssetCode12 XDR types will render without the trailing zero bytes as a result of the following change, but that change doesn't impact the asset code strings that the soroban-env writes to ledger entries:
- https://github.com/stellar/rs-stellar-xdr/pull/324

Note that this was a problem in the soroban-env in the past for where it exposed asset codes to contracts. It seems like we fixed the problem in that area, but it is still leaking the trailing characters in the area of ledger entries:
- https://github.com/stellar/rs-soroban-env/issues/1032

graydon:
Fixes #1283 by reusing the rendered asset code prefix of `metadata.name` for `metadata.symbol`

Unfortunately creates a new host object during SAC init, which requires re-observing basically all the SAC tests.


graydon:
(Note the other option -- trimming `asset_info.asset_code` and re-padding it when converting to XDR -- would probably cause a little less churn in the testsuite due to not having to make a new object; but it would still perturb a bit due to the hash of storage and the host object repertoire changing. Not sure if it matters, but I'm willing to try that variant too?)

leighmcculloch (approved):

Issue #1297, opened 2023-12-14
Fix auth entry charge


jayz22:
https://github.com/stellar/rs-soroban-env/blob/9c5394083725bacf32a56107d7fa2adf643dc7b5/soroban-env-host/src/auth.rs#L855-L856

```diff
- Vec::<Val>::charge_bulk_init_cpy(auth_entries.len() as u64, host)?;
+ Vec::<InvokerContractAuthorizationTracker>::charge_bulk_init_cpy(auth_entries.len() as u64, host)?;
```

jayz22:
Using this to track a couple of other issue surfaced during the same review:

> On line 1018 in auth.rs in function snapshot you seem to allocate a vector to store authorization snapshots but then charge for the allocation afterwards. Can you confirm that the charge should happen before the allocation?

and 

> On lines 694-700 in auth.rs in the function new_enforcing (for AuthorizationManager) I noticed that you allocate a vector of AccountAuthorizationTrackers of size num_entries. However, the vector you allocate is of type Vec<RefCell<AccountAuthorizationTracker>> so the amount of memory allocated  is actually num_entries*(sizeof(RefCell<AccountAuthorizationTracker>)). The difference in memory allocated is n*(sizeof(RefCell<AccountAuthorizationTracker>) - sizeof(AccountAuthorizationTracker)).  Why not try and account for the difference when charging?

jayz22:
resolved by https://github.com/stellar/rs-soroban-env/pull/1340

PR #1340, opened 2024-01-31
Tightening up metering in auth

jayz22:
### What

Fix a few small but protocol-breaking metering issues discovered by the security review:
i. auth entry charging the wrong amount https://github.com/stellar/rs-soroban-env/issues/1297
ii. (not protocol breaking) auth `snapshot()` function charge after allocation
iii. auth manager not charging `RefCell` wrapping. 

I also did a clean up by wrapping the `with_capacity` method with metering into a new `MeteredContainer::with_metered_capacity` method, which makes sure bugs like (i) and (ii) above cannot happen. 

Unfortunately these breaks a lot of existing observations that depends on auth, basically all `mem_bytes` have increased by 8 (size of the RefCell counter) and `cpu_insns` have increased by 2, which is caused by fixing (iii).

### Why

### Known limitations

[TODO or N/A]


graydon (approved):
LGTM, thanks!

PR #1344, opened 2024-02-05
Enable publish of soroban-simulation crate

leighmcculloch:
### What
Enable publish of soroban-simulation crate

### Why
It is depended on by the soroban-rpc and should have been included in the v20.2.0 release, but appears we forgot to set it as publishable.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -72,7 +72,7 @@ jobs:
       with:
         name: cargo-semver-checks
         version: 0.27.0
-    - run: cargo semver-checks
+    - run: cargo semver-checks --exclude soroban-simulation
 
   build-and-test:
     strategy:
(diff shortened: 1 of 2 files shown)


graydon (approved):

anupsdf (approved):
LGTM! Thanks for taking care of this.

.github/workflows/rust.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       with:
         name: cargo-semver-checks
         version: 0.27.0
-    - run: cargo semver-checks
+    - run: cargo semver-checks --exclude soroban-simulation
```
leighmcculloch, line 75: The simulation crate isn't published yet and the check fails trying to pull it from crates.io. Excluding it is required because semver-checks requires the crate to be published first. See:
- https://github.com/obi1kenobi/cargo-semver-checks/issues/481

PR #1348, opened 2024-02-14
Host function protocol gating

jayz22:
### What

Allow host functions to specify its supported protocol bounds, e.g. 

```diff
        {
            "name": "test",
            "export": "t",
            "functions": [
                {
                    "export": "_",
                    "name": "dummy0",
                    "args": [],
                    "return": "Val",
                    "docs": "A dummy function taking 0 arguments and performs no-op.                
+                  "min_supported_protocol": 20,
+                   "max_supported_protocol": 21
                }                
            ]
        },
```

And adds logic (mostly macro generated) that automatically checks and enforces these protocol bounds in various paths: 
- Host running as native. Covered by `impl Env for VmCallerEnv`.
- Host running from the guest Wasm contract. Covered by Vm link-time check, as well as the dispatch functions (as a redundant safe guard). 

Tests are added in `tests/protocol_gate.rs`.

### Why

### Known limitations

[TODO or N/A]


soroban-env-host/src/test/hostile.rs
```diff
@@ -773,22 +772,13 @@ fn test_corrupt_custom_section() -> Result<(), HostError> {
...
-
-    let interface_meta = |proto: u32, pre: u32| -> Vec<u8> {
-        let iv = (proto as u64) << 32 | pre as u64;
-        let entry = ScEnvMetaEntry::ScEnvMetaKindInterfaceVersion(iv);
-        let bytes = Vec::<u8>::new();
-        let mut w = Limited::new(bytes, Limits::none());
-        entry.write_xdr(&mut w).unwrap();
-        w.inner
-    };
+    let ledger_pre = meta::get_pre_release_version(meta::INTERFACE_VERSION);
```
graydon, line 775: This is interesting -- presumably this is why the test changed, was it a bug?

graydon (approved):
I pushed a couple minor fixes, looks good overall!

soroban-env-host/src/test/hostile.rs
jayz22: Yes, this was a bug in the test. We were passing the protocol version number as the prerelease number. The test passed because it still satisfied the condition: input mismatched with the expected prerelease number. 
Yes, this is why the test changed.

graydon (approved):

soroban-env-common/env.json
```diff
@@ -2042,7 +2042,16 @@
...
                     "docs": "A dummy function taking 0 arguments and performs no-op. This function is for test purpose only, for measuring the roundtrip cost of invoking a host function, i.e. host->Vm->host."
-                }
+                },
+                {
+                    "export": "0",
+                    "name": "protocol_gated_dummy",
+                    "args": [],
+                    "return": "Val",
+                    "docs": "A dummy function for testing the protocol gating. Takes 0 arguments and performs no-op",
+                    "min_supported_protocol": 19,
```
dmkozh, line 2052: Isn't this a protocol change? I thought it was only intended for testing. I think it should still be gated by protocol 21, or otherwise we would risk a divergence.

jayz22: This won't have an effect on protocol because no contract can link to it. It's min and max protocol are both specified as 19. So if a contract tries, it will fail the validation check during the upload (same as linking to a non-existent function today).

The reason this has to be 19 (or any value lower than 20) is because the test includes the case where the ledger protocol version > host function max protocol (see `test_host_protocol_gating_for_wasm`), and the ledger protocol version cannot be larger than the env version (20, which is compiled in and non-overridable).

Note setting the min_supported_protocol to 21 is using the exact same mechanism for protection against contract built with protocol 20. So effectively this function is always protocol gated out.

dmkozh: Ah, that's clever. Thanks for the explanation.

jayz22: Good point. I've added a comment to the function docs in https://github.com/stellar/rs-soroban-env/pull/1356

PR #1350, opened 2024-02-18
bump proc-macro2

salaheldinsoliman:
### What
1- Bump proc-macro2 crate in soroban-builtin-sdk-macros and soroban-env-macros
2- add Cargo.lock to .gitignore

### Why
A version conflict is happening while building solang, after adding soroban-env-host as a dependancy:
`error:` failed to select a version for `proc-macro2`.
    ... required by package `soroban-builtin-sdk-macros v20.2.2`
    ... which satisfies dependency `soroban-builtin-sdk-macros = "=20.2.2"` of package `soroban-env-host v20.2.2`
    ... which satisfies dependency `soroban-env-host = "^20.2.2"` of package `solang v0.3.3`
versions that meet the requirements `=1.0.69` are: 1.0.69

all possible versions conflict with previously selected packages.

  previously selected package `proc-macro2 v1.0.78`
    ... which satisfies dependency `proc-macro2 = "^1.0.78"` of package `parse-display-derive v0.9.0`
    ... which satisfies dependency `parse-display-derive = "=0.9.0"` of package `parse-display v0.9.0`
    ... which satisfies dependency `parse-display = "^0.9"` of package `solang v0.3.3`

failed to select a version for `proc-macro2` which could resolve this `conflict`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1333,7 +1333,7 @@ checksum = "4dccd0940a2dcdf68d092b8cbab7dc0ad8fa938bf95787e1b916b0e3d0e8e970"
 
 [[package]]
 name = "soroban-bench-utils"
-version = "20.2.1"
+version = "20.2.2"
 dependencies = [
  "perf-event",
  "soroban-env-common",
@@ -1342,7 +1342,7 @@ dependencies = [
 
 [[package]]
 name = "soroban-builtin-sdk-macros"
-version = "20.2.1"
+version = "20.2.2"
 dependencies = [
  "itertools",
  "proc-macro2",
@@ -1352,7 +1352,7 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-common"
-version = "20.2.1"
+version = "20.2.2"
 dependencies = [
  "arbitrary",
  "crate-git-revision",
@@ -1370,7 +1370,7 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-guest"
-version = "20.2.1"
+version = "20.2.2"
 dependencies = [
  "soroban-env-common",
  "static_assertions",
@@ -1378,7 +1378,7 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-host"
-version = "20.2.1"
+version = "20.2.2"
 dependencies = [
  "arbitrary",
  "backtrace",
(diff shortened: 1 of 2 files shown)


soroban-builtin-sdk-macros/Cargo.toml
```diff
@@ -15,7 +15,7 @@ proc-macro = true
 [dependencies]
 syn = {version="=2.0.39",features=["full"]}
 quote = "=1.0.33"
-proc-macro2 = "=1.0.69"
+proc-macro2 = "1.0.78"
```
leighmcculloch, line 18: It would be enough to remove the = while keeping the number the number the same. Changing the number changes the min version required which I don't think we could do as it would force stellar-core to use a newer version of the crate.

leighmcculloch (commented):
One request inline. 

@graydon I think this is a good example of how pinning deps is going to play out. This won't be the last time that or pinning creates a conflict for embedders.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
 .vscode
 /vendor
-lcov.info
\ No newline at end of file
+lcov.info
+Cargo.lock
```
leighmcculloch, line 5: Thanks, this shouldn't have been in here in the first place. 👍🏻

Cargo.lock
```diff
@@ -1022,9 +1022,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.69"
+version = "1.0.78"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "134c189feb4956b20f6f547d2cf727d4c0fe06722b20a0eec87ed445a97f92da"
+checksum = "e2422ad645d89c99f8f3e6b88a9fdeca7fabeac836b1002371c4367c8f984aae"
```
leighmcculloch, line 1027: Can you change this back to .69 since it isn't necessary for this to select the newer version inside this repo?

leighmcculloch (commented):
Couple more asks.

Cargo.lock
salaheldinsoliman: Shouldn't I remove the whole file instead?

leighmcculloch: There shouldn't be any changes to the lock file, yup.

.gitignore
leighmcculloch: I misread this diff, I thought it was removing Cargo.lock, but on relook I see it is adding it. We don't want the Cargo.lock file to be excluded from git, for the reasons described here: https://github.com/stellar/rs-soroban-env/pull/1351#issuecomment-1951513131

PR #1354, opened 2024-02-24
Refactor soroban-simulation to use the e2e_invoke

dmkozh:
### What

Refactor soroban-simulation to use the e2e_invoke and also to be more modular/reusable.

- Use e2e_invoke for invoking host functions in recording mode
- Separate network config and adjustment config into separate structs for easier testing/caching etc.
- Clean up the storage logic to clearly distinguish between snapshots with and without archive access
- Added some documentation for public functions/structs
- Added some basic test coverage. I didn't try to verify the math for every number, but I did sanity-check them. The host function simulation accuracy is mostly covered on `e2e_invoke` side.
- Reuse host code for common test and non-test utilities

### Why

Improving simulation quality, as well as maintainability.

### Known limitations

N/A


sisuresh (approved):

2opremio:
Another option is to do this in a two-step process, keeping an externally-compatible layer at first.

dmkozh:
> Did you take into consideration the impact at https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/lib/preflight/src/lib.rs and whether it satisfies all the current functionality?

Of course, these changes are meant to be functionally no-op (besides the additional capabilities, like ledger diffs we've discussed before - but there is no need to use them immediately). I'll definitely help with the migration, but this PR has to be finalized first.

soroban-env-host/src/e2e_invoke.rs
```diff
@@ -151,8 +160,15 @@ pub fn get_ledger_changes(
                 new_live_until_ledger: 0,
             });
         }
-        if init_storage_snapshot.has(key)? {
-            let (old_entry, old_live_until_ledger) = init_storage_snapshot.get(key)?;
+        #[cfg(any(test, feature = "unstable-next-api"))]
+        let entry_with_live_until = init_storage_snapshot.get(key)?;
+        #[cfg(not(any(test, feature = "unstable-next-api")))]
+        let entry_with_live_until = if init_storage_snapshot.has(key)? {
```
leighmcculloch, line 166: The fact that `test` is included in here as an any condition suggests that the non-unstable code is not tested. How can we still test it?

leighmcculloch (changes requested):
A couple asks are inline. At minimum can we fix the docs issue before this is released?

soroban-env-host/src/e2e_invoke.rs
dmkozh: I don't think there is anything to test in this particular case. `SnapshotSource` is not really covered directly. The new field is covered in the tests as well and I don't see much value in covering not returning this field.

PR #1355, opened 2024-02-24
Proposal: Allow Extending Instance and Code TTL With Separate Values on the Host Environment For More Cost-Efficient Implementations

heytdep:
### What

Proposal specification at https://docs.google.com/document/d/1f2hWD4LRL9A27rGImvavbBCiPUa3mYIXpcxtTfxFsLI/edit?usp=sharing.

In short, this would enable the guest environment to bump the instance and the contract code with different ttl and threshold values leading to better distribution of rent costs among binaries designed to be associated with more than one instance.


namankumar:
Discussion around this proposal is here: https://github.com/stellar/stellar-protocol/discussions/1447

soroban-env-common/env.json
```diff
@@ -1422,6 +1422,58 @@
                 },
                 {
                     "export": "a",
+                    "name": "extend_current_contract_instance_and_code_ttl_separately",
```
dmkozh, line 1425: We can't modify the host function signatures, only new functions can be added. They also need a protocol guard.

soroban-env-common/env.json
```diff
@@ -1437,7 +1489,7 @@
                     "docs": "Get the id of a contract without creating it. `deployer` is address of the contract deployer. `salt` is used to create a unique contract id. Returns the address of the would-be contract."
                 },
                 {
-                    "export": "b",
+                    "export": "d",
```
dmkozh, line 1492: The other host functions can't be changed as well.

soroban-env-host/src/builtin_contracts/account_contract.rs
```diff
@@ -160,6 +160,13 @@ pub(crate) fn check_account_authentication(
             len
         ));
     }
+    if len == 0 {
```
dmkozh, line 163: Seems like you need to rebase your PR to get the proper diffs.

soroban-env-common/env.json
heytdep: Ah this is the feedback I was looking for, makes sense thanks. Is there an interface reference for setting the protocol guard for the function?

dmkozh: Protocol gating has just been merged: https://github.com/stellar/rs-soroban-env/pull/1348 . Just follow the example of `protocol_gated_dummy`, but use protocol 21 for min supported version.

soroban-env-host/src/host/data_helper.rs
```diff
@@ -236,18 +236,23 @@ impl Host {
     pub(crate) fn extend_contract_instance_and_code_ttl_from_contract_id(
         &self,
         contract_id: &Hash,
-        threshold: u32,
-        extend_to: u32,
+        instance_threshold: u32,
+        code_threshold: Option<u32>,
+        extend_instance_to: u32,
+        extend_code_to: Option<u32>,
```
jayz22, line 242: I think in general having optional fields in function signatures are confusing. 
I would prefer the other way around: make these inputs non-optional and change the existing callers to explicitly passing them in. The change surface area should be similar and it's less confusing.

heytdep: Yeah if we go with https://github.com/stellar/stellar-protocol/discussions/1447#discussioncomment-8591665 this fn would be "duplicated" into `extend_contract_code_ttl_from_contract_id` for bumping contract code and the existing helper modified to `extend_contract_instance_ttl_from_contract_id` which only bumps the instance.

soroban-env-host/src/host.rs
```diff
@@ -3003,6 +3054,13 @@ impl Host {
     {
         f(self.0.budget.clone())
     }
+
+    pub fn retrieve_entry_with_lifetime(
```
dmkozh, line 3058: It seems like this function can be moved to "testutils". It also should be `pub(crate)`, not `pub`

soroban-env-host/src/host/data_helper.rs
```diff
@@ -236,18 +236,28 @@ impl Host {
     pub(crate) fn extend_contract_instance_and_code_ttl_from_contract_id(
         &self,
         contract_id: &Hash,
-        threshold: u32,
-        extend_to: u32,
+        instance_threshold: u32,
+        code_threshold: Option<u32>,
```
dmkozh, line 240: This change is no longer necessary given the interface changes, right?

soroban-env-host/src/storage.rs
```diff
@@ -476,3 +476,19 @@ impl Storage {
...
+impl Storage {
+    /// Unmetered function to access entry with lifetimes to test lifetime extensions.
+    pub fn get_entry_with_lifetime(
+        &self,
+        key: Rc<LedgerKey>,
+    ) -> Result<Option<EntryWithLiveUntil>, HostError> {
+        let entry = self.map.map.iter().find(|e| e.0 == key);
+
+        match entry {
+            None => Err((ScErrorType::Storage, ScErrorCode::InternalError).into()),
```
dmkozh, line 490: Why is this an error? It seems like you should just return the retrieved `entry` (i.e. `None`)

soroban-env-host/src/storage.rs
```diff
@@ -476,3 +476,19 @@ impl Storage {
...
     }
 }
+
+#[cfg(any(test, feature = "testutils"))]
+impl Storage {
+    /// Unmetered function to access entry with lifetimes to test lifetime extensions.
+    pub fn get_entry_with_lifetime(
+        &self,
+        key: Rc<LedgerKey>,
+    ) -> Result<Option<EntryWithLiveUntil>, HostError> {
```
dmkozh, line 486: The tests seem to be only interested in lifetime... So why not make this to just return lifetime?

soroban-env-host/src/test/lifetime_extension.rs
```diff
@@ -0,0 +1,111 @@
...
+    contract: Hash,
+    code: Hash,
+}
+
+impl InstanceCodeTest {
+    // We can potentially add some customizability for the ledger here.
+    fn setup() -> Self {
+        let host = Host::test_host_with_recording_footprint();
+        let contract_id = host.register_test_contract_wasm(CONTRACT_STORAGE);
+        let hash = if let ScVal::Address(addr) =
```
dmkozh, line 20: You can use `host.contract_id_from_address` instead.

soroban-env-host/src/test/lifetime_extension.rs
```diff
@@ -0,0 +1,111 @@
...
+        let code = if let ContractExecutable::Wasm(hash) = host
+            .retrieve_contract_instance_from_storage(
+                &host.contract_instance_ledger_key(&hash).unwrap(),
+            )
+            .unwrap()
+            .executable
+        {
+            hash
+        } else {
+            panic!("Expected WASM executable")
```
dmkozh, line 40: WASM->Wasm

soroban-env-host/src/test/lifetime_extension.rs
```diff
@@ -0,0 +1,111 @@
...
+        };
+
+        host.set_ledger_info(crate::LedgerInfo {
+            protocol_version: 21,
+            sequence_number: 4090,
+            timestamp: Default::default(),
```
dmkozh, line 46: Can't you just use `..Default::default()`?

soroban-env-common/env.json
```diff
@@ -1398,7 +1398,8 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the TTL for the current contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the TTL for the current contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger",
+                    "max_supported_protocol": 21
```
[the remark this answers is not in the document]
jayz22: Correct. You can't do that yet. The `max_supported_protocol` mechanism is there to retire a function if no contract on chain still links to it (maybe gradually over time it can be achieved, but the exact mechanism to determine that is still a bit TBD), and we can't break contracts by forcing a function to retire.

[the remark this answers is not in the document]
heytdep: Yeah that's true, I completely overlooked the fact that Soroban is already on pubnet 😬 , removing the protocol guard now.

soroban-env-host/src/storage.rs
heytdep: Yeah I copied the logic from try_get_full but since this is just a testutil without prepare_read_only_access a missing key doesn't mean it's an internal error so returning just the option sounds more correct.

soroban-env-host/src/storage.rs
heytdep: Yes, the test only needs the lifetime, but since there seems to be no other way to do an unmetered get on an entry from the host this might be used in the future. So it's just an addition to allow for more test customizability in the future, but can for sure reshape it to return the lifetime, let me know.

soroban-env-host/src/host.rs
heytdep: If I move this to testutils I won't be able to access the storage which is private. Should I keep this as is or add a testutils borrower within the host mod and call it from the testutils?

soroban-env-host/src/host.rs
```diff
@@ -3003,6 +3050,14 @@ impl Host {
     {
         f(self.0.budget.clone())
     }
+
+    #[allow(dead_code)]
```
sisuresh, line 3054: This should be behind a `#[cfg(test)]` feature gate instead.

soroban-env-host/src/host.rs
```diff
@@ -2087,11 +2108,37 @@ impl VmCallerEnv for Host {
...
+                self.try_borrow_storage_mut()?
+                    .extend_ttl(self, key, threshold.into(), extend_to.into())
+                    .map_err(|e| self.decorate_contract_code_storage_error(e, &wasm_hash))?;
+            }
+            ContractExecutable::StellarAsset => {}
+        }
+
+        Ok(Val::VOID)
+    }
+
```
jayz22, line 2141: Nit: I notice this code (and one above) is duplicate of `extend_contract_instance_and_code_ttl_from_contract_id`. Would you mind a minor refactoring to avoid duplication? The logic here isn't too complex so it isn't a huge deal, it's just for better maintainability. 
You can just add a new helper in `data_helper` to extend the code, and change `extend_contract_instance_and_code_ttl_from_contract_id` to only extend contract instance (or vice versa).

jayz22 (changes requested):
Looks good in general, just a few nits and we should be ready to land it (don't forget to run `cargo fmt`)

soroban-env-host/src/test/lifetime_extension.rs
```diff
@@ -0,0 +1,96 @@
...
+            .extend_contract_code_ttl(contract_id, 5.into(), 5000.into())
+            .is_ok());
+        assert_eq!(
+            host.retrieve_entry_with_lifetime(host.contract_code_ledger_key(&code).unwrap())
+                .unwrap()
+                .1,
+            Some(9090)
+        );
+    }
+}
```
jayz22, line 96: Could you also add a test case for extending both (the currently existing host function)?

sisuresh: Done

soroban-env-host/src/storage.rs
```diff
@@ -476,3 +476,19 @@ impl Storage {
...
+        key: Rc<LedgerKey>,
+    ) -> Option<EntryWithLiveUntil> {
+        let entry = self.map.map.iter().find(|e| e.0 == key);
+
+        match entry {
+            None => None,
+            Some(pair_option) => pair_option.1.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
sisuresh: I removed the helper

soroban-env-common/env.json
```diff
@@ -1447,6 +1447,48 @@
...
+                            "name": "threshold",
+                            "type": "U32Val"
+                        },
+                        {
+                            "name": "extend_to",
+                            "type": "U32Val"
+                        }
+                    ],
+                    "return": "Void",
+                    "docs": "If the TTL for the provided contract instance (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger.",
```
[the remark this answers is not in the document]
sisuresh: Fixed

soroban-env-host/src/host.rs
```diff
@@ -2071,11 +2071,32 @@ impl VmCallerEnv for Host {
         self.extend_contract_instance_and_code_ttl_from_contract_id(
             &contract_id,
             threshold.into(),
-            extend_to.into(),
+            extend_to.into()
```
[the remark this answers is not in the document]
sisuresh: Done

soroban-env-host/src/host.rs
sisuresh: Removed

jayz22:
Closing. Superseded by https://github.com/stellar/rs-soroban-env/pull/1368

PR #1375, opened 2024-03-28
Module cache followup

graydon:
This is a set of followups to #1359 to cover missing testing and wiring around the module cache. It includes calibration of the `VmCachedInstantiation` cost type, a shift to using a single shared linker when using a module cache (this turns out to be critical -- the existing code deadlocks on cross-contract calls) and some fairly subtle wiring to patch up costs in recording mode, along with some tests to check that it's mostly working. 

Several tests remain broken with --feature=next, so I wouldn't say it's quite ready yet. I tried, but am out of time for tonight.


soroban-env-host/src/e2e_invoke.rs
```diff
@@ -533,6 +534,19 @@ pub fn invoke_host_function_in_recording_mode(
...
+        // don't know what contracts will be accessed -- to populate the cache
+        // with -- until after invocation, so we still treat every access as a
+        // hit while the invocation is running and then we charge the population
+        // here _after_ the invocation. It's a bit of a hack but produces the
+        // right cost (on success), just in the wrong order. The main
+        // disadvantage is that invocations that return errors will have
+        // observably-different costs than their real invocations would have;
+        // but we assume people don't care a lot about the cost accuracy of
+        // invocations that return errors.
+        host.maybe_add_module_cache()?;
```
dmkozh, line 548: Can we do that on exit from the top-level frame, similar to what we do `maybe_emulate_authentication`. This will make the metering a bit more accurate for tests as well. This also is resilient to failures and probably will have a bit better error reporting in case if we run out of budget during populating the cache.

soroban-env-host/src/test/e2e_tests.rs
```diff
@@ -1737,3 +1737,69 @@ fn test_classic_account_auth_using_simulation() {
...
+            (
+                cd.wasm_entry.clone(),
+                Some(ledger_info.sequence_number + 100),
+            ),
+            (
+                cd.contract_entry.clone(),
+                Some(ledger_info.sequence_number + 1000),
+            ),
+        ];
+        let res = invoke_host_function_recording_helper(
```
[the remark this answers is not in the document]
graydon: I don't think so -- I need the resources to feed from the first call to the second. Unless I misunderstand.

soroban-env-host/src/e2e_invoke.rs
graydon: Oh, yeah, I think that works. Ok doing so.

soroban-env-host/src/vm.rs
```diff
@@ -148,41 +153,41 @@ impl Vm {
...
+                    if !module_symbols.contains(&(hf.mod_str, hf.fn_str)) {
                         continue;
                     }
-                }
-                if let Some(max_proto) = hf.max_proto {
-                    if parsed_module.proto_version > max_proto || ledger_proto > max_proto {
-                        // We skip linking this hf instead of returning an error
-                        // because we have to support old contracts during replay.
-                        continue;
+                    if let Some(min_proto) = hf.min_proto {
```
graydon, line 168: @jayz22 I would appreciate your review of this section. I note you explicitly say _not_ to return an error here in the previous code, but unfortunately we really can't "omit the symbols" in the new scheme, since multiple contracts share a single linker. I have tried to emulate the error code correctly but I don't understand why the comment before suggested that returning an error wouldn't work.

soroban-env-host/src/test/e2e_tests.rs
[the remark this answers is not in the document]
dmkozh: That's what that function does - it runs preflight with recording mode, then in enforcing mode using the resources from the recording mode. I need to add documentation to these function, sorry for being unclear.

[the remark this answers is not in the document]
dmkozh: I see that you have some additional code related to the Wasm entries; let's keep this as is for now. I have a more generic change that allows testing the whole module with different protocol versions.

soroban-env-host/src/vm.rs
dmkozh: IIUC we need to have indistinguishable behavior between host functions that don't exist and host functions that are not supported yet by the protocol. What would happen if I upload a contract that imports a secp host function today? (it's not yet present in the env) If we return an error during VM instantiation, then I guess we're good.

soroban-env-host/src/vm.rs
```diff
@@ -260,8 +270,76 @@ impl Vm {
...
+    /// _after_ a top-level recording-mode invocation completes, by reading the
+    /// storage and parsing all the modules in it, in order to charge for
+    /// parsing each used module _once_ and thereby produce a mostly-correct
+    /// total cost.
+    ///
+    /// We still charge the reparses to the shadow budget, to avoid a DoS risk,
+    /// and we still charge the instantiations to the real budget, to behave the
+    /// same as if we had a cache.
+    ///
+    /// Note that this will only produce "mostly correct" costs for recording in
```
dmkozh, line 317: Is this paragraph still true, given that we build the cache independently of the return value?

soroban-env-host/src/vm.rs
jayz22: I believe you've handled it correctly here and there won't be an observable difference. 

But let me elaborate on my code comment with a slightly contrived example. 
In the protocol today (p20), if a contract contains a call to a non-existent host function, `{"c", "3"}` (i.e. secp256r1_verify, available in p21). What the p20 Vm does is:
1. for every existing host function, define it (`linker.define(func)`). There could be an error arriving from `linker.define` call, such as duplicate definition, missing/invalid type etc. 
2. `linker.instantiate`. Attempting to link `{"c", "3"}` in p20, arrives at a `LinkerError::MissingDefinition` in this step. 

Imagine a function in the future protocol replays this contract. It sees the contract attempting to link a host function non-existent to it, in step 1 above, and throws an error immediately. We could potentially be modifying the contract's expected behavior. What if the contract actually failed due to `linker.define` of an existing function in step 1 (that would be have an order later than `{"c", "3"}`), we would be hijacking that by throwing a different error as soon as we saw `{"c", "3"}`. 

This case cannot happen today for p20, because all linker errors are treated as `(WasmVm, InvalidAction)`. Thus we can't be changing observable result. 

My comment above was trying to emphasize that by skipping linking (`continue` instead of `Err`), it was the most safe/future-proof way to ensure the linker preserves the expected protocol behavior.

soroban-env-host/src/vm.rs
graydon: No, you're right, it's misleading now. I'll remove that paragraph.

soroban-env-host/src/test/lifecycle.rs
```diff
@@ -942,4 +995,148 @@ mod cap_54_55_56 {
...
+    // endregion: CAP-0054 refined cost model
+
+    // region: CAP-0056 ModuleCache related tests
+
+    // Test that running on protocol vOld does not make a ModuleCache at all.
+    #[test]
+    fn test_v_old_no_module_cache() -> Result<(), HostError> {
+        let host = upload_and_make_host_for_next_ledger(
+            "test_v_old_no_module_cache_upload",
+            V_OLD,
```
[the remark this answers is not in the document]
graydon: V_OLD and V_NEW are explicitly defined in this test with reference to the protocol boundary at which the module cache is made-available, as just-before and just-after

dmkozh (approved):

soroban-env-host/src/vm.rs
graydon: @dmkozh if you upload a function today that is not supported in v21, you get a linker error from wasmi concerning a missing definition during the trial instantiation, which turns into (WasmVm, InvalidAction) error code.

in this new code, the explicit branch that tests the "calling a function not yet supported" condition on line 169 will fail, and return a (WasmVm, InvalidAction).

There are tests for all these sorts of host-vs-contract-vs-protocol ordering cases in `test_host_protocol_gating_for_wasm` that @jayz22 wrote before. This new code passes them all (produces the same results).

The only difference -- as far as I intended and can tell, please point to a way this might be wrong! -- is that the possibility of _registering host functions with the linker_ used to occur inside this loop, whereas now it occurs before the loop altogether (when constructing the shared linker). But I think this is ok: registering host functions shouldn't depend in any way on user input, it's driven by a static list of host functions compiled-in to the env and it should always succeed (as it does in all the tests we run here). If it fails it's an internal error, a bug in the env or something, I don't think we need to consider "the possibility of a change in the timing of an internal error that, if it ever happened, would be based on solely static data" to be an "observable change".

(And if this error _did_ occur in p20, and someone _did_ somehow figure out a way to observe it, I think we'd discover the divergence during the sealed-replay phase of the planned upgrade and we could just mark it as a special case. But since it doesn't fail in practice, as far as we know, even this shouldn't be possible).

jayz22: Agree that no observable difference should be able to occur due to rearranging the order where host functions are registered. Host function registration (`linker.define` in the old code) is just inserting a static list of host functions and checking duplication. There is no chance for it to fail unless there is an internal bug. And we have taken measures to ensure (statically) that host function definitions can not duplicate.

soroban-env-host/src/vm/parsed_module.rs
```diff
@@ -153,32 +153,62 @@ impl ParsedModule {
...
+    pub fn with_import_symbols<T>(
+        &self,
+        callback: impl FnOnce(&BTreeSet<(&str, &str)>) -> Result<T, HostError>,
+    ) -> Result<T, HostError> {
+        // Cap symbols we're willing to import at 10 characters for each of
+        // module and function name. in practice they are all 1-2 chars, but
+        // we'll leave some future-proofing room here. The important point
+        // is to not be introducing a DoS vector.
+        const SYM_LEN_LIMIT: usize = 10;
+        let symbols: BTreeSet<(&str, &str)> = self
```
jayz22, line 175: Erring on extra caution, would it be plausible to use a `MeteredVec` here? 
There is a chance the contract can have lots of import entries (greater than number of available host functions), in which case the set construction and followup look ups costs are non-trivial.
Using `MeteredVec` also avoids dependency on the hash algorithm that isn't totally under our control.

dmkozh:
> There is seemingly one more deadlock that has to be fixed before merging this

I've actually accidentally picked up the old revision of this PR. I've double-checked that in the current state it doesn't deadlock. Sorry for confusion.

dmkozh (approved):

soroban-env-host/src/vm/parsed_module.rs
graydon: The costs are accounted for in the new cost model: it has a term specifically for the number of imports, which is calibrated to cover this cost.

There is no dependency on a hash algorithm here. BTreeSet is a b-tree, it does comparisons. I also bounded the length of import symbols at 10 characters, so there's no possibility of an over-long comparison.

jayz22: sounds good

PR #1382, opened 2024-04-04
Module cache upload and run

graydon:
This covers a two cases I failed to cover in #1375 that @dmkozh and @sisuresh have identified:

  - User calls contract without putting wasm for call in footprint or storage
  - User calls `upload_contract` (or its equivalent, `update_current_contract_wasm`) mid-execution and then calls the updated-or-new contract

As well as a third case that is, as far as I know, not possible; but it seems to me it doesn't hurt to be defensive since someone could make it possible in the future fairly easily and forget to cover this case in the cache code:

  - User deletes wasm in transaction, and then tries to call it.

As always, there is probably room for a bit more testing here, but I wanted to get these cases covered and fixed asap.

This should make-redundant PR #1380


soroban-env-host/src/test/lifecycle.rs
```diff
@@ -1140,5 +1142,193 @@ mod cap_54_55_56 {
...
+        let wasm_key = host.contract_code_ledger_key(&wasm_hash)?;
+        host.with_mut_storage(|storage| {
+            let budget = host.budget_cloned();
+            storage.footprint.0 = storage
+                .footprint
+                .0
+                .remove::<Rc<LedgerKey>>(&wasm_key, &budget)?
+                .unwrap()
+                .0;
+            storage.map = storage
```
dmkozh, line 1175: Can we also have a test case where the entry is `None` in map, i.e. Wasm is in the footprint, but doesn't exist? The expected error should be `Storage, MissingValue` (I believe the code is correct, but it's nice to have a supporting test as well).

graydon: sure I'll tack one on

dmkozh (approved):

PR #1383, opened 2024-04-05
Protocol-version-qualify test observations

graydon:
This does two things, which are steps towards removing next-guards:

  - Make all existing observations protocol-qualified, in a subdirectory like `observations/20/...`
  - Update all tests to use hosts that at least _have_ a protocol, taken from the `Host::test_host` path that recently gained an environment variable to control its protocol


leighmcculloch (approved):
👏🏻

jayz22 (approved):

PR #1384, opened 2024-04-05
Protocol 21

graydon:
This is now ready to merge. Does the following:

  - [x] take new XDR from https://github.com/stellar/rs-stellar-xdr/pull/359
  - [x] bump current protocol 20->21 and next 21->22 
  - [x] remove `feature="next"` gates guarding protocol 21 functionality (note: this is not _all_ `feature="next"` gates, some test the very concept of next-ness and should remain)
  - [ ] defer to next PR: ~bump workspace version to 21.0.0~
  - [ ] defer to next PR: ~remove `feature="unstable-next-api"` gates~
  - [x] run observations for protocol 21
  - [x] modify CI to run tests twice, once with TEST_PROTOCOL=20 and once with TEST_PROTOCOL=21
  - [x] modify recent VMCaller-using secp256r1 tests to go through the env now that it supports those host functions
  - [x] fix all tests broken by the above


soroban-synth-wasm/src/mod_emitter.rs
```diff
@@ -84,17 +83,33 @@ impl Default for ModEmitter {
 impl ModEmitter {
     pub fn from_configs(mem_pages: u32, elem_count: u32) -> Self {
         let mut me = Self::new();
-        me.custom_section(
-            soroban_env_common::meta::ENV_META_V0_SECTION_NAME,
-            &soroban_env_common::meta::XDR,
-        );
+        me.add_test_protocol_version_meta();
```
dmkozh, line 86: Is the plan here to let core to programmatically set the env variable if it wants a different protocol version? Without that Core risks generating p21 Wasm while being in p20 ledger. That's probably fine, though I think it would be a bit less footgun-y if we could pass the protocol version directly through the function interface (we could do that only the the Core wrapper though).

soroban-env-host/src/test/lifecycle.rs
```diff
@@ -750,7 +748,6 @@ mod cap_54_55_56 {
         budget: Budget,
     ) -> Result<ObservedHost, HostError> {
         let host = Host::with_storage_and_budget(storage, budget);
-        host.enable_debug()?;
```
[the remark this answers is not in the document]
graydon: Oh, I don't feel strongly, I just removed them because I had added them earlier to debug the tests. I can put them back.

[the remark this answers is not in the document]
dmkozh: Ah, I see. I think it's a good idea to have this on by default, if only to save some time modifying and recompiling with this.

soroban-synth-wasm/src/mod_emitter.rs
graydon: Well, absent an env var, you will get what you got before: the current crate's env version. This change is a no-op in that sense, assuming the caller doesn't set TEST_PROTOCOL in their environment.

You're right that this code path -- both the existing code and the code under this change -- will probably cause loadgen tests and so forth to break when spraying v21 wasms at a core still running on v20. I think the tests on core side will need to change at least a bit to plumb an expected protocol number in. They're not doing that at all currently. I think loadgen-y tests will just be broken until that happens, regardless of this change.

I wonder if it's wise for us to have this nontrivial `ModEmitter::default` at all. Maybe it should just call the empty-emitter-making `ModEmitter::new`, and we can make some more convenience ctors that do useful things (`with_protocol`, `with_current_test_protocol`, `with_current_test_protocol_and_misc_content`, etc.)

I'm not sure I want to do that in _this_ PR, though it'll break semantics to do so in the future, it'll probably be just additive API. Not sure. Do you have any preference?

dmkozh: I think it's fine to leave this as is for this PR, just wanted to check if there is any plan around this.

dmkozh (approved):

sisuresh (approved):

jayz22 (approved):

soroban-synth-wasm/src/mod_emitter.rs
graydon: Reverted ModEmitter::default to emit a hard-wired-to-20 module and added default_with_test_protocol, switched tests to use that. Thanks for the suggestion! This will likely make core break less.

PR #1386, opened 2024-04-06
Observe xdr

graydon:
While working on https://github.com/stellar/rs-soroban-env/pull/1384 we discovered that some of the trace/observation hashes changed (specifically those of storage).

This happens because we currently (on main) hash the Rust representation of elements in storage, and between v20 and v21 we replaced an `ExtensionPoint` in the `ContractCode` LE type with a new type (storing the new cost models), and unfortunately while replacing `ExtensionPoint`s with different-enums-that-have-a-zero-case in XDR is unobservable, replacing them in Rust changes the size of the type from 0-sized and trivially-hashed to whatever-the-new-size-is and nontrivially-hashed. This changes observations.

The solution to this is kinda gross: we start hashing the XDR representations of storage (and footprints). And moreover we want to do this _before_ we start ungating stuff in https://github.com/stellar/rs-soroban-env/pull/1384 and risking masking a change to the actual semantics.

So this change -- that _only_ changes the storage-and-footprint hashes and obviously _only_ does so because the trace/observe code changed -- should go in first. Before https://github.com/stellar/rs-soroban-env/pull/1384.

NB: when we were originally discussing this, we thought maybe events should also be hashed this way. Turns out we store events in the host in a decomposed form -- a bunch of plain structs full of `Val`s, similar to `HostObject`s -- to minimize the cost of never-read diagnostic events. So they don't need to be done. There are no XDR fields much less any with `ExtensionPoints` in them, in the form we're hashing them.


soroban-env-host/src/host/trace.rs
```diff
@@ -278,8 +279,31 @@ impl Host {
...
+        // we can't use normal Rust hashing for this type. Instead we manually walk the
+        // map and hash the XDR of all the XDR-bits in it.
         if let Ok(store) = self.0.storage.try_borrow() {
-            self.hash_one(&store.map)
+            let mut state = CountingHasher::default();
+            let budget = self.budget_ref();
+            budget.with_shadow_mode(|| {
+                store.map.len().metered_hash(&mut state, budget)?;
+                for (k, v) in store.map.iter(budget)? {
+                    k.metered_hash_xdr(&mut state, Limits::none(), budget)?;
```
dmkozh, line 292: Why not use `DEFAULT_XDR_RW_LIMITS`? I realize that we mostly use observations for tests, but since we might want to reuse them for debugging, it would be safer to use the default limit here (unless we want to observe deeply nested values? but these would lead to stack overflow anyway...)

dmkozh (approved):

PR #1393, opened 2024-04-10
Give necessary permissions to bump-version workflow.

dmkozh:
### What

Give necessary permissions to bump-version workflow.

This is likely a temporary fix until we figure out whether this should be done in the top-level workflow, or fixed in the default settings and removed.

### Why

To unblock bump-version workflow

### Known limitations

N/A

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -11,5 +11,9 @@ jobs:
 
   bump-version:
     uses: stellar/actions/.github/workflows/rust-bump-version.yml@main
+    permissions:
+      actions: write
+      contents: write
+      pull-requests: write    
     with:
       version: ${{ inputs.version }}


sisuresh (approved):

dmkozh:
We don't need to merge this right now, let's keep this around until we either need to do the bump, or figure out a more general fix.

leighmcculloch:
Discussing with the ops team [here](https://stellarfoundation.slack.com/archives/C02U19A2A/p1712783860153229?thread_ts=1712776319.252229&cid=C02U19A2A) to understand if we can change the default permissions back or if this is the new permission setup. If we can't change the default back this change will need applying to the 11 repositories listed here:
- https://github.com/search?q=org:stellar+path:**/bump-version.yml&type=code

PR #1401, opened 2024-04-12
Improvements for recording mode metering of module cache

dmkozh:
### What

This involves several improvements around various module cache usage edge cases:

- Use the initial storage state to build the cache 'post-hoc' in order to not build cache for Wasms that wouldn't be cached in enforcing mode
- Only charge the module parse costs to 'shadow' budget when we would be using the cached module in enforcing mode
- Only rebuild the module cache in recording mode when it would be built in the enforcing mode.

### Why

Improving simulation accuracy.

### Known limitations

N/A


soroban-env-host/src/host/frame.rs
```diff
@@ -737,7 +721,24 @@ impl Host {
                     let (code, costs) = self.retrieve_wasm_from_storage(&wasm_hash)?;
 
                     #[cfg(any(test, feature = "recording_mode"))]
-                    let cost_mode = crate::vm::ModuleParseCostMode::PossiblyDeferredIfRecording;
+                    // let cost_mode = crate::vm::ModuleParseCostMode::PossiblyDeferredIfRecording;
```
graydon, line 724:
```suggestion
// In recording mode: if a contract was present in the initial snapshot image, it is part of
// the set of contracts that would have been built into a module cache in enforcing mode;
// we want to defer the cost of parsing those (simulating them as cache hits) and then charge
// once for each such contract the simulated-module-cache-build that happens at the end of
// the frame in [`Self::pop_context`].
//
// If a contract is _not_ in the initial snapshot image, it's because someone just uploaded
// it during execution. Those would be cache misses in enforcing mode, and so it is right to
// continue to charge for them as such (charging the parse cost on each call) in recording.
```

graydon (changes requested):
Two giant thumbs up. This is a huge improvement. Thanks. I added one comment-suggestion.

graydon (approved):

Issue #1329, opened 2024-01-17
Extending the ttl of a temporal storage to exisiting maximum live until ledger would always fail


yanliu18:
### What version are you using?
20.1.0

### What did you do?
Formal Auditting for Client Code
https://github.com/stellar/rs-soroban-env/blob/36d33cb6c986c9a8a9200b7eb04cf02e2c3f0ef4/soroban-env-host/src/storage.rs#L382

When call to the `extend_ttl()` function in host/storage, when the function parameter `extend_to` = `li.max_entry_ttl` (`li.max_entry_ttl` = `host.with_ledger_info(|li| li.max_entry_ttl)`) and parameter `key` matches `ContractData(LedgerKeyContractData {_, _, StorageType::Temporal, })`, the call to `extend_ttl` will always return an error.

This is because of the following condition checking where,
`new_live_until` = `li_sequece_number` + extend_to
`host.max_live_until_ledger()` = `li_sequece_number` + li.max_entry_ttl - 1
will always go to the `else` branch.
https://github.com/stellar/rs-soroban-env/blob/36d33cb6c986c9a8a9200b7eb04cf02e2c3f0ef4/soroban-env-host/src/storage.rs#L436-L446

### What did you expect to see?
Though it is rare for a contract instance to set its temporal stroage to the exisiting max livable ledger sequence, it is still possible.
Maybe it is reasonable to calculate the `new_live_until` in the same way as the `max_live_live_until_ledger`.

jayz22:
Thanks @yanliu18 for reporting this issue. Indeed some of the terminologies defined can be a little confusing. 
You've brought up a couple good points worth clarifying:
- The internal logic for calculating `new_live_until` is treating the input as "extend by" (not including the current ledger) where as the `li.max_entry_ttl` is being treated as number of ledgers to live including the current ledger. 
- The `Temporary` entry cannot be attempted to be extended past the ledger max, it throws error instead of being clamped to the max. This was an intentional design (@dmkozh gave a good explanation [here](https://discord.com/channels/897514728459468821/1235586978247082024/1235628500975747113)) but it's worth clarifying in the documentation.

Although not a protocol bug, I think both of these points worth being clarified via better documentation, I will open a PR to do so. 
The SDK has already been updated to make sure the `max_entry_ttl` getter takes care of the boundary offset: https://github.com/stellar/rs-soroban-sdk/pull/1269, so that user can't trigger the error using it.

yanliu18:
[...]

PR #1413, opened 2024-05-03
Update docs to clarify `extend_ttl` return conditions

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/1329

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-common/env.json
+++ b/soroban-env-common/env.json
@@ -1382,7 +1382,7 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the entry's TTL is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the entry's TTL is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger. If attempting to extend the entry past the maximum allowed value (defined as the current ledger + `max_entry_ttl` - 1), and the entry is `Persistent`, its new `live_until_ledger_seq` will be clamped to the max; if the entry is `Temporary`, an error is returned."
                 },
                 {
                     "export": "8",
@@ -1398,7 +1398,7 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the TTL for the current contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the TTL for the current contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger. If attempting to extend the entry past the maximum allowed value (defined as the current ledger + `max_entry_ttl` - 1), and the entry is `Persistent`, its new `live_until_ledger_seq` will be clamped to the max; if the entry is `Temporary`, an error is returned."
                 },
                 {
                     "export": "9",
@@ -1418,7 +1418,7 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the TTL for the provided contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the TTL for the provided contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger. If attempting to extend the entry past the maximum allowed value (defined as the current ledger + `max_entry_ttl` - 1), and the entry is `Persistent`, its new `live_until_ledger_seq` will be clamped to the max; if the entry is `Temporary`, an error is returned."
                 },
                 {
                     "export": "a",
--- a/soroban-env-common/env.json
+++ b/soroban-env-common/env.json
@@ -1382,7 +1382,7 @@
                         }
                     ],
                     "return": "Void",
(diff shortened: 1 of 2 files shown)


soroban-env-common/env.json
```diff
@@ -1382,7 +1382,7 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the entry's TTL is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the entry's TTL is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger. If attempting to extend the entry past the maximum allowed value (defined as the current ledger + `max_entry_ttl` - 1), and the entry is `Persistent`, its new `live_until_ledger_seq` will be clamped to the max; if the entry is `Temporary`, an error is returned."
```
dmkozh, line 1385: Env doesn't return an error, it traps. Please update the doc.

soroban-env-common/env.json
```diff
@@ -1398,7 +1398,7 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the TTL for the current contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the TTL for the current contract instance and code (if applicable) is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger. If attempting to extend the entry past the maximum allowed value (defined as the current ledger + `max_entry_ttl` - 1), and the entry is `Persistent`, its new `live_until_ledger_seq` will be clamped to the max; if the entry is `Temporary`, an error is returned."
```
dmkozh, line 1401: Instance and code are always persistent, please remove temporary storage mentions here and below.

soroban-env-common/env.json
jayz22: Fixed.

soroban-env-common/env.json
jayz22: yeah good point. removed.

soroban-env-common/env.json
```diff
@@ -1382,7 +1382,7 @@
                         }
                     ],
                     "return": "Void",
-                    "docs": "If the entry's TTL is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger"
+                    "docs": "If the entry's TTL is below `threshold` ledgers, extend `live_until_ledger_seq` such that TTL == `extend_to`, where TTL is defined as live_until_ledger_seq - current ledger. If attempting to extend the entry past the maximum allowed value (defined as the current ledger + `max_entry_ttl` - 1), and the entry is `Persistent`, its new `live_until_ledger_seq` will be clamped to the max; if the entry is `Temporary`, it traps."
```
dmkozh, line 1385: nit: 'it traps' -> 'the function traps' (it's unclear what's 'it' in this sentence)

soroban-env-host/src/storage.rs
```diff
@@ -449,10 +449,15 @@ impl Storage {
         Ok(self.try_get_full_with_host(key, host, key_val)?.is_some())
     }
 
-    /// Extends `key` to live `extend_to` ledgers from now (not counting the
-    /// current ledger) if the current live_until_ledger for the entry is
+    /// Extends `key` to live `extend_by` ledgers from now (not counting the
```
dmkozh, line 452: It's actually been correctly called `extend_to` (albeit it's a bit confusing). `extend_by` would mean that we take current TTL and extend it by a given number. Instead, we're not looking at the current TTL at all, just extend the TTL to be at least `extend_to` ledgers.

jayz22: This is an internal function so I think we should pick a name that best reflect what it actually does.  
All it does is adds the current sequence number by an amount, check some bounds, then set it as the new ttl value. I think the term that better interprets that amount is "extend by", i.e. new_live_until = sequence_number + extend_by.
We are not looking at TTL in this function as you mentioned, so I think making the input parameter the number the TTL is being extended to is pretty confusing.
Or do you think we can rename this function name to be more specific, e.g. `compute_new_live_until_and_set_ttl`.
I think this will help avoiding confusions such as mentioned in the tracking [issue](https://github.com/stellar/rs-soroban-env/issues/1329).

soroban-env-common/env.json
jayz22: fixed.

soroban-env-host/src/storage.rs
dmkozh: But we do specifically extend the TTL, don't we? The equation you provided literally extends the TTL to be `extend_to/by` ledgers from now. I.e. the equation doesn't include the current live_until_ledger. This is also not quite internal, since you've updated the error message to include the new arg name.
We could try renaming this, but I'm not sure how does it address the issue, given that this function is indeed internal. The external interface has already been settled, for better or worse. We also don't really do two things here; TTL is just the *amount* of ledgers between now and live_until_ledger. I think if you have this definition in mind, this function would make more sense. The fact that we set it via setting a new `live_until_ledger` value is just an implementation detail; in an abstract sense this function extends TTL to be at least `extend_to` ledgers.

jayz22: Reverted the renaming change. Don't think this is worth arguing over right now. But I do feel the whole extension ttl, to include or not include current ledger is pretty confusing and footguny (as mentioned in the issue). If you think there are ways to improve it, I would be very glad.

dmkozh: Yeah, I agree this is not ideal, though in the SDK things should be more consistent now, at least when SDK methods are being used. The way the ledger settings are interpreted is kind of confusing, but changing it would be a protocol change, and I'm not sure it's worth the hassle.

dmkozh (approved):

PR #1425, opened 2024-06-24
Relex curve25519-dalek version

jayz22:
### What

curve25519 is also subject to the security advisory and needs to be bumped

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-env-host/Cargo.toml
+++ b/soroban-env-host/Cargo.toml
@@ -62,7 +62,7 @@ sha3 = "=0.10.8"
 # stellar-core ultimately ships. This is a temporary workaround until the next
 # protocol release when we can safely bump the minimum curve25519-dalek version
 # in stellar-core as well.
-curve25519-dalek = { version = ">=4.1.1, <=4.1.2", default-features = false, features = ["digest"]}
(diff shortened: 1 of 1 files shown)


dmkozh (approved):

soroban-env-host/Cargo.toml
```diff
@@ -62,7 +62,7 @@ sha3 = "=0.10.8"
 # stellar-core ultimately ships. This is a temporary workaround until the next
 # protocol release when we can safely bump the minimum curve25519-dalek version
 # in stellar-core as well.
-curve25519-dalek = { version = ">=4.1.1, <=4.1.2", default-features = false, features = ["digest"]}
+curve25519-dalek = { version = ">=4.1.1, <=4.1.3", default-features = false, features = ["digest"]}
```
dmkozh, line 65: Actually, can you just remove the upper boundary? We've also removed it for ed25519; I'm not sure if it provides any real value (especially now that ed25519 version is unrestricted)

jayz22: Ok, we will need to pin it to 4.1.3 after core is updated anyway. Removed the upper bound.

dmkozh (approved):

PR #1440, opened 2024-07-30
Patch XDR for binary fuse filter

SirTyson:


dmkozh (approved):
LGTM, but we can't merge this due to CI issues.

leighmcculloch:
Rust 1.80 just started validating cfg values. That is probably the cause here. Adding the missing feature is probably the quick fix.
- https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#checked-cfg-names-and-values

SirTyson:
Closing, we figured out how to patch XDR dependency with just a top level patch in Core.

PR #1462, opened 2024-09-16
Bump tracy-client to 0.17 to expose delayed-init feature

graydon:
This bumps the tracy-client crate to 0.17, so that we can access the `delayed-init` feature (it's not present in 0.16.4 but necessary to use with core). Previously we were patching this dep entirely with a local vendored build of tracy in core, so this potentially moves us closer to upstream (if it work).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,7 +198,7 @@ dependencies = [
  "iana-time-zone",
  "num-traits",
  "serde",
- "windows-targets",
+ "windows-targets 0.48.5",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


tomerweller (approved):

Cargo.lock
```diff
@@ -198,7 +198,7 @@ dependencies = [
  "iana-time-zone",
  "num-traits",
  "serde",
- "windows-targets",
+ "windows-targets 0.48.5",
```
leighmcculloch, line 201: This seems not great. Would be good for deps if we can resolve in a way to not require multiple windows-targets.

leighmcculloch (approved):

Cargo.lock
graydon: hmm. they're dev-deps of tracy-client-sys when it's being tested under loom. I agree I don't love the duplication in the lockfile but none of this is making it into any actual builds. I've tried updating quite a lot of things-that-cause-the-dependency and ultimately I get blocked on one that's just quite out of date, textplots -> drawille -> colored -> windows-sys 0.48 -> windows-targets 0.48.

PR #1466, opened 2024-09-23
Update dalek crates, k256, and wasmi for protocol 22

### LINKED ISSUE #1313: Evaluate wasmi v0.32
Wasmi recently released [v0.32.0-beta.0](https://github.com/paritytech/wasmi/releases/tag/v0.32.0-beta.0) which contains new major features:
- the new register-machine bytecode based executor
- lazy Wasm function compilation.
- unchecked Wasm compilation.

This issue tracks the experimentation to use the new version in soroban-env, and accessing the performance impacts of various compilation/validation modes.

### LINKED ISSUE #1353: Tracking issue to bump dependent rust crates prior to protocol 22
### What?
Audit our dependent crates to see if we need to bump their versions. Also, compare stellar-core pinned versions with that of rs-soroban-env crates.

### Why?
For example, we have curve25519-dalek crate in env that allows for the version range of 4.1.1 to 4.1.2 while core pins to 4.1.1. Evaluate to see if we can pin env to a specific version and point core to that.
```[tasklist]
### Tasks
- [ ] curve25519-dalek crate v4.1.2
- [ ] k256 v13.3.0
- [ ] Related issue, https://github.com/stellar/rs-soroban-env/issues/1422
- [ ] Another related issue, https://github.com/stellar/stellar-core/issues/4278
```

graydon:
Just update deps for p22

fix #1353 
fix #1313

### DIFF
--- a/soroban-env-host/Cargo.toml
+++ b/soroban-env-host/Cargo.toml
@@ -33,7 +33,7 @@ num-traits = "0.2.17"
 num-integer = "0.1.45"
 num-derive = "0.4.1"
 backtrace = { version = "0.3.69", optional = true }
-k256 = {version = "0.13.1", default-features = false, features = ["ecdsa", "arithmetic"]}
+k256 = {version = "0.13.3", default-features = false, features = ["ecdsa", "arithmetic"]}
 p256 = {version = "0.13.2", default-features = false, features = ["ecdsa", "arithmetic"]}
 ecdsa = {version = "0.16.7", default-features = false}
 sec1 = {version = "0.7.2"}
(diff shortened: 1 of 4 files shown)


anupsdf:
Should we remove dalek from [deny.toml](https://github.com/stellar/rs-soroban-env/blob/main/deny.toml#L57)? @dmkozh might have context on temporarily having there.

soroban-env-host/Cargo.toml
```diff
@@ -46,27 +46,7 @@ getrandom = { version = "0.2.11", features=["js"] }
...
-# range-requirement here. IOW nothing about this range-requirement changes what
-# stellar-core ultimately ships. This is a temporary workaround until the next
-# protocol release when we can safely bump the minimum curve25519-dalek version
-# in stellar-core as well.
-#
-# NB: we relaxed this further in order to comply with the security advisory
-# which requires "4.1.3". We will need to pin to this version once Core is 
-# updated to the new env version.
-curve25519-dalek = { version = ">=4.1.1", default-features = false, features = ["digest"]}
+curve25519-dalek = { version = ">=4.1.3", default-features = false, features = ["digest"]}
```
anupsdf, line 49: I guess stellar-core's curve25519-dalek v4.1.1 will get regenerated in [p22-expect.txt](https://github.com/stellar/stellar-core/blob/master/src/rust/src/dep-trees/p22-expect.txt)?

anupsdf:
Should we ping dalek to a specific version of 4.1.3 instead of `>=`4.1.3 ?

soroban-env-host/Cargo.toml
graydon: yes

PR #1479, opened 2024-10-24
Fill gaps in ref conversions to Val

leighmcculloch:
### What
Fill gaps in ref conversions to Val.

### Why
Needed for:
- https://github.com/stellar/rs-soroban-sdk/pull/1382

### DIFF
--- a/soroban-env-common/src/convert.rs
+++ b/soroban-env-common/src/convert.rs
@@ -86,6 +86,14 @@ impl<E: Env> TryFromVal<E, i64> for Val {
     }
 }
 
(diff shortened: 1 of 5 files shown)


graydon:
Mmm I _think_ this is mostly-ok but I also think that since it's an impl of a public type and a public trait it falls under a major semver change? Maybe?

leighmcculloch:
It does, but that's okay, it's targeting the v22 major release because it's required for the SDK changes being made for constructors.

dmkozh (approved):

PR #1482, opened 2024-10-29
Add a test-only mechanism for metering most of the relevant resources per-invocation

dmkozh:
### What

This is an opt-in feature that will reset the budget and take a storage snapshot for every 'logical' invocation (such as `call` or a lifecycle operation). Then when the invocation is done we use the snapshot and current budget to produce the estimate for the resources consumed by the invocation (budget-related, IO-related and rent bumps).

This also provides a rough estimation for the respective fee breakdown given a fee config.

### Why

Make unit tests more useful for rough performance evaluation. This is the initial env-side implementation for https://github.com/stellar/rs-soroban-sdk/issues/1319

### Known limitations

Some of the resources are tricky to model. Specifically, this omits:

- transaction size
- return value size
- some XDR roundtrips that always happen for production scenarios

These shouldn't be too significant though and are likely better addressed via e2e runs (like simulation or some `e2e_invoke`-based mechanism)

### DIFF
--- a/soroban-env-host/src/budget/dimension.rs
+++ b/soroban-env-host/src/budget/dimension.rs
@@ -131,8 +131,12 @@ impl BudgetDimension {
 
     pub(crate) fn reset(&mut self, limit: u64) {
         self.limit = limit;
-        self.total_count = 0;
         self.shadow_limit = limit;
+        self.reset_count();
+    }
+
+    pub(crate) fn reset_count(&mut self) {
+        self.total_count = 0;
         self.shadow_total_count = 0;
     }
 
--- a/soroban-env-host/src/budget/util.rs
+++ b/soroban-env-host/src/budget/util.rs
@@ -69,6 +69,15 @@ impl Budget {
         Ok(())
     }
 
+    pub fn reset(&self) -> Result<(), HostError> {
+        self.with_mut_budget(|mut b| {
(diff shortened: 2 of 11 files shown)


leighmcculloch (commented):
Will the idea be in the SDK to call `budget.reset()` before each top-level invocation? That sounds like a great idea.

I think we should move to that approach env wide, except for types that are in memory and storage, treat everything else like a reset. One thing we need to start doing that for is logs. Right now you can output diagnostic logs for many invocations, which is quite confusing. In short, yeah, this approach of resetting 👍🏻, and it should just be mandated in the SDK and always happen.

If someone needs to test they can always wrap their calls in a single `.as_contract`.

dmkozh:
> Will the idea be in the SDK to call budget.reset() before each top-level invocation? 

SDK just needs to enable invocation metering, env will take care of the resets.

> I think we should move to that approach env wide, except for types that are in memory and storage, treat everything else like a reset. One thing we need to start doing that for is logs. 

Yeah, we're getting this automatically by clearing the events. I'm resetting the budget, footprint, and events. I think that should be everything, but we can reset more if necessary.

> If someone needs to test calls together they can always wrap their calls in a single .as_contract.

Good point, this actually hasn't been covered by the metering scope.

sisuresh (approved):
Nice work!

PR #1485, opened 2024-11-05
Temporarily disable semver checks.

dmkozh:
### What

Disable semver checks temporarily.

### Why

The check is broken due to `arbitrary` crate update (https://github.com/rust-fuzz/arbitrary/issues/203).

### Known limitations

N/A

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,11 @@ jobs:
     # temporarily disabling publish-dry-run, it appears to be broken with recent
     # changes to dependency specifications?
     # needs: [fmt, cackle, cargo-deny, check-git-rev-deps, build-and-test, publish-dry-run]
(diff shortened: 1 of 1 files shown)


sisuresh (approved):

.github/workflows/rust.yml
```diff
@@ -58,16 +58,19 @@ jobs:
...
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v3
-    - run: rustup update
-    - uses: stellar/binaries@v29
-      with:
-        name: cargo-semver-checks
-        version: 0.35.0
-    - run: cargo semver-checks --exclude soroban-simulation
+  # This check is currently broken due to incompatible `arbitrary` crate
```
leighmcculloch, line 61: Rather than remove the job I think we should remove it from the needs list on the `complete` job. That way we know when it starts passing again but it won't block builds.

dmkozh: Makes sense, done.

PR #1506, opened 2025-01-09
Surface module cache for reuse.

graydon:
The goals of this PR are:

 1. Surface the `ModuleCache` from inside soroban to its embedding environment.
 2. Allow a module cache to _outlive_ the soroban `Host`, and be reused from one `Host` to the next.
 3. Allow the `ModuleCache` to be _pre-populated_ from before the lifetime of a single `Host`, (eg. by stellar-core, with all the non-archived contracts currently live in the ledger, at startup or such.)

These goals are accomplished by the following structural changes:

 1. Factoring out error-handling / error-mapping functionality of `Host` into a separate trait `ErrorHandler` that `Host` implements, and adding a new trait `CompilationContext` that extends both `ErrorHandler + AsBudget`. This abstracts the services the `ModuleCache` _needs_ from the `Host`, such that the embedding environment can provide its own `CompilationContext` in the times pre- or post-`Host` lifetime.
 2. Modifying the internal map that stores `Module`s inside the `ModuleCache` to use a refcounted-shared-mutable `std::map::BtreeMap` that can be incrementally and efficiently added-to or removed-from, and is _not_ metered.
 3. `Module` storage and arguments that were `Rc<>` are made `Arc<>` and the whole `ModuleCache` is made `Send+Sync`. This is mainly because we want to work with the reusable `ModuleCache` across C++ threads, and while Rust has no idea about C++ threads it at least helps us shoot ourselves in the foot less if all the Rust code that works with it preserves the sense of "this would be threadsafe in Rust if Rust was making the threads".
 4. There is also a bit of gratuitous churn where things get renamed from `foo` => `wasmi_foo` and `Foo` => `wasmi::Foo`. This is because this change was done as part of (and there'll be a followup PR including) work equipping the `Vm` and `ModuleCache` with an experimental wasmtime/winch backend. It turns out this PR's work (on just the `ModuleCache` reuse part) is high enough value, and agnostic enough to which backend it's using, that we want to split this work out, move it up in the schedule and explore getting it into as early a release as possible. But I didn't bother reverting-out the `wasmi_`-qualification since it's intertwined with other changes here and harmless to keep.

This PR is a continuation of https://github.com/stellar/rs-soroban-env/pull/1511 and should be merged after it.

This PR implements [CAP-0065](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0065.md)


soroban-env-host/src/vm/module_cache.rs
```diff
@@ -18,20 +20,125 @@ use wasmi::Engine;
...
+// - Adding an existing module is a harmless no-op, not an error.
+// - The linkers are set to "maximal" mode to cover all possible imports.
+// - The cache easily scales to a large number of modules, unlike MeteredOrdMap.
+// - There is no metering of cache map operations.
+// - The cache can be cloned, but the clone is a shallow copy.
+// - The cache is mutable and shared among all copies, using a mutex.
+
+#[derive(Clone)]
+enum ModuleCacheMap {
+    MeteredSingleUseMap(MeteredOrdMap<Hash, Arc<ParsedModule>, Budget>),
```
dmkozh, line 48: If we don't try to preserve p22 compatibility (which we are not going to), then this should not be necessary in production code paths, right? It might make sense for simulation and unit tests though, so maybe it should be gated with testutils/test/recording_mode configurations. The debug budget would also be needed for metering this.

FWIW for simulation it would be nice to also use the reusable cache, but that requires some additional discussion and doesn't belong to this PR.

dmkozh (commented):
Looks good to me overall, only not sure about the gating of the old behavior (left a comment about that).

soroban-env-host/src/e2e_invoke.rs
```diff
@@ -336,6 +336,44 @@ pub fn invoke_host_function_with_trace_hook<T: AsRef<[u8]>, I: ExactSizeIterator
     base_prng_seed: T,
     diagnostic_events: &mut Vec<DiagnosticEvent>,
     trace_hook: Option<TraceHook>,
+) -> Result<InvokeHostFunctionResult, HostError> {
```
jayz22, line 339: Do we still need this version? Seems like it's sufficient to keep the `_and_module_cache` version with the optional module cache.

soroban-env-host/src/vm/module_cache.rs
```diff
@@ -73,43 +180,96 @@ impl ModuleCache {
...
+            curr_ledger_protocol,
+            &contract_id,
+            wasm,
+            VersionedContractCodeCostInputs::V0 {
+                wasm_bytes: wasm.len(),
+            },
+        )
+    }
+
+    pub fn parse_and_cache_module<Ctx: CompilationContext>(
```
jayz22, line 218: Nit: this can be `pub (crate) fn`, so is `add_stored_contracts`

soroban-env-host/src/vm/module_cache.rs
graydon: I'll see if I can remove it -- we still use it in a few code paths like upload where there's no reuse -- but possibly that doesn't need a cache at all, or can ignore the cache for the sake of measurement since it'd only ever be a 1-element cache that's thrown away anyways.

Concerning simulation: this is an interesting point! We're going to greatly overcharge people (and kinda negate the point of the change entirely) if we don't adjust simulation. I think the correct thing to do is populate a 1-entry cache before the simulation call (on a budget with a high-though-not-DoS-able limit), then pass that populated cache into the host being simulated. @dmkozh WDYT?

soroban-env-host/src/vm/module_cache.rs
graydon: No, a major point of this change is that the module cache can now be (publicly) operated by stellar-core before and after a host exists.

soroban-env-host/src/vm.rs
```diff
@@ -234,31 +192,52 @@ impl Vm {
         } else {
             None
         };
+        Ok((store, instance, memory))
+    }
+
+    /// Instantiates a VM given the arguments provided in [`Self::new`],
+    /// or [`Self::new_from_module_cache`]
```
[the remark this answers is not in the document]
graydon: Fixed.

soroban-env-host/src/vm/module_cache.rs
graydon: @dmkozh So .. this seems to more-or-less work as you describe -- I can mostly rip-out all the weird immutable 1-tx-cache logic as well as all the ghastly emulation of its lifecycle we had to put in for recording mode -- but there is a small wrinkle: observations change. For two reasons:

 1. we're not charging the (small) costs of `MeteredOrdMap` operaions on the cache anymore at all
 2. we're no longer able to distinguish, in unit tests that use recording mode, between:
     - "upload and invoke" cases which we _should_ charge for and
     - "invoke something that would have been in the cache", which we _want_ to suppress and direct the charge to the shadow budget for

We cannot distinguish case 2 -- again only in unit tests -- because the distinguishing happens in frame::instantiate_vm and works by looking at "stuff that was in the storage snapshot", and in unit tests the storage snapshot is always empty. IOW unit tests actually _are_ usually more like "upload and invoke" transactions; we've just been (somewhat inaccurately) observing some of them, in recording mode, as though they're recording things-that-were-just-uploaded. I attempted to compensate for this by adding a new call to the testutils routine that registers a contract, to automatically build a module cache that contains them, so we get module cache _hits_ in that case, but I'm not sure that's a good idea either -- recording mode in tests is a weird case where it's not really trying to be "like simulation" nor is it trying to be "like real apply", it's just a sort of convenience for tests to not need footprints. Idk, open to suggestions here!

I've uploaded a version of this that has these changes. Will keep poking at them tomorrow.

soroban-env-host/src/e2e_invoke.rs
graydon: I thought I'd keep it to avoid gratuitous breakage of clients (it is after all the API most are likely to call). I can remove it if you strongly prefer. @dmkozh this is kinda "your API", do you have a preference?

soroban-env-host/src/vm/module_cache.rs
dmkozh: I'll take a look at the code, and in any case maybe it would make more sense to leave the testing/simulation improvements for the followups. That said, here is how I conceptually see the budget accounting for tests:
- A new test host is instantiated together with an empty mutable module cache
- Any wasm upload (be it a test utility, or a host fn call) charges the Wasm parse and puts the module into cache
- Any future call to the cached contract within this host reads the contract from the cache
It seems like this is as close as it gets to the on-chain behavior, unless I'm missing something. Basically, if ledger snapshots are not involved, then every test builds the storage from scratch anyways, so scenario of 'invoking something that would have been in the cache' doesn't seem possible to me at all.

> recording mode in tests is a weird case where it's not really trying to be "like simulation" nor is it trying to be "like real apply", it's just a sort of convenience for tests to not need footprints

FWIW recording mode is a de-facto standard for the unit tests (I'm not even sure if the SDK provides a way to specify the footprint - it really wouldn't cover anything interesting about the contract logic). And we also want the metering to be sane. For sure, we'll miss XDR parsing costs here and there, and testing with a compiled-in contract is less accurate than testing with Wasms. However, I think we should strive to keep the main cost components correct (like charging/not charging for Wasm parsing, given a Wasm contract).

soroban-env-host/src/e2e_invoke.rs
dmkozh: AFAIK only Core uses this as of today. Clippy won't allow us to break this though, unless we bump the version or use unstable-next-api feature. I think breaking this in p23 is very low risk (if someone has been using it, they must be advanced enough to migrate to the module cache interface as well). But we don't have to do that in this PR.

jayz22: Ok makes sense, I don't feel strongly so feel free to keep it for backward compatibility.

soroban-env-host/src/vm/module_cache.rs
graydon: @dmkozh Ok, I've done this now (and fixed all the test fallout). One interesting thing is I had to bump up the margin of error on simulations fairly significantly. I _think_ this is mostly due to the dominant-shared-cost of parsing modules going away. But I'm not 100% certain.

soroban-env-host/src/vm.rs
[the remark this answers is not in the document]
graydon: fixed / made-obsolete

soroban-env-host/observations/23/test__address__invalid_strkey_to_address_conversion.json
dmkozh, line 1: Shall we generate the p23 observations in a separate PR in order to isolate the diffs caused by the logic changes in this PR?

dmkozh: I mean create a separate PR that just bumps the protocol version and updates the observations, merge it, then rebase this PR. I think this is how we should normally do the env protocol version bumps, unless there is some issue with that. I expect this to limit the scope of the observation changes significantly.

graydon: ok

graydon: @dmkozh posted https://github.com/stellar/rs-soroban-env/pull/1511

graydon: Updated so now the residual reobservation churn of _this_ PR is a3882e9ef17f414e8e1634231d65c86c500acb06 which should mostly only have CPU/memory cost changes.

dmkozh (approved):

PR #1512, opened 2025-01-23
Tighten the budget estimations in e2e tests.

dmkozh:
### What

Tighten the budget estimations in e2e tests.

There was a small bug in the module cache construction - we didn't consider the proper version of the code entry when building the module cache to pass to the enforcing mode. This fix allows us to bring the discrepancy between the recording and enforcing modes back to 0.02.

### Why

Test fixes

### Known limitations

N/A


graydon (approved):
Aha! Nice catch. Thanks for tracking this down.

PR #1517, opened 2025-02-14
Implement CAP-68 - host function to get the executable given an address.

dmkozh:
### What

Implement [CAP-68](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0068.md) - host function to get the executable given an address.

### Why

See CAP-68 for some motivation

### Known limitations

N/A


soroban-env-common/env.json
```diff
@@ -2403,6 +2403,19 @@
...
+                    "export": "4",
+                    "name": "get_address_executable",
+                    "args": [
+                        {
+                            "name": "address",
+                            "type": "AddressObject"
+                        }
+                    ],
+                    "return": "Val",
+                    "docs": "Returns the executable corresponding to the provided address. When the address does not exist on-chain, returns `Void` value. When it does exist, returns a value of `AddressExecutable` contract type. It is an enum with `Wasm` value and the corresponding Wasm hash for the Wasm contracts, `StellarAsset` value for Stellar Asset contract instances, and `Account` value for the 'classic' (G-) accounts.",
```
leighmcculloch, line 2417: "StellarAsset" feels pretty odd given that we don't have StellarAccount for the second item. Is there a reason we can't just call the asset contract the asset contract?
```suggestion
                    "docs": "Returns the executable corresponding to the provided address. When the address does not exist on-chain, returns `Void` value. When it does exist, returns a value of `AddressExecutable` contract type. It is an enum with `Wasm` value and the corresponding Wasm hash for the Wasm contracts, `Asset` value for Stellar Asset contract instances, and `Account` value for the 'classic' (G-) accounts.",
```

dmkozh: I'm not really comfortable changing this, give that `ContractExecutable` enum that already exists in the protocol uses `StellarAsset` (both in the XDR and in the host). Also we refer to SAC as 'Stellar Asset Contract' everywhere, so I'm not sure if just 'Asset' is sufficient to disambiguate (but the main reason is really consistency).

leighmcculloch: That's fair enough. The only other place in the API would be in the auth framework API, is that right? Seems like a pretty small surface area.

dmkozh: Well, it appears in the XDR as well (like `ScContractInstance` etc.). I'm not sure it's worth having divergent terminology in different places, even though the current name might be a bit unfortunate (though as I've mentioned above, I don't think it's too bad, given that SAC is a well-known term).

graydon, line 2417: The env interface doesn't define a mapping of env types to SDK `#[contracttype]` types. So discussing a "contract type" here, or an enum, seems a little like mixing layers of abstraction. Though I see that it's already being done in one of the other comments and this is not a technical blocker, it does concern me a bit. I wonder if it'd be any better to try to be more explicit about the type here? Or I guess since we generate a corresponding type using the env macros, whether we ought to at some point spell out the meaning of those Rust language env-macro types in terms of env-interface types (Val and host objects and such)?

graydon (approved):
Looks generally ok. One comment that I'm not sure whether you want to try to address here or not, we could leave it for some later work or just ignore.

I also don't know why there's a label on this bug to ignore the env interface bump? Every new host function is a protocol change.

soroban-env-common/env.json
dmkozh: > So discussing a "contract type" here, or an enum, seems a little like mixing layers of abstraction.

Yeah, I agree this is underspecified/not quite the right abstraction (now), but yes, we're already doing that because coming up with a new host object type for every rather small use case is not really sustainable. 

I think this is probably worthy of a documentation improvement pass, but I don't think it should be blocking this particular PR:
- Soroban UDTs should be specified explicitly, i.e. specify how exactly structs/enums map to the `Val`s. While these are inspired by Rust data types, I don't think the spec has to be tied to Rust necessarily. That would be generally beneficial, as even if one were to build a non-Rust SDK, they would probably want to interact with the existing contracts that use the Rust UDT format.
- Then we could expose the UDT itself (e.g. in JSON) and refer to the UDT definition in the docs. I don't think we need anything too fancy here - just a Rust struct + a reference to the exact UDT encoding spec is way more readable than low-level 'raw' spec.

'Quicker' solution is to list the exact contents of `Val` for every case. It's probably not too bad for this particular function (though still way too verbose IMHO), but is still much worse for auth (especially `authorize_as_curr_contract`), as it has to use a recursive data structure.

dmkozh: I'll mark this resolved for the time being, since this is just a docs issue it should be easy to followup.

leighmcculloch: > coming up with a new host object type for every rather small use case is not really sustainable.

This surprises me. I thought we designed the Val type so that it was specifically extendable for a not-insignificant number of types, so we could do just this.

It looks like we're taking a shortcut. UDTs were never intended to be used by the host, and are completely outside the protocol.

What makes adding new Val types unsustainable?

> * Then we could expose the UDT itself (e.g. in JSON) and refer to the UDT definition in the docs. I don't think we need anything too fancy here - just a Rust struct + a reference to the exact UDT

Contracts already encode a descriptions of UDTs in XDR, and there already exists an XDR-JSON schema that dev tooling is using for the XDR (see the `stellar contract info interface --format json` command that uses it). There's no SEP for this yet, but it's in my queue. There's a new SEP proposed for the meta side here, and the spec side will take similar shape: https://github.com/stellar/stellar-protocol/discussions/1656.

graydon: > It looks like we're taking a shortcut. UDTs were not intended to be used by the host, and are completely outside the protocol.

Somewhat, yeah. But also somewhat not. "The protocol" has a somewhat blurrier boundary in Soroban -- insofar as "contracts delegate logic to users" but we also ship built-in contracts that implement the same interface that users are expected to use among themselves: for example all the SAC types are specified as `contracttype`, not specialized XDR, and I think we'd consider changes to those types as "protocol changes" even if users didn't change their corresponding UDTs (and like .. old contracts will of course keep using any old UDTs). We also use `contracttype` for the structured auth type `InvokerContractAuthEntry` that is used directly by the `authorize_as_curr_contract` host function call.

I think it's plausible / a good idea to formalize the role of host-side `contracttype` UDTs as "part of the protocol"; but I just wanted to point out that we haven't really done that yet, we've just kinda drifted into using the concept in the env.json file without pinning down what it means in a CAP or whatever.

I think there are 2 rationales for this, though I might be missing some, please @dmkozh correct me if I'm missing anything:

1. (Minor motive): It's less work and quicker to iterate on `contracttype`-defined types. Defining something as XDR-and-host-objects is a fairly heavyweight process and touches "all code that touches SCVal/Val", whereas changes to a `contracttype` macro are more localized to the two sides of its use, guest code and "some specific subsystem of the host". I think that to some extent the availability of the macro allowed rapid iteration on both the SAC and auth subsystems of the host (which are two of the larger chunks of code in the host, but also operate and were developed somewhat separately / in parallel to other host logic).
2. (Major motive): Guest / contract code _does_ need to work with these types explicitly -- construct, deconstruct, access, modify -- and if we were to define the types as XDR-and-host-objects they would be _opaque_ to guests: we'd have to make a whole pile of separate host functions to pattern-match each variant and/or access each field of the structures (or we'd be forcing users to serialize and decode XDR in their contracts). UDT definitions are much more convenient for users to work with, so "types that users need to work with" are obvious candidates for UDT-ness.

To some extent all of this reprises the question of whether we have UDTs _at all_! I was initially quite opposed to them (even in guest code) because they produce a bunch of excess code in the guest / do eager decoding of all fields rather than point-access of only a few. But the argument for convenience won out.

leighmcculloch: Thanks, I can buy into using the format more over adding new Vals.

I still think documenting the relationship is technically a SEP rather than a CAP, and I've opened this issue to track the writing of a SEP that captures the contract spec:
- https://github.com/stellar/stellar-protocol/issues/1662

dmkozh: > This surprises me. I thought we designed the Val type so that it was specifically extendable for a not-insignificant number of types, so we could do just this.

Graydon has provided a great summary on this and I don't have too much to add. Even adding rather simple types is not that trivial (see CAP-69). Every new type requires the XDR representation, a rather significant number of host function that allow users to do anything useful to that type and proper downstream support for rendering and building these types. Every step gets significantly more complicated with more complex types.

UDTs on the other hand have all the necessary utilities already built for them and there is no need for the XDR changes.

> It looks like we're taking a shortcut. UDTs were not intended to be used by the host, and are completely outside the protocol.

Yes, it's a shortcut. I don't see an issue with that though. They're already inside the protocol in various ways, like Graydon has pointed out. I don't really see any downside to standardizing the UDT representation used by host in a more strict fashion.

PR #1520, opened 2025-02-20
Small optimization for Symbol.

dmkozh:
### What

Use `Symbol::try_from_val` in `map_new_from_linear_memory` which both ensures the proper symbol validation (which doesn't happen now) and makes the conversion slightly cheaper.

This is a non-functional protocol change: while it has to go into env for the new protocol, the only observable change is the slight reduction of CPU costs.

### Why

Implementation cleanup/optimization

### Known limitations

N/A


leighmcculloch (commented):
Previous attempts to optimise Symbol code have caused significant changes guest side that was unintuitive and unexpected.

Does this function ever get triggered contract side? If yes, what impact does this have on WASM size and instructions this translates into on the contract side?

Although, I expect and hope this function doesn't get called contract side because doing so would embed Symbol encoding logic into contracts.

On the surface this looks fine to me.

dmkozh:
> Previous attempts to optimise Symbol code have caused significant changes guest side that was unintuitive and unexpected.

FWIW the main 'optimization' here is rather about the maintenance cost and reducing the bug probability (IIRC there were some validation inconsistencies caused by the 'fallback' conversion). But I think it's not that important today. I've made a quick check and it seems like one SDK contract had a bit of size increase. I don't think it's worth spending more time debugging this, so I just rolled back the change for the time being. The remaining change is env-only.

PR #1522, opened 2025-03-07
Implement fee-related changes of CAP-66.

dmkozh:
### What

Implement fee-related changes of [CAP-66](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0066.md).

This includes the following changes:
- use the disk read bytes in place of overall read bytes, i.e. only account for the classic and restored entries (the latter is no a part of this PR)
- use the module memory size for the contract code entries instead of their XDR size for rent computations
- rename the fee-related fields appropriately

This also temporarily switches env to vnext XDR (until we merge next changes into curr).

### Why

Support CAP-66

### Known limitations

N/A


graydon (approved):
I tried a few variants of "record the cost in the budget and read it back" and it was sufficiently clunky that I agree for now it is probably best to just accept the code duplication here.

PR #1546, opened 2025-05-13
Remove enabling of base64 feature in stellar-xdr dep

leighmcculloch:
### What
  Remove enabling of base64 feature from in stellar-xdr dep.

  ### Why
  I can't see a use for it in the crates outside tests.

Looks like it was added in https://github.com/stellar/rs-soroban-env/pull/375 to make it more convenient to import the env and have base64 already enabled in the xdr lib also imported, which I approved. However, downstream deps should be able to import the stellar-xdr crate themselves at the same version and enable the base64 feature in the same way they do for other features like the serde feature.

### Why not

If core is also using the base64 feature then there's not really any value removing it here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,12 +224,6 @@ checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
 name = "base64"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


leighmcculloch:
> If core is also using the base64 feature then there's not really any value removing it here.

It does appear that core imports the base64 package directly too (see below), so this PR doesn't add much value.

```toml
[[package]]
name = "stellar-core"
version = "0.1.0"
dependencies = [
 "base64",
```

Ref: https://github.com/stellar/stellar-core/blob/475f1448074ad5b45f63ff0068121e798e4b1b67/Cargo.lock#L524-L528

PR #1555, opened 2025-05-18
Rename Host::check_same_env to Host::is_same

### LINKED ISSUE #1554: Change `check_same_env` to `is_same_env`
### What
Change `Host` `check_same_env -> Result` to `Host` `is_same_env -> bool`.

### Why
The check function is used only in the SDK, not in the Env. In the SDK it is used in two ways:
1. To hard stop in tests when a value is used with the wrong Host where that use is unsupported.
2. To change the behaviour used when comparing vals across Envs which is the only cross-Env supported action.

In both cases the SDK is using the `Host` `check_same_env` to find out if the env is the same, and then the SDK is choosing to present that information in different ways either by panicking or branching.

Given that the SDK owns the presentation, and the Env is just providing information, the Env should provide this information plainly as a bool. The SDK can present the error however it needs.

This oddity was pointed out here:

> It also doesn't really make sense for `check` to be used in `is` and not vice versa. It would probably make more sense to expose the `is` function from env if we think it's necessary.

_Originally posted by @dmkozh in https://github.com/stellar/rs-soroban-sdk/pull/1467#discussion_r2094183786_

leighmcculloch:
### What
  Rename the host comparison method from check_same_env to is_same, changing it to return a boolean instead of a Result.

  ### Why
  The new name better reflects the method's purpose of being used by the SDK to compare two Hosts and determine if they in fact are the same underlying host implementation. The check function returns a result with the host assuming that there is in fact an error, whilst the is function returns a bool, merely providing information to the SDK so that the SDK can then decide in the context of its use whether there is an error or a change in behaviour is required.

Close #1554 

Related:
- https://github.com/stellar/rs-soroban-sdk/pull/1467#discussion_r2094183786

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -744,7 +744,8 @@ impl Host {
         self.create_contract_internal(Some(deployer), args, constructor_args_vec)
     }
 
-    /// Returns true if the Host contains the same instance of HostImpl.
+    /// Returns true if the Host contains the same instance of HostImpl and therefore changes to
+    /// one will be observable via the other. If true, both are essentially the same Host.
     pub fn is_same(&self, other: &Self) -> bool {
         Rc::ptr_eq(&self.0, &other.0)
     }
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -744,17 +744,9 @@ impl Host {
         self.create_contract_internal(Some(deployer), args, constructor_args_vec)
     }
 
-    pub fn check_same_env(&self, other: &Self) -> Result<(), HostError> {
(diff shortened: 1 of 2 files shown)


soroban-env-host/src/host.rs
```diff
@@ -744,17 +744,10 @@ impl Host {
         self.create_contract_internal(Some(deployer), args, constructor_args_vec)
     }
 
-    pub fn check_same_env(&self, other: &Self) -> Result<(), HostError> {
```
anupsdf, line 747: Should we maintain backward compatibility to not break contracts that call this old host function?

anupsdf: Nevermind. I don't see this exposed in env.json so it won't impact any contract.

Issue #1288, opened 2023-12-03
Allow users to request more events in debuginfo dump


graydon:
Currently the `DebugInfo::write_events` method stops after the most recent 25 events. This is hardcoded and possibly not ideal for users, some will want more, but it limits output when a contract has gone awry and produced thousands of events. We should provide a way to adjust it.

Note this only has to do with native testing mode, and only with debugging turned on, and can be changed without perturbing anything protocol-level. It's relatively low priority, just to help with debugging.

silence48:
will this ever get updated? i'm one of those users wanting more.

silence48:
for others who stumble across this.  The reason i wanted the limit higher was to log everything using diagnostic events during cargo tests... however during the tests, you can use println in your contract by doing an extern import of std... just need to remove it before you build, but it can allow better logging ...

anyways,  it would still be nice to be able to configure it because why not... suggested fix make it check if an env var is set for the max events, if its not set, have it use the limit of 25. otherwise use the variable

leighmcculloch:
@graydon @dmkozh I think we could at least set this number to be really high for SDK test util usage, whilst core usage could keep the level at 25. They don't need to be configured the same.

silence48:
> [@silence48](https://github.com/silence48), in the meantime, I would suggest using a custom formatting utility - you should have the access to the events and they have `core::fmt::Display` implemented.

Absolutely, I figured it out; I appreciate you looking into it, It took me a minute to figure out what to do as the other contracts I was calling were emitting beyond 25 events before it ever got to mine... but with a little investigation i was able to find they could be dump to stdout... so while i still would like to see a higher limit, it did force me to learn something 🚀

PR #1561, opened 2025-05-22
Remove the limit from the event debug info implementation.

dmkozh:
### What

Remove the limit from the event debug info implementation.

Resolves https://github.com/stellar/rs-soroban-env/issues/1288

### Why

It doesn't seem like this provides any tangible safety benefits, as events are already charged to budget (or shadow budget) and Debug is only used for tests or verbose logging in Core. If a consumer is concerned about the amount of data output, it would make more sense to limit to overall size of the output buffer.

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/host/error.rs
+++ b/soroban-env-host/src/host/error.rs
@@ -38,25 +38,14 @@ impl Into<Error> for HostError {
 
 impl DebugInfo {
     fn write_events(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


graydon (approved):
Oh, yeah, sure. I guess, if it's annoying people in practice more than the existing behaviour.

silence48:
> Oh, yeah, sure. I guess, if it's annoying people in practice more than the existing behaviour.
Just a minor inconvenience but thank you!

PR #1567, opened 2025-06-10
Support simulating restore with entries with no live until (backport to v22)

leighmcculloch:
### What
  Support in simulation a restore operation with entries that do not have a live until value.

  ### Why
  Accommodating snapshot sources that no longer track a live until for an archived entry. Stellar-rpc won't provide a live until for archived entries because stellar-core's getledgerentries endpoint also won't provide a live until for archived entries.
  
Related:
- https://github.com/stellar/rs-soroban-env/pull/1565

Unblocks:
- https://github.com/stellar/stellar-rpc/pull/435


leighmcculloch:
Closing in favour of:
- https://github.com/stellar/stellar-core/issues/4766

Issue #1510, opened 2025-01-16
error messages for incorrectly sorted maps are misleading / confusing


leighmcculloch:
### What version are you using?

22.0.0

### What did you do?

Simulate or submit the following tx, that has map fields out of order:

> AAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGQAAAAAAAAAAQAAAAAAAAAAAAAAAQAAAAAAAAAYAAAAAAAAAAGdTgdKCaqpjHKj82w07K3eFBGfoi5WwVXPHFK0XTX7GgAAAAVoZWxsbwAAAAAAAAEAAAARAAAAAQAAAAIAAAAPAAAAAWIAAAAAAAAOAAAAAmJiAAAAAAAPAAAAAWEAAAAAAAAOAAAAAmFhAAAAAAAAAAAAAAAAAAA=

### What did you expect to see?

Helpful diagnostic errors that indicate that an input map was not sorted correctly, the order of the fields, and the expected order of the fields.

### What did you see instead?

```
❌ error: transaction simulation failed: HostError: Error(Object, InvalidInput)

Event log (newest first):
   0: [Diagnostic Event] topics:[error, Error(Object, InvalidInput)], data:"failed to convert ScVal to host value"
```

PR #1583, opened 2025-08-15
Add error message for converting non-sorted ScMaps to Val.

dmkozh:
### What

Add error message for converting non-sorted ScMaps to Val.

### Why

Fixes https://github.com/stellar/rs-soroban-env/issues/1510

### Known limitations

N/A

### DIFF
--- a/soroban-env-host/src/host/conversion.rs
+++ b/soroban-env-host/src/host/conversion.rs
@@ -485,7 +485,9 @@ impl Host {
                     let v = self.to_host_val(&pair.val)?;
                     mm.push((k, v))
                 }
-                Ok(self.add_host_object(HostMap::from_map(mm, self)?)?.into())
+                Ok(self
+                    .add_host_object(HostMap::from_map_with_host(mm, self)?)?
+                    .into())
             }
             ScVal::Vec(None) => Err(self.err(
                 ScErrorType::Value,
--- a/soroban-env-host/src/host/metered_map.rs
+++ b/soroban-env-host/src/host/metered_map.rs
@@ -327,8 +327,8 @@ where
             host.with_debug_mode(|| {
                 // Currently we only return Object, InvalidInput error from
                 // `from_map` when the input vector is not sorted by key.
-                if !err.error.is_type(ScErrorType::Object)
-                    || !err.error.is_code(ScErrorCode::InvalidInput)
+                if err.error.is_type(ScErrorType::Object)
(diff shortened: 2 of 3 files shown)


soroban-env-host/src/host/conversion.rs
```diff
@@ -485,7 +485,27 @@ impl Host {
                     let v = self.to_host_val(&pair.val)?;
                     mm.push((k, v))
                 }
-                Ok(self.add_host_object(HostMap::from_map(mm, self)?)?.into())
+                Ok(self
+                    .add_host_object(HostMap::from_map(mm, self).map_err(|err| {
+                        let mut err = err;
+                        self.with_debug_mode(|| {
+                            if !err.error.is_type(ScErrorType::Object)
+                                || !err.error.is_code(ScErrorCode::InvalidInput)
```
graydon, line 493: I'm not sure I understand why these these conditions are necessary or sufficient to know "it's unsorted"

dmkozh: That's how from_map is implemented. I realize that's a bit brittle, so I moved the error decoration code to metered_map.rs, so that at least it lives near the code that produces the error. I've also noticed that we create maps in two flows actually, so the helper function is useful at least for deduplication.

soroban-env-host/src/host/metered_map.rs
```diff
@@ -315,6 +315,37 @@ where
...
+    Host: AsBudget + Compare<K, Error = HostError>,
+{
+    pub(crate) fn from_map_with_host(map: Vec<(K, V)>, host: &Host) -> Result<Self, HostError> {
+        Self::from_map(map, host).map_err(|err| {
+            let mut err = err;
+            host.with_debug_mode(|| {
+                // Currently we only return Object, InvalidInput error from
+                // `from_map` when the input vector is not sorted by key.
+                if !err.error.is_type(ScErrorType::Object)
+                    || !err.error.is_code(ScErrorCode::InvalidInput)
```
graydon, line 331: Ok but .. this is checking that the error type is _not_ `(Object, InvalidInput)`

dmkozh: That's embarrassing, thanks for catching this. It was probably an unfortunate copy-paste from another similar function.

graydon (approved):
Phew! Thought I was misunderstanding. LGTM!

PR #1595, opened 2025-09-13
Added batch transfer test wasm function

SirTyson:
Adds new test wasm function to initiate many c address -> c address SAC transfers from a single contract call for performance testing.


soroban-test-wasms/wasm-workspace/contract_sac_transfer/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![no_std]
 
-use soroban_sdk::{contract, contractimpl, Env, symbol_short, IntoVal, Address,
+use soroban_sdk::{contract, contractimpl, Env, symbol_short, IntoVal, Address, Vec,
```
dmkozh, line 3: We shouldn't ever update older protocol Wasms as new SDK changes their compatibility. We also don't need p21/22 ones (and it's not like they're correct here). So please only update p23 wasm.

soroban-env-host/Cargo.toml
```diff
@@ -70,7 +70,7 @@ wasmprinter = "0.2.72"
 expect-test = "1.4.1"
 more-asserts = "0.3.1"
 pretty_assertions = "1.4.0"
-backtrace = "0.3.69"
+backtrace = "0.3.75"
```
dmkozh, line 73: Is this change necessary? Note, that we'll likely not pick this up in core until the next protocol release (unless there is some breaking issue, in which case we'll need an env release).

dmkozh (approved):
We normally keep the wasms frozen after a protocol release, so that the observations don't ever change. But I think in this case it's very clear that the diff is caused just by the wasm update, and just 1 test is affected, so I think this is fine

soroban-env-host/Cargo.toml
SirTyson: This was to get the CI to pass. cargo-deny was complaining about a deprecated transitive dependency, and bumping the backtrace version made the CI happy.

.github/workflows/rust.yml
```diff
@@ -58,15 +58,12 @@ jobs:
...
+  semver-checks:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - uses: stellar/binaries@v40
-      with:
-        name: cargo-semver-checks
-        version: 0.42.0
+    - run: cargo install cargo-semver-checks --locked
```
dmkozh, line 66: I don't think we should do this change. If there is a semver update necessary, we should update it in the binaries repo for everyone, like here: https://github.com/stellar/binaries/pull/43

soroban-env-host/Cargo.toml
dmkozh: FWIW we don't have to update the crates and we can relax the cargo-deny check if necessary. But this update seems harmless enough to me.
cc @graydon to confirm

PR #1607, opened 2025-10-08
Implement CAP-71 - built-in auth delegation mechanism.

dmkozh:
### What

Implement [CAP-71](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0071.md) - built-in auth delegation mechanism.

This adds 2 new host functions: `delegate_account_auth` for protocol-supported auth delegation from the current `__check_auth` invocation to another Address's auth verification procedure, and `get_delegated_signers_for_current_auth_check` that returns the list of all the built-in delegated signers present in the current context.

### Why

Better support for the modular custom account and foundation for G-account auth customization work. See [CAP-71](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0071.md) for more detailed motivation

### Known limitations

N/A


soroban-env-host/src/auth.rs
```diff
@@ -817,8 +987,122 @@ impl AuthorizationManager {
...
+        match sc_addr {
+            ScAddress::Account(acc) => {
+                check_account_authentication(host, acc, signature_payload, signature)?
+            }
+            ScAddress::Contract(acc_contract) => host
+                .call_n_internal(
+                    &acc_contract,
+                    ACCOUNT_CONTRACT_CHECK_AUTH_FN_NAME.try_into_val(host)?,
+                    &[signature_payload.into(), signature, context],
+                    CallParams::default_internal_call(),
```
sisuresh, line 1061: From copilot - 

> delegated contract signers are invoked with CallParams::default_internal_call(), which prohibits reentry. Normal custom-account __check_auth uses check_account_contract_auth with
>   ContractReentryMode::SelfAllowed, so a custom account that works as a regular signer can fail when used as a CAP-0071 delegate. The delegated path should likely call check_account_contract_auth too.

soroban-env-host/src/e2e_invoke.rs
```diff
@@ -594,14 +594,13 @@ impl RecordedAuthPayload {
 #[cfg(any(test, feature = "recording_mode"))]
 fn clear_signature(auth_entry: &mut SorobanAuthorizationEntry) {
     match &mut auth_entry.credentials {
-        SorobanCredentials::Address(address_creds) => {
+        SorobanCredentials::Address(address_creds)
+        | SorobanCredentials::AddressV2(address_creds) => {
```
sisuresh, line 598: Does `into_auth_entry_with_emulated_signature` need to be updated to return the new credential type?

soroban-env-host/src/auth.rs
dmkozh: No, we don't want re-entry here - it makes no sense for account to delegate auth to itself. `a custom account that works as a regular signer can fail when used as a CAP-0071 delegate.` makes no sense to me.

soroban-env-host/src/e2e_invoke.rs
dmkozh: Good catch, yeah, we want to encourage the new type when possible.

soroban-env-host/src/auth.rs
```diff
@@ -435,6 +454,155 @@ struct InvocationTracker {
...
+    root_snapshot: DelegatedAccountAuthSignerTreeNodeSnapshot,
+    match_stack: Vec<usize>,
+}
+
+/// Tracks the state of authentication delegation while processing
+/// authentication logic for an account contract.
+/// Delegation mechanism allows an account to forward its signature payload
+/// and authorization context to the authentication routine of another
+/// `Address` (i.e. `__check_auth` for a contract, or the builtin-in
+/// authentcation for G-addresses).
```
sisuresh: Nit: authentcation -> authentication

soroban-env-host/src/host/frame.rs
```diff
@@ -1247,4 +1248,43 @@ impl Host {
...
+
+    pub(crate) fn ensure_check_auth_frame(&self, context_host_fn: &str) -> Result<(), HostError> {
+        let fn_name = self.with_current_frame(|f| {
+            let fn_name = match f {
+                Frame::ContractVM { fn_name, .. } => fn_name,
+                Frame::HostFunction(_) => {
+                    return Err(self.err(
+                        ScErrorType::Context,
+                        ScErrorCode::InternalError,
+                        "ensure_auth_check_frame is not supported for host fns",
```
sisuresh: Nit: ensure_auth_check_frame -> ensure_check_auth_frame.

Also applies below.

soroban-env-host/src/auth.rs
```diff
@@ -435,6 +454,155 @@ struct InvocationTracker {
...
+
+impl std::hash::Hash for AccountAuthorizationCredentials {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            Self::SourceAccount => 0u32.hash(state),
+            Self::Address => 1u32.hash(state),
+            Self::AddressV2 => 2u32.hash(state),
+            Self::AddressWithDelegates(delegated_auth_tracker_cell) => {
+                3u32.hash(state);
+                if let Ok(delegated_auth_tracker) = delegated_auth_tracker_cell.try_borrow() {
```
sisuresh, line 574: Is it alright to silently skip this if the borrow fails? I know this isn't supposed to happen, but was curious about this.

soroban-env-host/src/auth.rs
dmkozh: Fixed

soroban-env-host/src/auth.rs
dmkozh: Yeah, I think it's the best we can do. Hash is only used for tracing, so it's not critical if this fails. OTOH if we panic here, then we would introduce a divergence on a node that has tracing enabled (on an off chance this ever triggers, that is). I've added a debug_assert though, I don't think we should be running tracing with debug builds, so crashing seems fine.

soroban-env-host/src/host/frame.rs
dmkozh: Fixed

sisuresh (approved):

Issue #1586, opened 2025-08-18
Host function support for field-based hash functions (e.g., Poseidon)


jayz22:
### What
Add host-level support for field-based hash functions (e.g., Poseidon). 

### Why

Many privacy applications (e.g. [privacy pools](https://docs.privacypools.com/)) maintain states (e.g. coin-commitments and nullifiers) in Merkle trees and must generate SNARK proofs of membership/updates. Using bit-based hashes (SHA-2/Keccak) inside circuits inflates constraints, resulting in more complex circuit and longer proving time.

Field arithmetic is computationally heavy and typically not suitable to be implemented in the guest contract. 

Although isn't a protocol requirement, supporting field-based hash can lower the barrier for new and migrating existing apps, by reduce off-chain circuit complexity and proof time.

PR #1608, opened 2025-10-08
Host function for Poseidon/Poseidon2 crypto primitives

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/1586

This PR implements two host functions as specified in [CAP-75](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0075.md): 
- poseidon_permutation
- poseidon2_permutation

Both functions can be initialized with either `BN254` or `BLS12_381` scalar field (both we already support).

The only new cost types added are additional scalar operations required by BN254 field (which are generic and reusable by future protocols), no poseidon specific cost types are added. Their calibrated cost numbers are also included in this PR. 

### Choice of library
Internally it uses the [poseidon2](https://github.com/HorizenLabs/poseidon2) library provided by HorizenLabs ([crate](https://crates.io/crates/zkhash), this is the reference implementation listed by [Ethereum Foundation Poseidon Group](https://www.poseidon-initiative.info/)).  The code is cloned into`src/crypto/poseidon`, with adaptation and tightening (add metering, remove panics), but no logic change. 

### Testing
Test vectors for bn254 and bls12-381 from the original repo were cloned and included in our test suite. 

The [reference test vectors](https://extgit.isec.tugraz.at/krypto/hadeshash) provided by the original Poseidon are also included in our test suite. 

### Why

See  [CAP-75](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0075.md) and its [discussion](https://github.com/orgs/stellar/discussions/1780) thread


soroban-env-host/src/test/poseidon/poseidon2_instance_bls12.rs
```diff
@@ -0,0 +1,1532 @@
...
+    pub static ref MAT_INTERNAL2: Vec<Vec<Scalar>> = vec![
+    vec![from_hex("0x0000000000000000000000000000000000000000000000000000000000000002"),
+    from_hex("0x0000000000000000000000000000000000000000000000000000000000000001"),
+    ],
+    vec![from_hex("0x0000000000000000000000000000000000000000000000000000000000000001"),
+    from_hex("0x0000000000000000000000000000000000000000000000000000000000000003"),
+    ],
+    ];
+
+    pub static ref RC2: Vec<Vec<Scalar>> = vec![
```
dmkozh: Do we have any references to the source of these test vectors? It would be nice to add the references where applicable.

soroban-env-host/src/crypto/poseidon/poseidon_params.rs
```diff
@@ -0,0 +1,123 @@
...
+        d: u32,
+        rounds_f: u32,
+        rounds_p: u32,
+        mds: Vec<Vec<S>>,
+        round_constants: Vec<Vec<S>>,
+    ) -> Result<Self, HostError> {
+        if !SUPPORTED_SBOX_DEGREES.contains(&d) {
+            return Err(host.error(
+                INVALID_INPUT,
+                "Unsupported s-box degree",
```
dmkozh: nit: Shall we prefix all the errors with `Poseidon: ` and `Poseidon2: ` where possible for clarity?

soroban-env-host/src/crypto/poseidon/poseidon2.rs
```diff
@@ -0,0 +1,267 @@
...
+                )
+            })?;
+            current_state = utils::add_rc(host, &current_state, rc)?;
+            current_state = utils::sbox(host, &current_state, self.params.d)?;
+            self.matmul_external(host, &mut current_state)?;
+        }
+        Ok(current_state)
+    }
+
+    fn matmul_m4(&self, host: &Host, input: &mut [F]) -> Result<(), HostError> {
```
dmkozh: A comment on what this does would be useful. Also, is there some implementation/specification reference for this?

dmkozh: nit: It's confusing that the de-facto output argument is called `input`. Just `m` or `matrix` would probably make more sense (and the name should probably contain `inplace`)

soroban-env-host/src/test/poseidon/poseidon.rs
```diff
@@ -0,0 +1,288 @@
...
+            from_hex("0x51f3e312c95343a896cfd8945ea82ba956c1118ce9b9859b6ea56637b4b1ddc4"),
+            from_hex("0x3b2b69139b235626a0bfb56c9527ae66a7bf486ad8c11c14d1da0c69bbe0f79a"),
+        ];
+
+        // Run the permutation
+        let result = poseidon.permutation(&host, &input).unwrap();
+
+        // Verify the output matches expected values
+        assert_eq!(result.len(), expected_output.len());
+        for (i, (actual, expected)) in result.iter().zip(expected_output.iter()).enumerate() {
```
dmkozh: Since it seems like Debug is implemented for the `Scalar`, couldn't this just be `assert_eq` on the vectors? Vector assertions are pretty readable IMO, and the test would be simplified a lot (especially given that you repeat this particular snippet a lot)

soroban-env-host/src/host.rs
```diff
@@ -3264,6 +3264,106 @@ impl VmCallerEnv for Host {
         self.bn254_check_pairing_output(&output)
     }
 
+    fn poseidon_permutation(
```
dmkozh: It doesn't seem like the host functions are covered by the tests. I also haven't found any tests for invalid arguments, and I think it would make sense to test those via the public (host function) interface.

soroban-env-host/src/crypto/metered_scalar.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn metered_double_in_place(&mut self, host: &Host) -> Result<(), HostError>;
+
+    /// Performs metered exponentiation, charging the budget for the operation.
+    /// Returns a new element that is `self` raised to the power of `exp`.
+    fn metered_pow(&self, exp: &u64, host: &Host) -> Result<Self, HostError>;
+
+    /// Performs metered addition, charging the budget for the operation.
+    /// Returns a new element that is the sum of `self` and `other`.
+    fn metered_add(&self, other: &Self, host: &Host) -> Result<Self, HostError> {
+        let mut result = *self;
```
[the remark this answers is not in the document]
jayz22: yeah good point. Changed it to using `metered_clone`

soroban-env-host/src/crypto/poseidon/poseidon2.rs
jayz22: changed `input` to `state` which is more accurate

soroban-env-host/src/test/poseidon/poseidon.rs
jayz22: done

soroban-env-host/src/budget.rs
```diff
@@ -618,43 +623,63 @@ impl Default for BudgetImpl {
...
                 }
                 ContractCostType::Bn254G1Mul => {
-                    cpu.const_term = 1034667;
+                    cpu.const_term = 1150435;
                     cpu.lin_term = ScaledU64(0);
                 }
                 ContractCostType::Bn254Pairing => {
-                    cpu.const_term = 5263357;
-                    cpu.lin_term = ScaledU64(391912693);
+                    cpu.const_term = 10117409;
```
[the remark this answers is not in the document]
jayz22: This came out of the re-calibration. I suspect the previous number was undercounting, probably due to not enough samples taken. The new number was done with 100k random samples and taking the maximum. This is more inline with the BLS pairing number too.

sacha699612-hub (approved):

PR #1620, opened 2025-12-04
Add some rollback smoke tests for the __check_auth instance storage.

dmkozh:
### What

Add some rollback smoke tests for the __check_auth instance storage.

### Why

These are added just to ensure that no internal errors happen. It's not possible to come up with a more interesting scenario for the re-entrant rollbacks, as both calls either succeed, or fail simultaneously (there is no way to gracefully handle an auth failure).

### Known limitations

N/A


dmkozh:
We can merge this after v25 release

PR #1628, opened 2026-01-02
feat: add try fn for with test contract frame

mootz12:
### What

Add a function that returns errors when executing a closure for a test contract frame

### Why

https://github.com/stellar/rs-soroban-sdk/issues/1434

### Known limitations

None

### DIFF
--- a/soroban-env-host/src/host/frame.rs
+++ b/soroban-env-host/src/host/frame.rs
@@ -707,16 +707,21 @@ impl Host {
 
                     if !recovered_error_from_panic_refcell {
                         self.with_debug_mode(|| {
+                            // only include func in log if a non-empty name is provided
+                            let func_str = match format!("{:?}", func).as_str() {
+                                "Symbol()" => String::new(),
+                                formatted => format!(" with fn name '{}'", formatted),
+                            };
                             if let Some(str) = panic_payload.downcast_ref::<&str>() {
                                 let msg: String = format!(
-                                    "caught panic '{}' from test contract frame '{:?}'",
-                                    str, func
+                                    "caught panic '{}' from test contract frame{}",
+                                    str, func_str
                                 );
                                 let _ = self.log_diagnostics(&msg, &[]);
                             } else if let Some(str) = panic_payload.downcast_ref::<String>() {
                                 let msg: String = format!(
-                                    "caught panic '{}' from test contract frame '{:?}'",
-                                    str, func
+                                    "caught panic '{}' from test contract frame{}",
+                                    str, func_str
                                 );
                                 let _ = self.log_diagnostics(&msg, &[]);
                             };


soroban-env-host/src/host/frame.rs
```diff
@@ -656,6 +656,75 @@ impl Host {
...
+        &self,
+        id: ContractId,
+        func: Symbol,
+        f: F,
+    ) -> Result<Val, HostError>
+    where
+        F: FnOnce() -> Result<Val, HostError>,
+    {
+        let _invocation_meter_scope = self.maybe_meter_invocation(
+            crate::host::invocation_metering::MeteringInvocation::CreateContractEntryPoint,
```
dmkozh: That's a wrong entry point, we should have used `::InvokeContract` instead. This also happened in the `with_test_contract_frame` - could you please fix both?

soroban-env-host/src/host/frame.rs
```diff
@@ -656,6 +656,75 @@ impl Host {
...
+                                let msg: String = format!(
+                                    "caught panic '{}' from contract function '{:?}'",
+                                    str, func
+                                );
+                                let _ = self.log_diagnostics(&msg, &[]);
+                            };
+                            Ok(())
+                        })
+                    }
+                    Err(self.error(error, "caught error from function", &[]))
```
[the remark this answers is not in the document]
dmkozh: I don't mind duplication much for now, I'd rather finalize this function first, then we can try de-duplication as a followup (though given that it's a test-only function, I don't think it's too important; I'd also prioritize the error message clarity over duplication for test utils).

soroban-env-host/src/host/frame.rs
```diff
@@ -656,6 +656,75 @@ impl Host {
...
+                            recovered_error_from_panic_refcell = true;
+                            error = err;
+                        }
+                    }
+
+                    if !recovered_error_from_panic_refcell {
+                        self.with_debug_mode(|| {
+                            if let Some(str) = panic_payload.downcast_ref::<&str>() {
+                                let msg: String = format!(
+                                    "caught panic '{}' from contract function '{:?}'",
```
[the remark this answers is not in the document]
mootz12: Sounds good. Removed func arg and renamed diagnostic log.

Now, diagnostic logs are emitted in line with `invoke` and `try_invoke` for `as_contract` and `try_as_contract`.

Admittedly, this log specifically is odd as it only shows during `try_*` if `panic` is used, but not if typical contract errors are used. It is the only place the panic string is caught, so it might be OK.

Either way - if we opt to change this it might make sense for a separate PR.

soroban-env-host/src/host/frame.rs
```diff
@@ -638,24 +638,85 @@ impl Host {
     /// that the closure returned (or any error caused during the frame
     /// push/pop). Used for testing.
     #[cfg(any(test, feature = "testutils"))]
-    pub fn with_test_contract_frame<F>(
-        &self,
-        id: ContractId,
-        func: Symbol,
-        f: F,
-    ) -> Result<Val, HostError>
+    pub fn with_test_contract_frame<F>(&self, id: ContractId, f: F) -> Result<Val, HostError>
```
dmkozh: While I think the change in the old function is reasonable, it breaks semver. The clean way around this would be to guard `func` with `unstable-next-api` feature, then enable the feature in the SDK. That's how we typically deal with the 'internal' API changes (i.e. those that are meant for integration with the SDK/RPC).

leighmcculloch: If I recall correctly, I think the function name was exposed so that you could give it some function name, but I don't remember why we thought the SDK should or someone might want to do that. @graydon do you remember?

dmkozh: > I don't think it's that important to remove the func field that isn't used, so rather than introducing use of unstable-next-api, I think we can keep it for now, and follow up and clean it up at the next major.

I'm pretty sure we'll forget to do that and it will stay there forever. `unstable-next-api` is almost guaranteed to be cleaned up. You don't even have to use that in the SDK. But I would prefer slating this for cleanup today, rather then trying to do that later (if we decide to clean this up at all, that is).

> I noticed that if require_auth is used inside the closure, the function names need to match, which might be a reason to keep it. 

That's fair, though the use case seems a bit suspicious to me. I guess one could use that for testing a custom account? If it's of any importance, than `as_contract` itself should provide a way to specify the function name.

dmkozh (approved):
Just a small nit-pick, I'm fine with just adding the new function for now otherwise.

dmkozh (approved):

PR #1634, opened 2026-01-17
Use `expect!` macros for e2e and simulation tests.

dmkozh:
### What

Use `expect!` macros for e2e and simulation tests to simplify intentional changes.

Note, that in the struct equality assertions we plug in the expected value from the actual data. That's intentional, as `expect!` macro update happens only at compile time, so if we used the expected value (from `expect!` macro), the test would fail a few times until all the expect! macro changes have been performed. That's fine though, as we ensure that the value is equivalent to the expected with an assertion before comparing structs.

### Why

This was always a bit of a pain to update, and it turns out that it trips up AI agents quite a bit, so doing this seems like a net productivity win - manual updates for these values never really provided any significant protection from unintentional changes.

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1667,6 +1667,7 @@ name = "soroban-simulation"
 version = "25.0.1"
 dependencies = [
  "anyhow",
+ "expect-test",
  "pretty_assertions",
  "rand",
  "soroban-env-host",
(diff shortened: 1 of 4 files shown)


leighmcculloch:
> would fail a few times until all the expect! macro changes have been performed

Is this because there are multiple expect macros in some tests? We should be able to define the expect contents as a struct or tuple, all of them at once, and then assert on the group of parameters. 

Would that allow the tests to update them in a single pass?

leighmcculloch (approved):

PR #1637, opened 2026-01-23
Implement CAP-78 - host functions for limited TTL extensions

dmkozh:
### What

This implements https://github.com/stellar/stellar-protocol/blob/master/core/cap-0078.md. It adds `extend_contract_data_ttl_v2` and `extend_contract_instance_and_code_ttl_v2`.

### Why

New host functions allow for setting lower and upper bounds on the amount of the actual extension ledgers for more fine-grained control over the extension strategy and thus user fees. See more detailed motivation in CAP.

### Known limitations

N/A


soroban-env-host/src/storage.rs
```diff
@@ -523,6 +523,132 @@ impl Storage {
...
+        let ttl_ext_final = ttl_ext_init.min(max_extension).min(max_network_extension);
+
+        // If final extension < min_extension, return without changes
+        if ttl_ext_final < min_extension {
+            return Ok(());
+        }
+
+        let new_live_until = old_live_until.saturating_add(ttl_ext_final);
+
+        if new_live_until > old_live_until {
```
sisuresh: This is only required because we allow `max_extension` == 0 right? I guess a noop here is better than throwing an error earlier, but a contract that does this is probably doing something it shouldn't.

dmkozh: I think it's useful to have an explicit invariant right where we update TTL (as we did with the v1 function). I removed the early return check from the new function for clarity, the difference should be insignificant.

sisuresh (approved):

soroban-env-host/src/storage.rs
```diff
@@ -523,6 +523,132 @@ impl Storage {
...
+                ScErrorCode::InvalidInput,
+                "max_extension must be >= min_extension",
+                &[max_extension.into(), min_extension.into()],
+            ));
+        }
+
+        #[cfg(any(test, feature = "recording_mode"))]
+        self.handle_maybe_expired_entry(&key, host)?;
+
+        // Extending deleted/non-existing/out-of-footprint entries will result in
```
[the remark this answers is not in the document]
sisuresh: I called out the two error checks because those were the same between the two functions, but that change you made is good too. Gets rid of more duplicated logic.

PR #1678, opened 2026-04-08
Bump to P27, CAP-0071 XDR, and add feature for xdr

sisuresh:
### What

This PR bumps env to p27, and adapts to the new xdr framework with ifdefs (https://github.com/stellar/pod-fsr/issues/12). It doesn't actually implement CAP-0071.


Cargo.toml
```diff
@@ -36,12 +36,10 @@ wasmparser = "=0.116.1"
 # NB: When updating, also update the version in rs-soroban-env dev-dependencies
 [workspace.dependencies.stellar-xdr]
 version = "=26.0.0"
-#git = "https://github.com/stellar/rs-stellar-xdr"
-#rev = "99c73b18ccd68bc3439be30801da6261b193d2da"
+git = "https://github.com/sisuresh/rs-stellar-xdr"
```
sisuresh: Should be updated to use https://github.com/stellar/rs-stellar-xdr/pull/524

sisuresh:
Not sure why, but CodeQL is failing on existing test code.

soroban-env-host/observations/27/test__address__invalid_muxed_address_to_strkey_conversion.json
```diff
@@ -0,0 +1,10 @@
+{
```
dmkozh: Why do we have two observation locations (observations/ and observations/27)?

soroban-simulation/src/test/simulation.rs
```diff
@@ -340,7 +340,10 @@ fn test_simulate_create_contract() {
     );
     assert!(res.contract_events.is_empty());
     assert!(res.diagnostic_events.is_empty());
+    #[cfg(not(feature = "cap_0071"))]
     expect!["2742756"].assert_eq(&res.simulated_instructions.to_string());
+    #[cfg(feature = "cap_0071")]
+    expect!["2742828"].assert_eq(&res.simulated_instructions.to_string());
```
dmkozh: I'm curious, did the feature gate work well with UPDATE_EXPECT?

sisuresh: I had claude deal with it and didn't have any issues. I can try to see what the manual process would look like.

sisuresh: You would have to run both 
```
UPDATE_EXPECT=true cargo test                         
UPDATE_EXPECT=true cargo test --features cap_0071
```

sisuresh: Hmm this will get messy with multiple features

soroban-env-host/observations/27/test__address__invalid_muxed_address_to_strkey_conversion.json
sisuresh: Good catch. I missed this on the rebase.

sisuresh: fixed

soroban-simulation/src/test/simulation.rs
sisuresh: I updated to just skip the line where cap_0071 caused a diff. Not sure what a good solution is here if we have multiple features.

dmkozh: I think the current solution is ok; I believe we've decided that we don't want to be super generic in terms of supporting cartesian product of all the possible feature combinations. So we can just agree that tests reflect the superset of the features that env supports (currently an empty set).

dmkozh (approved):

PR #1679, opened 2026-04-09
Multi protocol cleanup

sisuresh:
### What

We don't use the observations for older protocols, and the host doesn't need to maintain backwards compatibility.


dmkozh (commented):
Why haven't you removed p23-25?

sisuresh:
> Why haven't you removed p23-25?

Hmm not sure what you mean. I removed the observation directories for them. Did I miss something else?

dmkozh (approved):

PR #1684, opened 2026-05-12
Return error when invoking account addresses as contracts

### LINKED ISSUE #1683: `try_invoke_contract` panics if the invoked address is not a contract

main

### What did you do?

From the SDK, called `Env::try_invoke_contract` with an `Address` that is a Stellar account
address (`G...`) instead of a contract address (`C...`).

`try_invoke_contract` / `try_call` should return an error to the caller instead
of panicking or trapping.

Preferably, an account address should use the same external error shape as other
recoverable non-contract invocation failures, such as trying to call a function
that does not exist.

### What did you see instead?

`try_invoke_contract` panics/aborts before returning a `Result`.

mootz12:
### What

Updates contract invocation so `try_call` returns the default contract call host error when given an account Address instead of a contract Address.

Updates `call` to use the same logic, so diagnostic events remain consistent.

### Why

Fixes: #1683

This aligns errors to be along the mental modal of an account address is just an address with no invokable functions.

### Known limitations

- If we disagree this is the correct approach, we can instead return the invalid action error directly when parsing `contract_id` in `try_call` only, and leave `call` untouched. These errors will not appear in the diagnostics.
```rust
Ok(Error::from_type_and_code(
    ScErrorType::Context,
    ScErrorCode::InvalidAction,
)
.to_val())
```

### DIFF
--- a/soroban-env-host/src/host.rs
+++ b/soroban-env-host/src/host.rs
@@ -2589,12 +2589,18 @@ impl VmCallerEnv for Host {
         let argvec = self.call_args_from_obj(args)?;
         // this is the recommended path of calling a contract, with `reentry`
         // always set `ContractReentryMode::Prohibited`
-        let res = self.call_n_internal(
-            &self.contract_id_from_address(contract_address)?,
-            func,
-            argvec.as_slice(),
-            CallParams::default_external_call(),
-        );
+        // Treat non-contract `contract_address` errors as a contract call failure, as
+        // from the guest's perspective everything is an Address
+        let res = self
+            .contract_id_from_address(contract_address)
+            .and_then(|contract_id| {
+                self.call_n_internal(
+                    &contract_id,
+                    func,
+                    argvec.as_slice(),
+                    CallParams::default_external_call(),
+                )
+            });
         if let Err(e) = &res {
             self.error(
                 e.error,
@@ -2627,12 +2633,18 @@ impl VmCallerEnv for Host {
         // TODO: A `reentry` flag will be passed from `try_call` into here.
         // For now, we are passing in `ContractReentryMode::Prohibited` to disable
         // reentry.
-        let res = self.call_n_internal(
-            &self.contract_id_from_address(contract_address)?,
-            func,
-            argvec.as_slice(),
-            CallParams::default_external_call(),
-        );
+        // Treat `contract_id_from_address` errors as a recoverable contract call
+        // failure, as from the guest's perspective everything is an Address
+        let res = self
+            .contract_id_from_address(contract_address)
+            .and_then(|contract_id| {
+                self.call_n_internal(
+                    &contract_id,
+                    func,
+                    argvec.as_slice(),
+                    CallParams::default_external_call(),
+                )
+            });
         match res {
             Ok(rv) => Ok(rv),
             Err(e) => {
--- a/soroban-env-host/observations/test__invocation__invoke_with_account_address_err.json
+++ b/soroban-env-host/observations/test__invocation__invoke_with_account_address_err.json
@@ -1,0 +1,10 @@
+{
+  "   0 begin": "cpu:14488, mem:0, prngs:-/9b4a753, objs:-/-, vm:-/-, evt:-, store:-/-, foot:-, stk:-, auth:-/-",
+  "   1 call vec_new()": "cpu:14930, mem:80, objs:-/1@ad26cd6b",
+  "   2 ret vec_new -> Ok(Vec(obj#3))": "cpu:15433, mem:160, objs:-/2@c68e8e8a",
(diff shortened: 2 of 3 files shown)


soroban-env-host/src/host.rs
```diff
@@ -2589,12 +2589,18 @@ impl VmCallerEnv for Host {
...
-        let res = self.call_n_internal(
-            &self.contract_id_from_address(contract_address)?,
-            func,
-            argvec.as_slice(),
-            CallParams::default_external_call(),
-        );
+        // Treat non-contract `contract_address` errors as a contract call failure, as
+        // from the guest's perspective everything is an Address
+        let res = self
+            .contract_id_from_address(contract_address)
```
dmkozh, line 2595: I don't think we should do this change in `call`. It obfuscates the actual error behind the 'contract call failed' (like you could have a budget error with 'contract call failed', which is weird).

soroban-env-host/src/host.rs
```diff
@@ -2627,12 +2633,18 @@ impl VmCallerEnv for Host {
...
-        let res = self.call_n_internal(
-            &self.contract_id_from_address(contract_address)?,
-            func,
-            argvec.as_slice(),
-            CallParams::default_external_call(),
-        );
+        // Treat `contract_id_from_address` errors as a recoverable contract call
+        // failure, as from the guest's perspective everything is an Address
+        let res = self
+            .contract_id_from_address(contract_address)
```
dmkozh, line 2639: It's debatable whether calling non-contract id should be treated in the same way as a failure during the contract execution. I guess we gracefully handle calls to non-existent contract (which is a bit weird TBH), which sets a precedent for gracefully handling an error that occurs before the contract call, but it's not obvious if that is the right thing.
More importantly, it's a protocol change, and it's not clear if it's entirely safe. Like a caller may rely on the fact that `try_call` will not handle calls to non-contract addresses gracefully and something weird happens. That's quite unlikely, but is a consideration. The question is basically whether this provides more UX benefit vs the potential risk.

soroban-env-host/src/host.rs
mootz12: The only change here (should) be to include any `contract_id_from_address` HostError to `res`, so it follows the same codepath as a `call_n_internal` error.

The existing HostError behavior from `call_n_internal` is unchanged.

soroban-env-host/src/host.rs
mootz12: > More importantly, it's a protocol change, and it's not clear if it's entirely safe. Like a caller may rely on the fact that try_call will not handle calls to non-contract addresses gracefully and something weird happens.

Yeah, this is true. Any thoughts on how we could attempt to check this? I suppose it would probably need to go though some public review process?

In the same vein, it might be more likely a contract assumes an invalid, user provided `Address` value would get caught.

In the end, this might just need to be clearly documented on the SDK, so the behavior is more apparent, as I don't have any tangible UX improvement for the change. This was just caught in some ongoing auditing efforts.

soroban-env-host/src/host.rs
dmkozh: Yes, I understand that. I guess there is the same issue as the one I've raised below: we already do a weird thing here where we may wrap an unrelated error into 'contract call failed', like e.g. a budget error. I'm not really sure if we should be using e.error in the error report below, this piece of code is really old and quite weird...

soroban-env-host/src/host.rs
dmkozh: >  Any thoughts on how we could attempt to check this? 

Not really, that would require what the code does and what the assumptions are. try_call in general is kind of tricky and arguably it shouldn't be used most of the time. Again, I recognize that the risk is low, but the benefit might still not be worth it.

> In the same vein, it might be more likely a contract assumes an invalid, user provided Address value would get caught.

Yeah, maybe, we can only speculate at this point really. Like during testing panic may actually be preferable if you accidentally passed the wrong argument. 
OTOH if the concern is that arbitrary user input may cause the contract to fail even though it uses `try_call`, then I'd argue that's not an issue worth fixing, because there is a ton of ways of causing a non-recoverable error in the contract logic, so if your system allows arbitrary contract calls, it should be ready for these failing irrecoverably. 
I don't really like the inconsistency here either, but if I were to change this, I would maybe reduce the recoverable surface to just the contract logic itself instead of increasing it. 
Anyways, if there is no immediate UX benefit or concern, I would rather leave things as they are - I don't recall this issue coming up from any actual contract developers.

mootz12: > Anyways, if there is no immediate UX benefit or concern, I would rather leave things as they are - I don't recall this issue coming up from any actual contract developers.

I agree. Thanks for the feedback.

I'll close this PR for now, and instead document the behavior on the SDK side.

mootz12:
Closing after feedback in the comments.

PR #1690, opened 2026-06-12
Gate cap_0083 XDR under the next feature

sisuresh:
### What

Enable `stellar-xdr/cap_0083` through soroban-env-common's `next` feature, so next-enabled builds compile in the CAP-83 XDR while production builds keep it excluded. No `Cargo.lock` change; verified `cargo check -p soroban-env-host` passes with and without `--features next`.

### Why

Follow-up to #1687, which dropped the hardcoded `cap_0083` feature but left no way to enable it at all — core's vnext builds (which enable the CAP-83 XDR on the C++ side) had no matching knob for the host. Core already passes `--features next` to the max-protocol host in vnext builds, so no core changes are needed.

Going forward, gated XDR features for next-protocol CAPs ride `next`: each entry becomes unconditional at the protocol bump and is removed once folded into the base XDR upstream (documented in the Cargo.toml comment).

### DIFF
--- a/soroban-env-common/Cargo.toml
+++ b/soroban-env-common/Cargo.toml
@@ -37,7 +37,10 @@ std = ["stellar-xdr/std", "stellar-xdr/base64"]
 serde = ["dep:serde", "stellar-xdr/serde"]
 wasmi = ["dep:wasmi", "dep:wasmparser"]
 testutils = ["dep:arbitrary", "stellar-xdr/arbitrary"]
-next = ["soroban-env-macros/next"]
+# `next` also enables the gated XDR of CAPs targeting the next protocol. Each
+# cap_XXXX entry becomes unconditional (moved to the workspace stellar-xdr
+# dep) at the protocol bump, and is removed once folded into the base XDR.
+next = ["soroban-env-macros/next", "stellar-xdr/cap_0083"]
 tracy = ["dep:tracy-client"]
 shallow-val-hash = []
 


dmkozh:
What is the point in conditioning it at all? We've cut v27 env, all the changes from now on are v28 by default.

anupsdf:
> What is the point in conditioning it at all? We've cut v27 env, all the changes from now on are v28 by default.

Since v28 is in 50 days or so, maybe this helps in case we have to do an emergency release of env (for security fixes or simulation library changes) before v28. wdyt?

dmkozh:
> (for security fixes

That's unlikely to matter. Even if the change is non-protocol, it will still go in a separate branch.

> simulation library changes

That's more plausible, and just enabling an XDR flag is rather harmless, so we can keep this for now I suppose. But I'm still not a fan of excessive feature gating, especially given that CAP-83 has nothing to do with env and we could just skip bumping this until CAP-83 has been promoted to 'stable' in XDR (at which point it will be picked up by env automatically). The only value in enabling the feature in env is to ensure that XDR hashes are the same between Core and Env, which we normally do just for the stable protocol versions anyways.

anupsdf:
> > (for security fixes
> 
> That's unlikely to matter. Even if the change is non-protocol, it will still go in a separate branch.
> 
> > simulation library changes
> 
> That's more plausible, and just enabling an XDR flag is rather harmless, so we can keep this for now I suppose. But I'm still not a fan of excessive feature gating, especially given that CAP-83 has nothing to do with env and we could just skip bumping this until CAP-83 has been promoted to 'stable' in XDR (at which point it will be picked up by env automatically). The only value in enabling the feature in env is to ensure that XDR hashes are the same between Core and Env, which we normally do just for the stable protocol versions anyways.

Do we also match the xdr hash between core and env for vnext builds? I wonder if thats a motivation so we could test cap-83 in futurenet using vnext core package.

dmkozh:
> Do we also match the xdr hash between core and env for vnext builds? 

I don't think we do, though this has been switched back-and-forth a few times. But we actually match the hashes of the input files (including ifdefs), so this has nothing to do with the features (so disregard my comment above regarding hash matching; even that doesn't require env to have the feature enabled).

> I wonder if thats a motivation so we could test cap-83 in futurenet using vnext core package.

This shouldn't be a blocker for vnext package in any way; it would be suspicious if that's a blocker.

sisuresh:
> What is the point in conditioning it at all? We've cut v27 env, all the changes from now on are v28 by default.

This PR isn't required because cap83 xdr isn't required in the host, and we don't do the xdr feature matching in core for vnext (I've thought about changing this though). This PR is just setting up the state it should've been in. I was just looking into what we would need to avoid the issue I ran into with the v27 release where I removed the cap-0083 feature.

sisuresh:
After speaking with @dmkozh, we came to the conclusion that in scenarios like this where the XDR has no effect on the host, we can just not add the XDR feature for vnext because it technically isn't required (which is why vnext builds currently works for CAP 83 even though the feature isn't enabled in the host). This slightly reduces the operational burden of dealing with these ifdefs.

PR #1691, opened 2026-06-14
Add external read-only contract storage access

mihaieremia:
Discussion: https://github.com/stellar/stellar-protocol/discussions/1958

### What

Adds protocol-28 Soroban host functions for read-only external contract storage access:

- `has_external_contract_data(contract, key, storage_type)`
- `get_external_contract_data(contract, key, storage_type)`

Persistent and temporary reads resolve to the target contract-data key. Instance reads resolve to the target contract instance entry and look up the embedded instance-storage map.

This also adds host tests, protocol-gate coverage, dispatch coverage, and a Wasm fixture importing the new host functions.

### Why

Some contracts need read-only state owned by another contract, but do not need to execute that contract's Wasm. A lending controller is one example: it may need market config, reserve parameters, oracle references, collateral factors, or pause flags owned by market/governance/oracle contracts.

Today that read requires a cross-contract invoke to a getter. This PR adds a direct read-only path for intentionally public storage while preserving contract ownership boundaries.

### Local P28 preflight measurement

Measured with local protocol-28 RPC preflight using two minimal contracts:

- `Store`: `noop`, persistent getter, temporary getter, instance getter
- `Caller`: `noop`, cross-contract invoke-to-getter methods, direct external-storage read methods

The instance case uses a 32-entry target instance map.

| Case | Instructions | RO entries | Resource fee |
|---|---:|---:|---:|
| caller noop | 454,842 | 2 | 50,399 |
| caller -> store noop | 924,152 | 4 | 53,873 |
| store persistent getter | 558,111 | 3 | 52,052 |
| caller -> persistent getter | 950,847 | 5 | 55,197 |
| caller external persistent read | 483,703 | 3 | 52,563 |
| store temporary getter | 558,107 | 3 | 52,052 |
| caller -> temporary getter | 950,843 | 5 | 55,197 |
| caller external temporary read | 483,703 | 3 | 52,563 |
| store instance getter | 537,882 | 2 | 50,955 |
| caller -> instance getter | 929,322 | 4 | 54,096 |
| caller external instance read | 586,622 | 3 | 52,611 |

Result:

- Persistent external read: `483,703` instructions vs `950,847` for cross-contract getter.
- Temporary external read: `483,703` instructions vs `950,843` for cross-contract getter.
- Instance external read: `586,622` instructions vs `929,322` for cross-contract getter.

Persistent and temporary external reads are close to a direct getter baseline and avoid the nested VM dispatch plus extra footprint entries from invoking the target contract. Instance reads still save work compared to a getter call, but are less close to native reads because the target instance entry and storage map must be loaded and searched.

### Open protocol questions

- Should raw external reads require public storage declarations or layout metadata?
- Should instance storage be included in v1?
- What exact same-transaction visibility rules should apply when the target contract is active on the call stack?
- When should the SDK expose this API relative to storage layout standardization?

### Limitations

- Protocol 28 only.
- Read-only only: no external writes or TTL extension.
- Does not run target contract code, auth, schema validation, or public-storage checks.
- Companion SDK/RPC changes are needed for end-to-end use.

### Validation

- `cargo test -p soroban-env-host external_contract_data --features next`
- `cargo check -p soroban-sdk --features next`
- `cargo test -p soroban-sdk --test external_storage_next --features "next testutils"`
- local P28 RPC preflight scratch test comparing getter calls with external reads
- `git diff --check`


dmkozh:
Thank you for opening this PR! However, as this is a protocol change, we need a CAP and a protocol discussion first. Would you mind starting a discussion at https://github.com/orgs/stellar/discussions? You're also welcome to make the respective CAP  contribution at  https://github.com/stellar/stellar-protocol/tree/master/core, but it's also fine to just start with the discussion and we can work on the CAP on our side (either way, it should be rather straightforward). 

Also, I must say that while the change is straightforward implementation-wise, there might be some security implications for the existing contracts, which is why the discussion is necessary, and it's also not clear if we can proceed with this approach in the first place.

soroban-env-common/env.json
```diff
@@ -1708,6 +1708,94 @@
...
+                            "type": "StorageType"
+                        }
+                    ],
+                    "return": "Val",
+                    "docs": "Return the value stored by the provided contract at the provided contract data key. This read-only operation records the target contract data key in the transaction footprint.",
+                    "min_supported_protocol": 28
+                },
+                {
+                    "export": "k",
+                    "name": "try_get_external_contract_data",
```
dmkozh: This interface is rather awkward and inconsistent with the existing storage functions. I think we should either:
- get rid of this function - has and get have us covered on the SDK-side for the `try_` function implementation
- if there is a good evidence that dedicated `try_` function saves us a lot of instructions vs has + get, then we should return an 'option' value from the function instead of writing the output to linear memory. But I really think just dropping this is easier and not significantly more expensive.

dmkozh (commented):
Given my response on the PR re CAP, I haven't looked into this beyond the function interface (which is relevant for the discussion and spec).

mihaieremia:
> Thank you for opening this PR! However, as this is a protocol change, we need a CAP and a protocol discussion first. Would you mind starting a discussion at https://github.com/orgs/stellar/discussions? You're also welcome to make the respective CAP contribution at https://github.com/stellar/stellar-protocol/tree/master/core, but it's also fine to just start with the discussion and we can work on the CAP on our side (either way, it should be rather straightforward).
> 
> Also, I must say that while the change is straightforward implementation-wise, there might be some security implications for the existing contracts, which is why the discussion is necessary, and it's also not clear if we can proceed with this approach in the first place.

Hey,

I will start a discussion right away 👍 

I drafted this PR quickly to get some feedback since I hit different issues building our lending protocol which went a bit over the limits and made my read more of how it actually works behind 🙏

dmkozh:
> I drafted this PR quickly to get some feedback since I hit different issues building our lending protocol which went a bit over the limits and made my read more of how it actually works behind

BTW on this note, I'm not sure I trust the numbers you've shared in the PR - what was the methodology for collecting these? I would recommend testing your contract setup on the testnet or a local network and lookup the view call numbers that are coming from the simulation (e.g. compare a no-op function call with a function call that does a cross-contract call to a different function). The reason for why I'm skeptical about this is that 80M instructions seems like too much for a VM instantiation, even if your contract is large. I could be wrong, but it wouldn't hurt to double-check - the issue might not be as bad as tests present.

mihaieremia:
> > I drafted this PR quickly to get some feedback since I hit different issues building our lending protocol which went a bit over the limits and made my read more of how it actually works behind
> 
> BTW on this note, I'm not sure I trust the numbers you've shared in the PR - what was the methodology for collecting these? I would recommend testing your contract setup on the testnet or a local network and lookup the view call numbers that are coming from the simulation (e.g. compare a no-op function call with a function call that does a cross-contract call to a different function). The reason for why I'm skeptical about this is that 80M instructions seems like too much for a VM instantiation, even if your contract is large. I could be wrong, but it wouldn't hurt to double-check - the issue might not be as bad as tests present.

You were right to be skeptical of the original numbers. The 80-85M figure was not a single simulation result; it was an aggregate across 100 iterations, and it measured the old cross-contract getter path rather than the new external read path.

I reran the comparison using local P28 simulation/preflight with the local env + SDK + RPC changes.

Default Rust/Wasm stack:

| Case | Instructions | Memory bytes |
|---|---:|---:|
| caller no-op | 454,842 | 1,157,304 |
| persistent: external read | 483,703 | 1,165,474 |
| persistent: cross-contract getter | 950,847 | 2,380,138 |
| temporary: external read | 483,703 | 1,165,474 |
| temporary: cross-contract getter | 950,843 | 2,380,138 |
| instance: external read | 586,622 | 1,227,910 |
| instance: cross-contract getter | 929,322 | 2,373,328 |

So external reads are roughly 2x cheaper than cross-contract getters for persistent/temporary storage, and still cheaper for instance storage.

While looking into the memory numbers, I found that recent contracts built with the default Rust/Wasm settings seem to reserve about 1MiB of Wasm linear memory per contract invocation. That explains why cross-contract calls can become expensive in memory: each invoked contract brings its own baseline reservation.

After digging into the compile flags, I rebuilt my own contracts with:

`RUSTFLAGS='-C link-arg=-zstack-size=16384'`


That reduced the baseline memory a lot. Cross-contract calls between contracts deployed with this smaller stack are much cheaper in metered memory. But if my contract calls an already deployed oracle contract that was built with the default stack, the call still pays that contract’s larger memory reservation.

With `RUSTFLAGS='-C link-arg=-zstack-size=16384'`:

| Case | Instructions | Memory bytes |
|---|---:|---:|
| caller no-op | 332,098 | 174,288 |
| persistent: external read | 360,959 | 182,458 |
| persistent: cross-contract getter | 705,002 | 413,840 |
| temporary: external read | 360,959 | 182,458 |
| temporary: cross-contract getter | 704,998 | 413,840 |
| instance: external read | 463,878 | 244,894 |
| instance: cross-contract getter | 683,477 | 407,030 |

This matters for my lending deployment because instructions usually fit under the 400M limit, but many oracle price reads pushed memory toward the limit.

So I have two questions:

1. Does this simulation/preflight methodology look more accurate ?
2. Is the 1MiB default stack reservation expected/recommended for Soroban contracts, or should contracts generally use a smaller `zstack-size`?

dmkozh:
> So external reads are roughly 2x cheaper than cross-contract getters for persistent/temporary storage, and still cheaper for instance storage.

I think 2x is not exactly the right number to look at here; in the end we're talking about a few hundred thousand instructions per call. I realize it's not nothing, but it's also not that high either compared to the limit.

> Does this simulation/preflight methodology look more accurate ?

Well, I haven't seen your exact setup/code, so I can't tell if it's fully accurate, but the numbers seem to be in the right ballpark, so this is probably good enough.

> Is the 1MiB default stack reservation expected/recommended for Soroban contracts, or should contracts generally use a smaller zstack-size?

I don't think it's expected, but I need to double-check with the team. 1 MB seems excessive to me. 16K may or may not be sufficient. Default probably should be somewhere in-between, but closer to 16K - I wouldn't expect contracts to do many nested function calls and use many locals. Basically this requires some additional research on our side. But as long your contract works fine it definitely makes sense to reduce the stack size (if you can still upgrade it, that is).

mihaieremia:
> > So external reads are roughly 2x cheaper than cross-contract getters for persistent/temporary storage, and still cheaper for instance storage.
> 
> I think 2x is not exactly the right number to look at here; in the end we're talking about a few hundred thousand instructions per call. I realize it's not nothing, but it's also not that high either compared to the limit.
> 
> > Does this simulation/preflight methodology look more accurate ?
> 
> Well, I haven't seen your exact setup/code, so I can't tell if it's fully accurate, but the numbers seem to be in the right ballpark, so this is probably good enough.
> 
> > Is the 1MiB default stack reservation expected/recommended for Soroban contracts, or should contracts generally use a smaller zstack-size?
> 
> I don't think it's expected, but I need to double-check with the team. 1 MB seems excessive to me. 16K may or may not be sufficient. Default probably should be somewhere in-between, but closer to 16K - I wouldn't expect contracts to do many nested function calls and use many locals. Basically this requires some additional research on our side. But as long your contract works fine it definitely makes sense to reduce the stack size (if you can still upgrade it, that is).

We’ll most likely deploy with the 16 KB Wasm page-size flag if final tests keep confirming it’s safe.

This should help our lending protocol, but also anything built on top of it, like vaults or more complex position managers.

What I’d like to check is whether teams like Reflector or RedStone could also safely build with this flag. If yes, it could help us, Blend, and other DeFi protocols support more positions per user and more complex logic.

For reference, Rust’s Wasm target already exposes these kinds of low-level linker settings here:
https://github.com/rust-lang/rust/blob/main/compiler/rustc_target/src/spec/base/wasm.rs#L14

Not saying everyone should blindly use it, but it’s probably worth testing/reviewing at ecosystem level.

PR #1694, opened 2026-06-17
Update stellar-strkey to 0.0.18

leighmcculloch:
### What

Bumps the direct `stellar-strkey` dependency in `soroban-env-host` from 0.0.13 to 0.0.18, and adapts the two strkey-encoding call sites in `host/conversion.rs` to the new API. In 0.0.17+ the crate is `no_std` by default and `Strkey::to_string()` now returns a fixed-capacity `heapless::String` rather than `std::string::String`, so those sites now convert via `.as_str().to_string()`. The lockfile re-resolves to contain both 0.0.18 (for `soroban-env-host`) and 0.0.13 (transitively via the published `stellar-xdr` 27.0.0, which still pins 0.0.13); the 0.0.13 entry is expected and left in place, and `deny.toml` skips the duplicate.

### Why

This moves `soroban-env-host` onto the stellar-strkey release line targeted for protocol 28.

This supersedes the earlier 0.0.17 bump: 0.0.17 contained a CLI bug (`Decoded<Strkey>` rejected owned JSON keys, breaking `strkey encode`), fixed in 0.0.18 by stellar/rs-stellar-strkey#125.

### Known limitations

N/A

---

## Protocol 28 coordination

⚠️ This change targets versions of software that will ship for **Protocol 28**, and should **not** be merged unless this repository is ready to accept Protocol 28 changes.

This is one of a coordinated set of PRs bumping `stellar-strkey` to **0.0.18** across the Protocol 28 component stack. It is preferred that all Protocol 28 releases of these components depend on the **same** version of `stellar-strkey`:

- stellar/rs-stellar-xdr#547
- stellar/rs-soroban-env#1694
- stellar/rs-soroban-sdk#1913
- stellar/rs-stellar-rpc-client#99
- stellar/stellar-cli#2614

`stellar-strkey` 0.0.18 raises the minimum supported Rust version to 1.87 and updates/adds transitive dependencies (`heapless` 0.9, `zeroize`), so MSRV and supply-chain policy (cargo-deny / cackle) updates may be required in some repositories before CI is green.

### DIFF
--- a/cackle.toml
+++ b/cackle.toml
@@ -262,6 +262,11 @@ build.allow_apis = [
 
 [pkg.heapless]
 allow_unsafe = true
+build.allow_apis = [
+    "env",
+    "fs",
+    "process",
+]
 
 [pkg.tracking-allocator]
 allow_unsafe = true
--- a/cackle.toml
+++ b/cackle.toml
@@ -109,6 +109,9 @@ allow_apis = [
 
 [pkg.rand]
 allow_unsafe = true
+allow_apis = [
+    "rand",
+]
 
 [pkg.backtrace]
 allow_apis = [
@@ -144,6 +147,18 @@ allow_unsafe = true
 allow_unsafe = true
 
 [pkg.subtle]
+allow_unsafe = true
+
+[pkg.byteorder]
+allow_unsafe = true
+
+[pkg.hash32]
+allow_unsafe = true
+
+[pkg.stable_deref_trait]
+allow_unsafe = true
+
+[pkg.stellar-strkey]
 allow_unsafe = true
 
 [pkg.zeroize]
@@ -246,11 +261,7 @@ build.allow_apis = [
 ]
 
 [pkg.heapless]
-build.allow_apis = [
-    "env",
-    "fs",
-    "process",
-]
+allow_unsafe = true
 
 [pkg.tracking-allocator]
 allow_unsafe = true
--- a/cackle.toml
+++ b/cackle.toml
@@ -245,6 +245,13 @@ build.allow_apis = [
     "env",
 ]
 
+[pkg.heapless]
+build.allow_apis = [
+    "env",
+    "fs",
+    "process",
+]
+
 [pkg.tracking-allocator]
 allow_unsafe = true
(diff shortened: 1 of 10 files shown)


soroban-env-host/src/crypto/poseidon/poseidon2.rs
```diff
@@ -89,7 +89,7 @@ impl<F: MeteredScalar> Poseidon2<F> {
     /// (https://eprint.iacr.org/2023/323)
     fn matmul_m4(&self, host: &Host, state: &mut [F]) -> Result<(), HostError> {
         let t = self.params.t;
-        if t % 4 != 0 {
+        if !t.is_multiple_of(4) {
```
leighmcculloch, line 92: This change is a result of new clippy warning that appeared in the change from running clippy on Rust v1.84 vs v1.87.

Ref: https://rust-lang.github.io/rust-clippy/master/?search=manual_is_multiple_of#manual_is_multiple_of

soroban-env-host/src/host/conversion.rs
```diff
@@ -686,7 +686,7 @@ impl Host {
                 ))
             }
         };
-        Ok(strkey.to_string())
+        Ok(strkey.to_string().as_str().to_string())
```
leighmcculloch, line 689: The string returned from strkey's is now a heapless string, so this seemingly odd dance is converting that heapless string into an alloc String.

cackle.toml
```diff
@@ -146,6 +149,18 @@ allow_unsafe = true
...
+allow_unsafe = true
+
+[pkg.hash32]
+allow_unsafe = true
+
+[pkg.stable_deref_trait]
+allow_unsafe = true
+
+[pkg.stellar-strkey]
+allow_unsafe = true
```
leighmcculloch, line 162: The strkey package gained an unsafe line around a conversion of a base32 string into a byte list into a str type which does the conversion unchecked because the base32 alphabet is all ascii.

leighmcculloch (commented):
Few comments inline for why changes are required.

dmkozh (approved):

PR #1698, opened 2026-06-29
Lazily decode ledger entries passed in e2e flow.

dmkozh:
### What

Lazily decode every storage entry on first access, which allows us to avoid deserializing some entries that never need to be read (notably, contract code entries which we only need to read occasionally in some edge cases).

This also involved breaking e2e interface changes that need to be supported from the Core side:

- Pass input entries in TTL order, so that we don't need to extract entry keys
- Pass TTLs zipped with the storage entries instead of passing them as a separate `TTLEntry` (another small metering win for the users)
- Also pass initial entry size for rent for all the entries with TTL, which Core already knows. This allows us to skip code entry encoding when we're doing read-only rent bumps (we could partially parse just the extension entry because it's the only necessary input, but partial decoding is not supported cleanly)

I also encapsulated more of the storage-related internal data structures: nobody reads them and nobody should read them, as these are implementation details.

### Why

The motivation behind this change is the fact that we're charging a very significant chunk of budget for decoding the contract code entries that we actually never need to read thanks to the module cache (e.g. ~50% of instructions for a Soroswap transaction).

### Known limitations

N/A


soroban-env-host/src/budget.rs
```diff
@@ -371,8 +371,8 @@ impl Default for BudgetImpl {
                     cpu.lin_term = ScaledU64(29);
                 }
                 ContractCostType::ValDeser => {
-                    cpu.const_term = 59052;
-                    cpu.lin_term = ScaledU64(4001);
+                    cpu.const_term = 331;
+                    cpu.lin_term = ScaledU64(4369);
```
dmkozh, line 375: Our hardcoded cost types have never been updated since p20, even though we've done re-calibration shortly after. This is just one re-calibrated cost type update that I needed in order to get the test working properly (i.e. to demonstrate that we don't charge for the expensive deserialization). We should update the rest of the weights separately (cc @jayz22 )

soroban-env-host/src/e2e_invoke.rs
```diff
@@ -1021,100 +1088,114 @@ fn build_storage_footprint_from_xdr(
...
+        Storage::check_supported_ledger_key_type(key)?;
+        let key_rc = Rc::metered_new(key.metered_clone(budget)?, budget)?;
+
+        let maybe_entry_with_live_until = match maybe_entry {
+            Some(entry_buf) => {
+                let entry_bytes = entry_buf.as_ref();
+                budget.charge(ContractCostType::MemAlloc, Some(entry_bytes.len() as u64))?;
+                budget.charge(ContractCostType::MemCpy, Some(entry_bytes.len() as u64))?;
+                let entry_bytes_rc: Rc<[u8]> = Rc::from(entry_bytes);
+                let entry = Rc::metered_new(HostLedgerEntry::from_encoded(entry_bytes_rc), budget)?;
```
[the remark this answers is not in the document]
dmkozh: That's a reasonable suggestion, but I'm not sure if doing a bunch of additional work is worth it - we put a lot of responsibility on the embedder (like zipping that we've added with this PR), and I'm not sure if this particular check is valuable enough.

Curious to see what the others think though.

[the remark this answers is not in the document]
graydon: I don't love losing the check. Maybe it'd be possible to check footprint inclusion at the moment of lazy decode?

soroban-env-host/src/host/ledger_entry.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Decoding happens at most once and is memoized in place. Thanks to that, if
+/// the same `HostLedgerEntry` is shared between the containers via `Rc` (which
+/// is what the storage does), decoding will happen at most once across all of
+/// them.
+///
+/// This wrapper allows the host to avoid unnecessary XDR decoding in cases
+/// when the entry is not accessed during the contract execution (which is what
+/// typically happens to the contract code entries, which are normally fetched
+/// from the module cache).
+pub(crate) struct HostLedgerEntry {
```
graydon: The name feels like it's not especially suggestive -- like sure it's in the host but so what? Would it work to call it `LazyLedgerEntry` or does that confuse matters a different way?

graydon (approved):
I think I'm mostly ok with this. Two small comments but neither should block if you want to land now. Up to you. Nice work overall!

soroban-env-host/src/e2e_invoke.rs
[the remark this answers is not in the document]
dmkozh: Footprint inclusion is guaranteed by the storage creation invariant (storage is built from the footprint keys in the first place). What we are missing here is the check that ensures that after we have decoded the entry, its key corresponds to the footprint key we're looking up currently (i.e. it's a stronger invariant than just 'is it in the footprint', it's rather 'is it actually the entry corresponding to the specific footprint key'). Previously this check was a by-product of the storage building mechanism, but now it would be just metered work that technically doesn't perform any user-facing role.
I've added the check in the lazy decode path that happens on reads. It's still a bit weaker than the old check, as we never verify the entries that are never read, but given that this is a sanity check for something that shouldn't actually happen, it should be good enough to catch potential bugs.

soroban-env-host/src/host/ledger_entry.rs
dmkozh: My interpretation is 'LedgerEntry XDR' (known as just `LedgerEntry` in code) vs 'host wrapper around `LedgerEntry`. But `LazyLedgerEntry` sounds fine to me as well.

soroban-env-host/src/e2e_invoke.rs
```diff
@@ -1021,100 +1058,114 @@ fn build_storage_footprint_from_xdr(
...
-                ScErrorType::Storage,
-                ScErrorCode::InternalError,
-            )
-            .into());
-        }
-        storage_map = storage_map.insert(key, Some((le, live_until_ledger)), budget)?;
+                None
+            }
+        };
+        initial_entry_sizes_for_rent.push(maybe_ttl_meta.map(|ttl| ttl.entry_size_for_rent));
```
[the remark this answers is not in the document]
dmkozh: Raw entry bytes do not necessarily represent the size for rent, which is why we rely on the size being (cheaply) computed by the embedder beforehand.

soroban-env-host/src/storage.rs
```diff
@@ -249,20 +253,54 @@ impl Storage {
...
+    ) -> Result<Option<HostEntryWithLiveUntil>, HostError> {
         let _span = tracy_span!("storage get");
         Self::check_supported_ledger_key_type(key)?;
         self.prepare_read_only_access(key, host)?;
         match self.map.get::<Rc<LedgerKey>>(key, host.budget_ref())? {
             // Key has to be in the storage map at this point due to
             // `prepare_read_only_access`.
             None => Err((ScErrorType::Storage, ScErrorCode::InternalError).into()),
-            Some(pair_option) => Ok(pair_option.clone()),
+            Some(maybe_entry) => Ok(maybe_entry.clone()),
```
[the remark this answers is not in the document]
dmkozh: It's not possible to validate the entries without decoding them, which is why the check is only done for the decoding path. Avoiding unnecessary decoding is the whole point of the change. Slightly weaker verification at runtime is a tradeoff, but it's not really significant, especially now that we're actually checking a good fraction of accesses (again, we trust the embedder to supply the correct data; sanity checks are best-effort).

soroban-env-host/src/budget.rs
jayz22, line 375: Why were these numbers changed? Did you recalibrate them?

soroban-env-host/src/host/data_helper.rs
```diff
@@ -596,21 +596,42 @@ impl Host {
...
+                .map
+                .iter()
+                .map(|(k, v)| {
+                    let decoded = match v {
+                        Some((e, live_until)) => Some((e.decoded(budget)?, *live_until)),
+                        None => None,
+                    };
+                    Ok((k.clone(), decoded))
+                })
+                .collect()
```
jayz22, line 612: nit: let's use `metered_collect` here

```diff
diff --git a/soroban-env-host/src/host/data_helper.rs b/soroban-env-host/src/host/data_helper.rs
@@
 #[cfg(any(test, feature = "testutils"))]
+use crate::host::metered_clone::MeteredIterator;
+#[cfg(any(test, feature = "testutils"))]
 use crate::storage::{AccessType, EntryWithLiveUntil};
@@
-                .collect()
+                .metered_collect::<Result<Vec<_>, _>>(self)?
```

soroban-env-host/src/budget.rs
jayz22: Sorry, didn't see your earlier comment which was already resolved. I will comments there.

jayz22: Did you recalibrate these numbers yourself? Normally we don't recalibrate those numbers after the original protocol, unless if something looks obviously off. I'm happy to revisit this one.

FWIW, You can override individual budget parameter in test using test-only functions. We have plenty of examples in the budget_metering test. Just to avoid committing parameters before they are calibrated.

dmkozh: As comment mentions, these numbers were wrong and not what we have on pubnet (and I've updated this particular cost type to match pubnet). I've just updated a single value to make the test sensible, but keep diffs smaller and more focused on the specific change in the PR. But we need to do a separate PR that updates *all* the numbers to match pubnet.

soroban-env-host/src/host/data_helper.rs
dmkozh: That's unnecessary, it's a test utility (ideally, we shouldn't meter these at all, but that would result in duplication or using shadow budget which would make the code more complex).

soroban-env-host/src/budget.rs
jayz22: Okay, I didn't realize this was copied from pubnet. This makes sense. We should keep them match.

soroban-env-host/src/host/data_helper.rs
jayz22: Sure. Okay to omit for testutils. I just saw a metered operation down below. Thought it would be good to keep consistent.

jayz22 (approved):

soroban-env-host/src/e2e_invoke.rs
```diff
@@ -244,24 +294,28 @@ fn get_ledger_changes(
...
-            Some(AccessType::ReadOnly) => {
+        match access_type {
+            AccessType::ReadOnly => {
                 entry_change.read_only = true;
             }
-            Some(AccessType::ReadWrite) => {
-                if let Some((entry, _)) = entry_with_live_until_ledger {
+            AccessType::ReadWrite => {
+                if let Some((entry, _)) = maybe_new_entry_with_live_until_ledger {
+                    let entry = entry.decoded(budget)?;
```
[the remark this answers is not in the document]
dmkozh: This is technically a valid idea, but it would only verify the RW entries that were never read and never changed, which is something that would very rarely happen in reality (if at all), so the additional invariant coverage surface is pretty much negligible and is not worthy of making the code more complex IMO.

soroban-env-host/src/host/data_helper.rs
dmkozh: Yeah, that's a valid observation. I think we shouldn't use metering in testutils where we can (like clone/new etc.), but also I wouldn't try to work around the metered methods (like map methods) for the sake of simplicity.

soroban-env-host/src/storage.rs
```diff
@@ -279,7 +281,26 @@ impl Storage {
     ) -> Result<Option<EntryWithLiveUntil>, HostError> {
         match self.try_get_host_ledger_entry(key, host)? {
             None => Ok(None),
-            Some((entry, live_until)) => Ok(Some((entry.decoded(host.budget_ref())?, live_until))),
+            Some((entry, live_until)) => {
+                let was_decoded = entry.is_decoded();
+                let decoded = entry.decoded(host.budget_ref())?;
+                if !was_decoded {
```
graydon, line 287: Perfect, thanks.

graydon (approved):
Yep, that should do it

PR #1702, opened 2026-07-09
Update the default budget cost types to match pubnet

dmkozh:
### What

Update the default budget cost types to match pubnet.

### Why

The default is only used in the tests, notably in the SDK tests, where stale calibration is misleading the users who want to estimate their budget. The error is mostly insignificant (probably except the XDR deserialization cost, which has already been fixed), but there is really no reason for this to not be more accurate.

### Known limitations

N/A


jayz22 (approved):
LGTM, thanks!

PR #1703, opened 2026-07-10
Implement CAP-85

dmkozh:
This implements [CAP-85](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0085.md), which adds support for the executables that refer to an external contract data entry containing the Wasm hash.

A new object type `ExecutableTag` is also introduced to mark the entries that can be used an external executable reference.


soroban-env-host/src/host.rs
```diff
@@ -3984,6 +3991,21 @@ impl Host {
                     )
                 })
             }
+            ContractExecutable::ExternalRef(external_ref) => {
+                let wasm_hash = self.resolve_external_ref_wasm_hash(&external_ref)?;
+                let key = self.contract_code_ledger_key(&wasm_hash)?;
+                let (_, live_until) = self
+                    .try_borrow_storage_mut()?
+                    .get_with_live_until_ledger(&key, self, None)?;
```
sisuresh, line 4000: Not sure where this is used, but should it return the min of the contract code and the ref entry?

soroban-env-host/src/host/frame.rs
```diff
@@ -750,6 +750,21 @@ impl Host {
                     || vm.invoke_function_raw(self, func, args, treat_missing_function_as_noop),
                 )
             }
+            ContractExecutable::ExternalRef(external_ref) => {
+                let wasm_hash = self.resolve_external_ref_wasm_hash(external_ref)?;
+                let vm = self.instantiate_vm(id, &wasm_hash)?;
```
sisuresh: Nit: This block is almost identical to the one above.

soroban-env-host/src/host/data_helper.rs
```diff
@@ -159,9 +162,96 @@ impl Host {
...
+                self,
+                (ScErrorType::Storage, ScErrorCode::MissingValue),
+                "Wasm does not exist",
+                wasm_hash
+            ));
+        }
+        Ok(())
+    }
+
+    pub(crate) fn executable_ref_from_inputs(
```
sisuresh, line 179: From claude - 
`
 data_helper.rs, executable_ref_from_inputs — No validation that executable_owner is a contract address. An account owner fails much later with a misleading Storage/MissingValue, and in recording mode the junk key lands in the simulated footprint. An explicit InvalidInput check here is cheap and makes the failure self-explanatory.
`

dmkozh: The error is not really misleading, it's not any different from trying to refer to an entry from a non-existent contract, or non-existent tag. We don't really want to make a distinction between the account and contract addresses in general (not to mention the hallucinated second part of the comment re footprint).

soroban-env-host/src/host.rs
dmkozh: This is just a test utility (to assert that a TTL extension happened). I think the logic is fine here, it returns exactly the contract code TTL as the name suggests, and it's possible to test TTL of any contract data with a separate check. That's not entirely consistent with the protocol changes around the TTL extensions, but I think it's fine given that the intentions are different (in case of the test util we want to be precise, in case of the TTL extensions we want the existing functions to do the most useful thing).

soroban-env-host/src/host/frame.rs
dmkozh: Refactored

sisuresh (approved):

PR #1704, opened 2026-07-10
CAP-85: Externally managed contract executables (productionization)

sisuresh:
## What

Productionizes CAP-85 (externally managed contract executables) behind the
`cap_0085_executable_ref` leaf feature, which `next` aggregates.

Building on the CAP-85 implementation, this branch:

- Gates the change on the `cap_0085_executable_ref` leaf feature (`next`
  aggregates it; `common/next` adds nothing behavioral beyond the leaf).
- Regenerates observation traces under `testutils`.
- cfg-splits every budget-sensitive test so the suite is green with the
  feature both **on** and **off** — the enlarged `ScVal` shifts metering by
  a few dozen cpu insns / 128 mem bytes per affected invocation
  (`invocation_metering`, `budget_metering`, recording-mode `e2e_tests`,
  `auth`, `lifecycle`, `stellar_asset_contract`).
- Bumps the shadow-budget limit in `hostile::excessive_logging`:
  externalizing a full-page diagnostic log now consumes ~1.05MB of shadow
  memory (vs ~0.95MB), so block 3's shadow mem limit goes 1M → 2M under the
  feature. The production budget is unaffected.
- Reorders the `cap_85_executable_reference` e2e tests' expected
  `ledger_changes` to the host's intentional read-only-first ordering.

## Related / depends on

- XDR variants: stellar/rs-stellar-xdr#553
- Tracks @dmkozh's draft #1703

## Verification

- `--features testutils` (feature **off**, full suite): **778 passed, 0 failed** (observations regenerated)
- `--features next` (feature **on**, full suite): **801 passed, 0 failed** (incl. the `cap_85_executable_reference` module)
- `--features cap_0085_executable_ref` (leaf): affected modules green

## Notes

- **Draft** pending the XDR dependency landing and @dmkozh's review.
- CI `clippy` may surface pre-existing lints from a newer clippy (1.94.x)
  across untouched test files. These are **identical with and without this
  branch** (verified by diffing the clippy output against the clean tree)
  and are unrelated to CAP-85.


sisuresh:
Downstream (this release run): rs-soroban-sdk SPIKE pins this rev — stellar/rs-soroban-sdk#1929

soroban-env-host/src/host/conversion.rs
```diff
@@ -534,6 +534,22 @@ impl Host {
...
                         ScVal::Address(addr.0.metered_clone(self)?)
                     }
+                    #[cfg(feature = "cap_0085_executable_ref")]
+                    HostObject::ExecutableTag(et) => {
+                        ScVal::ExecutableTag(et.0.metered_clone(self)?)
+                    }
+                    // `HostObject::ExecutableTag` is unconditional; without the
+                    // gated `ScVal::ExecutableTag` there is nothing to convert
+                    // into (such an object can't be created in this build).
+                    #[cfg(not(feature = "cap_0085_executable_ref"))]
```
dmkozh, line 544: Code like this is precisely the reason for why I don't think we should use feature gates in host. I'm not sure it's a safe to just introduce an unusable host object, and I don't like maintaining the unused code. Not to mention that this messes up with all the observations.

soroban-env-host/src/test/e2e_tests.rs
```diff
@@ -33,6 +33,14 @@ use crate::{
     Host, HostError, LedgerInfo,
 };
 use crate::{ErrorHandler, ModuleCache};
+// CAP-0085 XDR types used only by the `next`-gated `cap_85_executable_reference`
+// test module (via `use super::*`).
+#[cfg(feature = "next")]
```
dmkozh, line 38: We shouldn't use next gates in code - if the intention is to gate it (which I don't think we should do, but still), then these should be gated by the XDR feature.

soroban-env-host/src/test/e2e_tests.rs
```diff
@@ -882,6 +890,9 @@ fn test_wasm_upload_success_in_recording_mode() {
         }]
     );
     assert!(res.auth.is_empty());
+    #[cfg(feature = "cap_0085_executable_ref")]
```
dmkozh, line 893: This pattern is going to become unwieldy if we make just one more change that happens to change the budgets slightly, as we'll need to maintain a cartesian product of all the feature combinations here.

soroban-env-host/src/test/lifecycle.rs
```diff
@@ -1833,6 +1831,90 @@ mod cap_58_constructor {
                     params,
                 )
                 .unwrap();
+                #[cfg(feature = "cap_0085_executable_ref")]
```
dmkozh, line 1827: I've left a comment re observations in e2e_tests, but this is especially awful to maintain

soroban-env-host/src/host/comparison.rs
```diff
@@ -677,6 +692,9 @@ mod tests {
             .filter(|t| {
                 // bad tags can't be converted to ScVal
                 !matches!(t, Tag::Bad)
+                    // ExecutableTag <-> ScVal conversion is gated behind CAP-0085.
+                    && (cfg!(feature = "cap_0085_executable_ref")
```
dmkozh, line 696: This is quite nasty as well

dmkozh (commented):
I've left a few comments highlighting the issues I see with feature gating in host. I don't think it's a good idea to maintain a whole lot of semi-dead code and test cartesian products of features in case if we have more than 1 CAP to maintain. I would rather prefer to use good old version control for this and roll back the code on the off chance it's not longer necessary. Surely, one can argue that AI can do the dirty work of adding the gates, but as this PR proves, it's not really perfect, requires human intervention anyways, and makes subsequent work more complex and slow than necessary. Even if the code diverges slightly (e.g. we make a followup change), making an AI-assisted rollback would not be too hard, and it would be much more focused (like if the object + host fns are removed, most of the remaining CAP-specific code will just become unused, which is detectable by compiler)

soroban-env-host/src/host/conversion.rs
sisuresh: I'm not sure why it decided to unconditionally include `ExecutableTag`. That should be gated as well, but I see your point.

PR #1706, opened 2026-07-15
Pin ed25519-dalek to 2.x.y

### LINKED ISSUE #1705: Fresh SDK 27 testutils lock resolves incompatible ed25519-dalek 3
### What happened

A new project using `soroban-sdk = 27.0.0` with `testutils` fails to compile when Cargo creates a fresh lockfile. `soroban-env-host 27.0.0` declares `ed25519-dalek = ">=2.0.0"`, so Cargo now selects `ed25519-dalek 3.0.0` / `rand_core 0.10`. The host's `rand_chacha 0.3.1` uses `rand_core 0.6`, and `SigningKey::generate(chacha)` therefore fails its `CryptoRng` bound.

The same open-ended constraint is still present on `main`. The adjacent comments already state that the `ed25519-dalek`, `rand`, and `rand_chacha` versions must match.

### Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "soroban-sdk27-fresh-lock-repro"
version = "0.1.0"
edition = "2021"

[dev-dependencies]
soroban-sdk = { version = "=27.0.0", features = ["testutils"] }
```

`src/lib.rs`:

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn sdk_testutils_compile() {
        let _env = soroban_sdk::Env::default();
    }
}
```

From an empty directory with no lockfile:

```console
$ cargo test
error[E0277]: the trait bound `ChaCha20Rng: ed25519_dalek::rand_core::CryptoRng` is not satisfied
  --> soroban-env-host-27.0.0/src/builtin_contracts/testutils.rs:29:58
   |
29 | host.with_test_prng(|chacha| Ok(SigningKey::generate(chacha)))
   |                                 -------------------- ^^^^^^
```

Cargo resolves both `rand_core 0.6.4` (through `rand_chacha 0.3.1`) and `rand_core 0.10.1` (through `ed25519-dalek 3.0.0`).

### Confirmed workaround

Adding this direct dev dependency makes the same project compile and its test pass:

```toml
ed25519-dalek = "=2.2.0"
```

### Suggested fix

Bound the host dependency to the compatible major, for example `ed25519-dalek = ">=2.0.0, <3.0.0"` (or the project's preferred equivalent), and add a fresh-lockfile resolver check so a newly published incompatible major cannot silently break `soroban-sdk/testutils` again.

I can prepare a PR if this direction is acceptable.

### Environment

- `soroban-sdk 27.0.0`
- `soroban-env-host 27.0.0`
- Rust/Cargo stable on macOS arm64
- Reproduced July 15, 2026 from a fresh Cargo lockfile

kalepail:
### What

Bound `soroban-env-host`'s `ed25519-dalek` dependency to the compatible 2.x major.

### Why

The host uses `rand_chacha 0.3.1` / `rand_core 0.6` when it calls `SigningKey::generate` in test utilities. The existing `>=2.0.0` constraint lets a fresh lockfile select `ed25519-dalek 3.0.0` / `rand_core 0.10`, which makes `soroban-sdk 27.0.0` with `testutils` fail to compile due to an unsatisfied `CryptoRng` bound.

This is the narrow constraint fix discussed in #1705. The adjacent manifest comments already require these RNG-related versions to match.

Validation:

- Minimal fresh-lock SDK 27 reproduction fails before the constraint/workaround and passes with the 2.x line.
- Fresh dependency resolution with this manifest selects `ed25519-dalek 2.2.0`, not 3.0.0.
- `cargo test -p soroban-env-host --lib`: 766 passed, 0 failed.

Fixes #1705.

### Known limitations

The published `soroban-env-host 27.0.0` crate will still require a patch release (or a downstream direct pin/lockfile) for existing SDK 27 consumers to receive this bound.


soroban-env-host/Cargo.toml
```diff
@@ -24,7 +24,7 @@ sha2 = "0.10.8"
 hex-literal = "0.4.1"
 hmac = "0.12.1"
 # NB: We'll need to pin this again after switching the Core to the new env version.
-ed25519-dalek = {version = ">=2.0.0", features = ["rand_core"] }
+ed25519-dalek = {version = ">=2.0.0, <3.0.0", features = ["rand_core"] }
```
leighmcculloch: This definition simply needs to be the version, it is equivalent.
```suggestion
ed25519-dalek = {version = "2.0.0", features = ["rand_core"] }
```

When no operator is provided, then it is equivalent to using a caret `^2.0.0` which is equivalent to `>= 2.0.0, <3.0.0`.

Ref: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax

leighmcculloch (approved):
This looks good although we should simplify it per below.

leighmcculloch:
This change is released in [v28](https://github.com/stellar/rs-soroban-env/releases/tag/v28.0.1), and in backports to patch releases for [v27](https://github.com/stellar/rs-soroban-env/commit/fe0eb61a947e1b0394c51aa7f692105da5013a3f), [v26](https://github.com/stellar/rs-soroban-env/commit/e15e2f57be88c5fa5b40b8007df378abe09c7695), [v25](https://github.com/stellar/rs-soroban-env/commit/e4ae3a617da429a52547e35e2806290cf125e20c).

PR #1712, opened 2026-07-24
Refund VM linear-memory budget

### LINKED ISSUE #1707: Refund transient VM linear-memory budget at VM teardown
Tracking issue for https://github.com/orgs/stellar/discussions/1970

jayz22:
Resolves https://github.com/stellar/rs-soroban-env/issues/1707

### What

- A new `VmStoreData` type that tracks linear-memory allocation per VM.
- A new budget function that refunds memory.
- A new scoped function `Vm::execute_with_mem_refund` that runs a VM and refunds its memory when the scope exits, applied to every production path that instantiates a VM - contract invocation, the protocol-version probe, and wasm upload validation. As a result the memory cost of all of these is reduced.
- Budget's previous `total_count` replaced by two fields:`net_count` and `peak_count`. where the `net_count` tracks the point-in-time memory usage (subject to the refund), and `peak_count` checks the peak usage and is used for reporting.

### Why

The memory budget has always measured cumulative memory and never credits back transient memory that is later freed. For most invocations that is a fine proxy. But each VM instantiation charges ~1 MiB of linear memory (the toolchain-default shadow stack) that is freed at teardown yet never refunded, so deep or wide cross-contract calling diverges from peak and overcharges contracts for that memory - making contract interactions needlessly expensive and prone to hitting the memory limit.

The tracker lives in `VmStoreData` which is the object wasmi owns (originally just a `Host` handle for the metering callback), per-VM. The refund happens when the VM is dropped, guarded automatically by the scoped function (like a `Drop` impl, but fallible).

The budget keeps both counts (net and peak). Refunds lower the running net, so freed memory is reusable within the invocation and the budget bounds the peak memory rather than the cumulative sum. But net count under-reports what an invocation actually needed, so the peak high-water mark is what's reported and provisioned against.


soroban-env-host/src/budget/wasmi_helper.rs
```diff
@@ -49,15 +56,42 @@ impl ResourceLimiter for Host {
...
+            // VM linear memory is charged only to the real budget and refunded
+            // only from the real budget (see `Vm::execute_with_mem_refund`);
+            // guest memory only grows during instantiation or execution,
+            // neither of which runs in shadow mode. This is just a defensive
+            // check.
+            if self
+                .host
+                .as_budget()
+                .is_in_shadow_mode()
+                .map_err(|_| errors::MemoryError::OutOfBoundsGrowth)?
```
dmkozh: Are there any other error codes besides OutOfBoundsGrowth? This will hide an invariant error behind a budget error, which is not desirable.

Also, this check looks redundant to me TBH, I don't think this should care which budget mode is being used.

soroban-env-host/src/host/invocation_metering.rs
```diff
@@ -1396,7 +1396,7 @@ mod test {
         expect![[r#"
             InvocationResources {
                 instructions: 320615,
-                mem_bytes: 1135822,
+                mem_bytes: 87230,
```
dmkozh: I don't think we need to block the PR on this, but this change is quite misleading (we have hidden the invocation completely).
I think we should report max mem_bytes for any invocation. That might be slightly confusing with sub-invocations, but that's still better than completely removing most of the memory observations because VM frame gets popped when the invocation ends.

soroban-env-host/src/host/invocation_metering.rs
```diff
@@ -1874,7 +1874,7 @@ mod test {
...
+                                    StringM(invocation resource limits are exceeded: instructions: 1788420 > 10, memory bytes: 114969 > 20, contract events size bytes: 800 > 50, contract data key 'ContractData(LedgerKeyContractData { contract: Contract(ContractId(Hash(2e0ff7a55065f3a896723a964c3d9862a4722bfc77229fe4875f390ef2a0027e))), key: LedgerKeyContractInstance, durability: Persistent })' size: 48 > 25, contract data entry with key 'ContractData(LedgerKeyContractData { contract: Contract(ContractId(Hash(2e0ff7a55065f3a896723a964c3d9862a4722bfc77229fe4875f390ef2a0027e))), key: LedgerKeyContractInstance, durability: Persistent })' size: 104 > 35, contract code entry with key 'ContractCode(LedgerKeyContractCode { hash: Hash(1a2ee5a89dd2162a20e716b3e2de70a2d662480a9565350378661871bcf8e398) })' size: 1840 > 45),
```
dmkozh: This is a more concerning side-effect of the issue I've mentioned above: we use the invocation metering for the resource limit enforcement, so not using the max memory results in wrong checks. I think the prod memory reporting path is also broken because we don't track the max memory anywhere. Maybe that's worth addressing in this PR after all.

soroban-env-host/src/vm.rs
```diff
@@ -73,6 +73,31 @@ impl Drop for VmInstantiationTimer {
...
+/// `mem_bytes` charged to the budget for this store's linear memory (initial
+/// allocation + runtime grows), accumulated by the store's
+/// [wasmi::ResourceLimiter].
+///
+/// A VM's linear memory is transient -- it is freed when the VM is torn down
+/// -- but the charge accumulated here stays in the budget unless explicitly
+/// refunded. The only refund site is [Vm::execute_with_mem_refund], the
+/// canonical way to use a VM (see the rationale there); a VM used outside
+/// that scope (tests and benches) keeps its memory charged.
+pub struct VmStoreData {
```
dmkozh, line 87: I'm not sure what to think of this change.  It seems to create a weird coupling between the host and budget charges, but host itself also manages the budget. Compared to the sketch that I did with Claude, the budget itself is affected much less (though you're missing the max memory tracking  that we still need), but this is much more invasive change scope-wise. 
So I came to think that it would make more sense to keep the refund stack in the host/budget in some way. I'm not sure if my sketch has the best version of this, maybe we could try tying the refunds to a `Frame` (that's also going to be more scalable if in the future we decide to also drop, say, host objects allocated by a given frame). If that's inconvenient, I think maintaining a separate stack is fine as well - we do that for a few different per-frame entities (like auth managers).

soroban-env-host/src/test/lifecycle.rs
```diff
@@ -2075,7 +2075,7 @@ mod cap_58_constructor {
                                 ),
                                 resources: SubInvocationResources {
                                     instructions: 546289,
-                                    mem_bytes: 1175536,
+                                    mem_bytes: 0,
```
jayz22, line 2078: The first sub-call drove the memory peak up and refunded on exit and the second sub-call here started from the lower base and at the end the peak memory did not exceed the previous peak that's why `mem_bytes` is zero.

jayz22: (Invocation metering's `mem_bytes` computation is already doing saturating sub, so there's no underflow risk here. )

soroban-env-host/src/host/invocation_metering.rs
jayz22: fixed (same as above)

soroban-env-host/src/vm.rs
jayz22: This change does not add any new couplings (it preserves the existing dependency relations). In Wasm metering, wasmi owns the structure which implements the `ResourceLimiter` trait, used for callback to perform metering. Previously that structure is the `Host`. With this change this now it is a dedicated type `VmStoreData` that wraps the host (for the same callbacks) plus a tracker for memory allocation per VM. This is actually the right abstraction level for the memory tracking, and the memory tracking updates and refunding occurrence matches *almost* precisely with the linear memory lifetime. (If we could do `Drop` trait (we can't because it's infallible), this would match even better and more cleanly, and we have no need for the closure guard).

soroban-env-host/src/host/invocation_metering.rs
jayz22: fixed it to use peak memory count for reporting

soroban-env-host/src/test/budget_metering.rs
```diff
@@ -552,3 +569,208 @@ fn total_amount_charged_from_random_inputs() -> Result<(), HostError> {
...
+    assert_eq!(budget.get_mem_bytes_consumed()?, after_charge);
+    // Refund is saturating: refunding more than consumed floors the net at 0
+    // (and still does not touch the peak).
+    budget.refund_mem(u64::MAX)?;
+    assert_eq!(budget.get_mem_bytes_net()?, 0);
+    assert_eq!(budget.get_mem_bytes_consumed()?, after_charge);
+    Ok(())
+}
+
+/// A VM's linear memory is charged during the call but refunded at teardown,
```
dmkozh: nit: I think all the comments on top of tests are redundant as the test names are pretty self-descriptive (also I assume these are AI-generated, I don't see much value in bloating the code base with trivial blurbs)

soroban-env-host/src/budget/wasmi_helper.rs
jayz22: dropped the check, it was purely for defensive

soroban-env-host/src/test/budget_metering.rs
```diff
@@ -552,3 +569,208 @@ fn total_amount_charged_from_random_inputs() -> Result<(), HostError> {
...
+        "peak {peak} should equal net {net} + refunded VM memory {vm_page}"
+    );
+    Ok(())
+}
+
+/// Calling the same contract sequentially does not accumulate VM linear memory
+/// in the budget: each call's VM memory is refunded before the next runs, so
+/// the second call adds its own host allocations but not another VM page.
+#[test]
+fn sequential_calls_do_not_accumulate_vm_linear_memory() -> Result<(), HostError> {
```
dmkozh: nit: this test and cross_contract_call_refunds_combined_vm_linear_memory are identical coverage-wise; I suggest we remove one of these (maybe this one, as it does a more realistic cross-contract call).

soroban-env-host/src/test/budget_metering.rs
```diff
@@ -552,3 +569,208 @@ fn total_amount_charged_from_random_inputs() -> Result<(), HostError> {
...
+    );
+    Ok(())
+}
+
+/// Creating a contract whose executable is an `ExternalRef` probes the
+/// referenced wasm's protocol version by instantiating a throwaway VM
+/// (`get_contract_protocol_version`); that probe VM's linear memory must be
+/// refunded just like on the `Wasm`-executable path.
+#[test]
+fn external_ref_probe_vm_linear_memory_refunded() -> Result<(), HostError> {
```
dmkozh: nit: This test is not really necessary, it covers exactly the same code path as the normal create contract code path (because in either test we need to get the contract protocol version and maybe call the constructor)

soroban-env-host/src/test/budget_metering.rs
```diff
@@ -552,3 +569,208 @@ fn total_amount_charged_from_random_inputs() -> Result<(), HostError> {
...
+    assert_eq!(budget.get_mem_bytes_net()?, 0);
+    assert_eq!(budget.get_mem_bytes_consumed()?, after_charge);
+    Ok(())
+}
+
+/// A VM's linear memory is charged during the call but refunded at teardown,
+/// so it is not part of the net memory consumed after the call returns.
+#[test]
+fn vm_linear_memory_refunded_on_teardown() -> Result<(), HostError> {
+    use crate::testutils::wasm::wasm_module_with_4n_insns;
```
dmkozh: nit: Can we just add all use statements to the top of the file, as we normally do with all the tests? There is really no value in defining these per-test

soroban-env-host/src/test/invocation.rs
```diff
@@ -56,21 +56,22 @@ fn invoke_alloc() -> Result<(), HostError> {
...
     assert!(res.shallow_eq(&8128_u32.into()));
-    let used_bytes = host.budget_cloned().get_mem_bytes_consumed()?;
-    // The general pattern of memory growth in this contract will be a sequence
-    // of vector-doublings, but these are masked by the fact that we only see
-    // the calls that cause the backing vector of wasm linear memory to grow,
-    // which happens as the guest vector crosses 64k boundaries (and eventually
-    // starts growing in steps larger than 64k itself).
+    let peak_bytes = host.budget_cloned().get_mem_bytes_consumed()?;
+    let net_bytes = host.budget_cloned().get_mem_bytes_net()?;
+    // This contract grows its WASM linear memory via a sequence of
```
dmkozh: nit: Wasm, not WASM

soroban-env-host/src/test/lifecycle.rs
dmkozh: Yeah, I think this behavior is fine - just show the peak increase in sub-invocation, so if peak doesn't move, it will end up being 0.

soroban-env-host/src/test/budget_metering.rs
jayz22: done

soroban-env-host/src/test/budget_metering.rs
jayz22: agree, removed this test

soroban-env-host/src/test/budget_metering.rs
jayz22: This test exercises a different branch (ExternRef vs Wasm), so I would keep it. But I cleaned it up to make it slightly simpler.

soroban-env-host/src/test/budget_metering.rs
jayz22: done

soroban-env-host/src/test/invocation.rs
jayz22: done

soroban-env-host/src/vm.rs
dmkozh: Alright, I'm still not entirely sure if that's the best alternative, but I also think it's good enough for now.

soroban-env-host/src/test/budget_metering.rs
dmkozh: I just want to point out that while these are technically different branches, they exercise exactly the same underlying VM-logic (i.e. the branching part comes down to how we fetch the Wasm, then after we have fetched Wasm it doesn't really matter how exactly that happened). Of course, one can argue that slightly different code is executed across the stack, but the same can be said about pretty much any input combination (like we could technically test G vs C deployer addresses etc. - the parameter space can be exponentially big if we want to). But leaving this up to you.

dmkozh (approved):

graydon (approved):
I have a slight preference for putting the tracking variable in Frame::ContractVM but I understand that's clunky in a different way and misses a few edge cases of its own, and .. this is acceptable as-is, ought to work. I guess it's ok for now; worst case we revisit at a future protocol boundary.

graydon:
subsumed by https://github.com/stellar/rs-soroban-env/pull/1716

PR #1699, opened 2026-07-02
VM memory refund sketch

dmkozh:

PR #1716, opened 2026-07-29
Refund VM linear memory (take 3)

### LINKED ISSUE #1707: Refund transient VM linear-memory budget at VM teardown

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-env/issues/1707 (see previous takes [1](https://github.com/stellar/rs-soroban-env/pull/1699) and [2](https://github.com/stellar/rs-soroban-env/pull/1712)). 
This approach requires no new data type (`VmCallerData` in take 2) or additional stacks (take 1). It works within the existing `frame` design. Here is the mechanism: 
- `Context` gains a `linear_mem_charged` counter.
- `memory_growing` attributes each real-budget `MemAlloc` charge to the current top frame. During execution that is the growing VM's own frame; during instantiation the VM has no frame yet, so the charge lands on the caller and `call_contract_fn` reclaims that delta onto the VM's own frame when it is pushed.
- `pop_context` refunds the popped frame's counter

#### Limitation:
This approach depends on the existence of a frame
- it doesn't cover one-off VM instantiation such as during upload contract with no frame.
- most tests invoke contracts using `host.call` with no top frame (traces show no top-level refund, which is correct).

Both can be done as follow-ups.


soroban-env-host/src/host/frame.rs
```diff
@@ -403,6 +453,22 @@ impl Host {
...
+        F: FnOnce() -> Result<Val, HostError>,
+    {
+        self.with_frame_seeded(0, frame, f)
+    }
+
+    /// Like [`Host::with_frame`], but seeds the pushed [`Context`]'s
+    /// `linear_mem_charged` with `linear_mem_charged`. Used by
+    /// [`Host::call_contract_fn`] to hand a freshly instantiated VM's initial
+    /// linear-memory charge to the VM's own frame.
+    fn with_frame_seeded<F>(
```
graydon: Could you rename this from "seeded" to something else? So far we only use the word "seed" to describe PRNG state, and this will confuse readers. Maybe `with_frame_and_initial_mem_charge` ?

graydon (changes requested):
One minor change, but yes, I like this approach much more!

soroban-env-host/src/host/frame.rs
jayz22: done

graydon (approved):

soroban-env-host/src/host/frame.rs
```diff
@@ -752,9 +819,17 @@ impl Host {
                 );
             }
         };
+        // The VM's initial linear memory is allocated (and charged) during
```
dmkozh, line 822: I don't quite follow this, aren't VM memories independent? Also, how does this interact with VM re-entrance (which we do allow in a narrow case of direct self-reentrance)

dmkozh: Actually, the we don't re-enter the VM, but the contract, it's still using a separate VM, so that's not an issue. But I still don't quite understand the 'handover' part.

graydon: They are independent. It's just that while VM #1 is on top of the frame stack we're and still _building_ VM #2 along with its frame, the memory allocation for VM #2 will get charged to VM #1's frame. So this is just transferring that charge to VM #2's frame constructor.

dmkozh (approved):

PR #1719, opened 2026-08-04
Synchronize budget with invocation resource limits.

dmkozh:
### What

Synchronize budget with invocation resource limits.

### Why

Due to an oversight these were disjoint, and thus e.g. disabling invocation resource limits would still not allow a test that requires too many instructions to pass.

https://github.com/stellar/rs-soroban-sdk/issues/1949

### Known limitations

The relationship is intentionally one-way: invocation resource limits change the budget, but budget can be updated independently. The reason for this is that SDK doesn't really need to mutate the budget directly anymore - converging everything under more broad umbrella makes the interface cleaner. However, making the mutating methods on budget private would require a major release, which is why we need to postpone it for now.


jayz22 (approved):

leighmcculloch (commented):
Nice!

PR #1720, opened 2026-08-05
Make native test contracts more similar to Wasm contracts.

dmkozh:
### What

Currently test contracts emulate being a Wasm replacement only via pointing at an empty Wasm entry. The call routing to native contracts bypasses the executable machinery and just uses a direct per-instance lookup.

With this change every registered native contract gets its own unique Wasm entry. The contract call routing is also changed to be always based on the actual Wasm hash (instead of per-contract override registry that we've used to have). Basically, the distinction between Wasm and native contracts is moved as far in the flow as possible.

Besides this structural change, this also addresses a few minor issues: we now avoid unnecessary storage lookups for the stub Wasms, and also rollback storage on constructor failures (which was an issue due to the SDK ability to call host fns outside of a frame).

### Why

This brings the native contract lifecycle in tests closer to the Wasm contract lifecycle and enables native contracts to be used in pretty much any context a Wasm contract can be used, such as contract factories that create contract instances based on a Wasm hash, or creating a native contract instance with constructor to check the constructor logic separately (https://github.com/stellar/rs-soroban-sdk/issues/1945)

### Known limitations

While this change only touches the code guarded with the test/testutils feature, it unfortunately changes the observations. This happens because observations are generated in the test mode, and we change the metering of the test-only operations here. It is not hard to prove that non-test code behavior hasn't changed, and also we're not really going to pick this change up in Core anyways.


soroban-env-host/src/host/lifecycle.rs
```diff
@@ -378,10 +407,141 @@ impl Host {
...
+    ///
+    /// Test contracts are stored in the host storage as Wasm code entries with
+    /// a short identifier in place of the actual Wasm code.
+    ///
+    /// The returned hash can be used in the same way as the regular Wasm hash,
+    /// so it's possible to create contract instances that refer to this hash.
+    pub fn upload_test_contract_wasm(
+        &self,
+        contract_fns: Rc<dyn ContractFunctionSet>,
+    ) -> Result<BytesObject, HostError> {
```
leighmcculloch: It looks like the host controls what wasm hash gets assigned? Can we let the caller, the sdk, choose the wasm hash? That way we can use this in the sdk as a mechanism to replace a wasm contract in a loaded ledger snapshot.

leighmcculloch: > That way we can use this in the sdk as a mechanism to replace a wasm contract in a loaded ledger snapshot.

This is functionality that exists today at the contract instance level, you can replace the executable of a contract address with a native impl. It supports taking a ledger snapshot and doing additional testing, or taking a ledger snapshot and doing testing with the native impl so you can do step through debugging.

I'm being opportunistic, but this is a great opportunity for us to bring that same capability a layer lower to the definition of the executable itself.

dmkozh: I don't think that's a good idea. Wasm hash is the hash of the underlying Wasm blob, so overriding would break the invariant. I'm also not sure I understand the reason for why SDK would need to replace a real Wasm blob with a native implementation. Wasms are by design immutable, and making them mutable seems like a footgun to me. It's already possible to override the executable of any contract in the tests, and overriding it with a hash of a native contract 'Wasm' seems much cleaner and more idiomatic.

leighmcculloch: > Wasm hash is the hash of the underlying Wasm blob

In this case where there is no wasm blob and instead a native contract, how does that invariant get maintained?

> Wasms are by design immutable, and making them mutable seems like a footgun to me.

This is in the test environment, not for production.

> It's already possible to override the executable of any contract in the tests, and overriding it with a hash of a native contract 'Wasm' seems much cleaner and more idiomatic.

It can be difficult to do this for multi contract systems where factories manage an existing wasm to deploy.

dmkozh: > This is functionality that exists today at the contract instance level, you can replace the executable of a contract address with a native impl. It supports taking a ledger snapshot and doing additional testing, or taking a ledger snapshot and doing testing with the native impl so you can do step through debugging.

Right, this makes sense to me at the instance level, but I think this is the right layer to perform this override, because the contract instances are mutable in general. Wasms are immutable, and also are not callable by themselves.  So I'm not sure I see a good case for when overriding the Wasm itself would be the right approach vs overriding the executable pointer inside the instance.

dmkozh: > In this case where there is no wasm blob and instead a native contract, how does that invariant get maintained?

There is still a unique, special Wasm blob that identifies the native contract, and we hash it in the same way as we hash the regular Wasms.

> This is in the test environment, not for production.

Sure, but that doesn't mean that we should expose potentially misleading patterns without a good reason, or provide multiple marginally distinct ways of achieving the same end result.

> It can be difficult to do this for multi contract systems where factories manage an existing wasm to deploy.

In case of a factory it seems like it should be sufficient to just update the Wasm hash in the factory config - the factory can't own or manage the Wasm itself. It also seems more convenient - there is no need to load the 'real' Wasm in the first place. This would only not work if the factory hardcodes the expected hash, which doesn't really seem like a reasonable use case to be motivational here (though even then one could natively compile and modify the factory itself).

leighmcculloch: I think we're already doing enough here that's removed from production behaviour to be misleading. Automatically generating a wasm hash from a native contract that won't match its final wasm hash could be misleading, and letting a developer set the arbitrary hash doesn't seem more or less misleading. Letting a developer set the wasm hash may be less misleading because it demonstrates the arbitrary nature of the hash in use.

I don't think we need to hold back this feature out of that concern.

Is there a technical reason we're not able to let the developer choose the wasm hash?

I'm imagining an SDK API that parallels the existing `register`/`register_at` functions.

For example:

New functions:

```rust
env.upload(Contract) -> BytesN<32>
```
```rust
env.upload_at(Into<BytesN<32>>, Contract) -> BytesN<32>
```

Existing functions:

```rust
env.register(Contract, (...)) -> Address
```
```rust
env.register_at(Address, Contract, (...)) -> Address
```

dmkozh: > Automatically generating a wasm hash from a native contract that won't match its final wasm hash could be misleading

It makes sense to me that the native and Wasm contracts have different hashes (the same way that e.g. optimized and non-optimized Wasms have different hashes). Hash is not a footprint of the contract behavior, it's a footprint of a concrete implementation

> Letting a developer set the wasm hash may be less misleading because it demonstrates the arbitrary nature of the hash in use.

I agree that the hash is arbitrary in the sense that you shouldn't worry about its exact value. But I would take this as an argument against overriding it, because the logic shouldn't depend on any specific hash value.

I'm thinking about Wasms as of literals in programming language terms. It doesn't make much sense to be able to override a literal, even in tests. Again, I'm not sure if 'it's for testing' argument is sufficient in a vacuum. Like we could arbitrarily allow developers to override the behavior of the host functions, or mutate the host objects in test environment. But I don't see a reason for why we should actually think that it's a good idea (and conversely if there are compelling use cases we could actually consider doing that).

> Is there a technical reason we're not able to let the developer choose the wasm hash?

Depending on what do you mean by a technical reason. To me, the reason is that normally Wasm hash is literally the hash of the Wasm. Like is there a technical reason to not let the developer choose what sha256 function returns in tests? It's not impossible, but it's weird and brings up new questions and edge cases that are exclusive to the test environment. I'd like to reduce the amount of those when possible.

> I'm imagining an SDK API that parallels the existing register/register_at functions.

Should we also allow uploading different Wasms under different hashes then? If we think that mocking out the actual Wasm implementation is useful, then it seems logical. It would be even more annoying to maintain though. But I don't really see a compelling use case yet.

leighmcculloch: > Should we also allow uploading different Wasms under different hashes then?

That could be useful, but I think the native use case is more useful because we can use it to replace a contract executable with an executable that supports step-through-debugging, without recalling the constructor and otherwise disturbing the contract instance storage.

So I don't think we need to expand the scope of this work to also support upload wasm to hash.

> It's not impossible, but it's weird and brings up new questions and edge cases that are exclusive to the test environment. I'd like to reduce the amount of those when possible.

The SDK already allows developers to take this same action with other entities, choosing contract IDs, account IDs. It's not _new_ weird.

I think either way the env is not the best place to make this call. This function can take in a hash — that seems like a very small change — and the SDK can determine the best way to expose or not expose hash choice to the developer, and how tooling might structure that in a way that is useful and as not misleading as possible.

There's another reason to let the SDK choose the ID. It's what we do with contracts, the SDK chooses contract IDs and account IDs. And that allows the SDK to deterministically select IDs to ensure every test run is reproducibly identical resulting in the same ledger state. The SDK cares about the test environment being deterministic because it generates ledger snapshots automatically for differential testing.

dmkozh: > This function can take in a hash — that seems like a very small change — and the SDK can determine the best way to expose or not expose hash choice to the developer, and how tooling might structure that in a way that is useful and as not misleading as possible.

This function doesn't just map a hash to a native contract, it also creates and uploads a stub entry for the native Wasm. I think it's reasonable to expect that this will satisfy the developers' test needs most of the time. For the override case is valuable, I suggest we add another function that maps an *existing* Wasm entry to the provided native contract (which seems to be what we want want for this use case). This way we do not break the hash<->contents matching invariant, and the intent of both functions is clear.  Does that make sense to you?

> There's another reason to let the SDK choose the ID. It's what we do with contracts, the SDK chooses contract IDs and account IDs. And that allows the SDK to deterministically select IDs to ensure every test run is reproducibly identical resulting in the same ledger state. The SDK cares about the test environment being deterministic because it generates ledger snapshots automatically for differential testing.

I'm not sure what does this discussion have to do with determinism, the hashes are deterministic. Also, it makes sense for every unit test to build the set of mock accounts in the same fashion (because normally addresses can be completely arbitrary), but it doesn't make sense for every test to deal with the same set of Wasm hashes (because all Wasms are functionally different from each other). We also don't really want different Wasms to yield the same snapshots.

leighmcculloch: > the hashes are deterministic

👍🏻 Ok I see now the generated wasm hash is deterministic based off the order contracts are registered, that's good. I thought the fn ptr was used in deriving the hash, but I see now it's just used in lookup.

leighmcculloch: > This function doesn't just map a hash to a native contract, it also creates and uploads a stub entry for the native Wasm.

It can override the existing entry or leave it there?

leighmcculloch: > we add another function that maps an _existing_ Wasm entry to the provided native contract

If we add a fn that is purely for replacing, we will have one fn for upload to a generated address, one fn for replacing, but none for upload to a new arbitrary hash that has no contract. The upload_at sdk function would ideally work like register_at, let uploading to any hash existing or not. It'd be surprising if upload_at errored if it wasn't a replacement.

Would the new fn support registering the same native contract for two wasm hashes? I think the code in the PR would need to remove the fn ptr and the invariant then anyway. It will be surprising if calling upload_at twice with different ids errored.

I think we might see some issues if someone reloads a snapshot too and tries to do more work after a reload. This has always been somewhat odd thing to do with native contracts involved anyway, but the behaviour will be even more odd with the implementation proposed here because the Env's deployment counter will reset to 0 where-as the SDK's id infra already handles this case.

The native contracts having to adhere to the invariant seems unnecessary, and creates more odd edge cases where the sdk fns will have limitations that don't seem logical from a dev just trying to test. None of these limitations exist if the env lets the sdk register the native contract at any hash.

Can we keep the env impl simpler, remove the fn ptr lookup, remove the built in hash generation on the env side, and the env can accept a hash when registering the wasm? It looks like the code will get significantly simpler: https://github.com/dmkozh/rs-soroban-env/compare/native_wasm_id...stellar:rs-soroban-env:native-contract-explicit-wasm-hash?w=1

This will provide more freedom to the SDK solve these problems in a more holistic consistent way with the existing test functionality and where the above edge cases work. The SDK can assign wasm hashes automatically and offer registration like it does contract addresses, it already has the infra to support that.

dmkozh: Alright, this seems fine to me. I've adjusted this a little bit to make the names a bit more specific, and also updated how we generate the stubs for the contract instance registration, on the off-chance anyone needs to observe them for debug or in the snapshots.

leighmcculloch:
Fyi I ran a spike using this change to see if there's anything I'm missing, and I think so far looks great:
- https://github.com/stellar/rs-soroban-sdk/pull/2007

Ran out of time to finish reviewing this today and will return tomorrow so we can merge this asap.

soroban-env-host/src/host/frame.rs
```diff
@@ -957,24 +1094,27 @@ impl Host {
...
-            ContractExecutable::StellarAsset => self.get_ledger_protocol_version(),
+            ContractExecutable::StellarAsset => return self.get_ledger_protocol_version(),
+        };
+        #[cfg(any(test, feature = "testutils"))]
+        if self
+            .try_borrow_test_contract_registry()?
+            .get_contract_fn_set(wasm_hash)
+            .is_some()
+        {
+            return self.get_ledger_protocol_version();
```
leighmcculloch, line 1114: Because the call to check with the test contract registry happens after the other checks now, I think this changes the behavior in the test environment if a test contract was registered over top of a stellar asset contract address. Or maybe that was never supported. I'm wondering what other impact this has.

Why does the test contract check come later rather than still happen first?

leighmcculloch (approved):
I just had one question inline wrt a possible behaviour change. I haven't checked the behaviour in the SDK to verify if it is indeed a behaviour change, but will do later today when I get a chance. But I'm guessing you already know if it is or not?

soroban-env-host/src/host/frame.rs
dmkozh: If you write a test contract over a SAC address, you'll end up with a Wasm executable (i.e. it's not possible to have a test contract with StellarAsset executable), so behavior doesn't really change. The current approach is just cleaner, as we override Wasms directly instead of using a combination of contract-id based registry and Wasm-based markers.

soroban-env-host/src/host/frame.rs
leighmcculloch, line 1191: It looks like `fn_return_diagnostics` gets called twice for native contract frames with the second occurrence being here. Because it's now called inside the callgraph below `call_contract_fn` -> `call_test_contract_fn` and here.

soroban-env-host/src/host/frame.rs
```diff
@@ -811,21 +923,46 @@ impl Host {
...
+        // after the executable has been resolved, so that it's possible to
+        // switch the compiled-in implementation back to Wasm (and the other way
+        // around) via `update_current_contract_wasm`.
+        // "testutils" is not covered by budget metering.
+        #[cfg(any(test, feature = "testutils"))]
+        if let Some(contract_fns) = self
+            .try_borrow_test_contract_registry()?
+            .get_contract_fn_set(wasm_hash)
+        {
+            return self.call_test_contract_fn(
```
leighmcculloch, line 950: I'm not sure but I think this piece of code could panic in some situations, because the borrow lasts for the whole `call_test_contract_fn` which would prevent borrowing from within in some situations.

In the code that was refactored/moved, it had this warning comment:
```rust
        if self.is_test_contract_executable(id)? {
            // This looks a little un-idiomatic, but this avoids maintaining a borrow of
            // self.0.contracts. Implementing it as
            //
            //     if let Some(cfs) = self.try_borrow_contracts()?.get(&id).cloned() { ... }
            //
            // maintains a borrow of self.0.contracts, which can cause borrow errors.
            let cfs_option = self.try_borrow_contracts()?.get(&id).cloned();
            if let Some(cfs) = cfs_option {
```

In the refactor the 'is a test contract' check was moved here and the borrow was rewritten in away that didn't preserve the way it was written before.

In general we should avoid borrowing in `if let` statements do to the potential for the borrow to be held for longer than we intended.

leighmcculloch (commented):
A couple things I had meant to post in my previous review but missed:

soroban-env-host/src/host/lifecycle.rs
```diff
@@ -378,10 +409,164 @@ impl Host {
...
+                    output
+                },
+            );
+            let wasm = format!("{TEST_CONTRACT_WASM_OVERRIDE_PREFIX}{hash_hex}").into_bytes();
+            (wasm, wasm_hash)
+        } else if let Some(contract_address) = contract_address {
+            let address_strkey = self.non_muxed_sc_address_to_strkey(&contract_address)?;
+            let wasm =
+                format!("{TEST_CONTRACT_WASM_FOR_CONTRACT_ID_PREFIX}{address_strkey}").into_bytes();
+            let wasm_hash = Hash(<Sha256 as sha2::Digest>::digest(wasm.as_slice()).into());
```
leighmcculloch, line 519: So something I just realised is that because the SDK needs to know what wasm-hashes are native contracts for the snapshot source filtering (https://github.com/stellar/rs-soroban-sdk/issues/1635) the SDK needs to know what wasm-hash was assigned when registering a native contract.

The SDK can run the same sha256 operation, but it'd be better if we could expose the hash assigned during registration. Or maybe we should have the SDK do that since the SDK will be doing that for the upload cases.

@dmkozh thoughts?

leighmcculloch: In https://github.com/stellar/rs-soroban-sdk/pull/2007 I've had the sdk just calculate what the hash is.

dmkozh: I didn't want to change the interface of the existing function. We could do a breaking change, if that helps (we'd need to bypass semver though...), but I'm not sure why do you need to infer the hash at all - you can just query it via get_address_executable as for any other contract. That seems like a simpler and more generic approach.

soroban-env-host/src/host/frame.rs
dmkozh: Good catch, the code looked fine on its own as this borrow behavior is non-obvious.

soroban-env-host/src/host/lifecycle.rs
leighmcculloch: It is possible to call `get_address_executable` after registration so we can filter after registration, but the problem is that during registration the snapshot source gets queried. We need to know the wasm hash before hand. 

For now I think we can have the sdk mirror the wasm hash sha256 logic. Maybe in v29 we change the api so the register sdk fns to do the upload step then call register on the host to create only the instance.

dmkozh: > but the problem is that during registration the snapshot source gets queried.

That's a problem that I've intended to solve with this PR - I really wanted to get rid of all the snapshot hacks. I've realized though that this has been violated with the refactor, as we've added a full storage `has` lookup. I've updated the code to bypass the snapshot, so now I don't believe we should lookup test Wasm snapshots in the test machinery. Is that sufficient to remove the snapshot logic on the SDK side and not worry about the hash? I would really prefer to solve this problem at the root.

leighmcculloch: That works and is better 👍🏻

leighmcculloch: While we're here would it also be possible to make it so that wasm's registered via the sdk also don't hit up the snapshot source during registration? We don't have to solve for this now, it's an existing behaviour, but just realising it and mentioning in case it is it's convenient to do so while we're changing this behaviour for native.

leighmcculloch (approved):

soroban-env-host/src/host/lifecycle.rs
dmkozh: > While we're here would it also be possible to make it so that wasm's registered via the sdk also don't hit up the snapshot source during registration?

We could try that, but I would do that as a followup. The regular Wasm contract uploads go through the shared host fn, so we need to refactor this carefully.

soroban-env-host/src/host/frame.rs
dmkozh: Interesting, it seems like it has always been like that. I've removed the unnecessary test fn call.

soroban-env-host/src/host/lifecycle.rs
leighmcculloch: I've opened an issue to track:
- https://github.com/stellar/rs-soroban-env/issues/1725

PR #1729, opened 2026-08-28
Clamp extend_to in extend_ttl (v1) instead of overflowing to InternalError

### LINKED ISSUE #1728: extend_ttl (v1) returns Context::InternalError on user-controlled u32 overflow of extend_to
## Summary

A contract can force `Error(Context, InternalError)` by calling `extend_contract_data_ttl`
on a **persistent** entry with a large `extend_to` (e.g. `u32::MAX`): v1
`Storage::extend_ttl` adds `extend_to` to the ledger sequence *before* clamping to the max
network TTL, so the addition overflows. The branch's "network misconfiguration" comment
does not hold here — `extend_to` is an unvalidated contract-supplied argument.

Consensus is unaffected (deterministic, reflected in the ledger hash), but every node that
applies or replays the ledger logs a `[Tx ERROR]` and bumps stellar-core's
`ledger.transaction.internal-error` counter, which is commonly wired to validator alerting.

## Root cause

`soroban-env-host/src/storage.rs`:

```rust
let mut new_live_until = host.with_ledger_info(|li| {
    li.sequence_number.checked_add(extend_to).ok_or_else(|| {   // <-- overflows here
        HostError::from(Error::from_type_and_code(
            ScErrorType::Context, ScErrorCode::InternalError))
    })
})?;
new_live_until = new_live_until.min(ttl_ext_info.max_live_until);   // <-- too late
```

The only pre-add guard is `matches!(durability, Temporary) && extend_to > max_ttl`, so
`Persistent` falls straight through. `extend_ttl_v2` clamps before adding and is correct.

jayz22:
### What

Closes #1728.

Use `saturating_add` instead of `checked_add` when computing `new_live_until` in `Storage::extend_ttl` (v1). Saturating is identical to `checked_add` in every non-overflowing case, because the sum is immediately clamped to `max_live_until` on the next line.


soroban-env-host/src/storage.rs
```diff
@@ -607,6 +607,13 @@ impl Storage {
 
         let ttl_ext_info = self.prepare_extend_ttl(host, &key, extend_to, key_val)?;
 
+        // Saturating is safe because the result is immediately clamped to
+        // `max_live_until` below
+        #[cfg(feature = "next")]
+        let mut new_live_until =
```
dmkozh, line 612: We don't need to guard this with 'next', p28 host has already been cut (in the worst case, SDK may pick this up, but we don't really care).

jayz22: Yeah, I was wondering about that. Should we also update the next ledger protocol version? 
```
#[cfg(feature = "next")]
pub(crate) const LEDGER_PROTOCOL_VERSION: u32 = 29;
```

dmkozh: I don't think it's necessary to do that in this particular PR, we can bump it later and as a separate PR.

jayz22: done

dmkozh (approved):

soroban-env-host/src/test/storage.rs
```diff
@@ -842,3 +842,71 @@ mod ttl_extension_v2_tests {
         assert_eq!(extend(ContractTtlExtension::Code, 10000), (0, 500));
     }
 }
+
+mod ttl_extension_v1_overflow_tests {
```
dmkozh: We don't need this new test, I would rather extend the existing one (it's also going to be more comprehensive, as we test all storage types)

dmkozh (changes requested):

soroban-env-host/src/test/storage.rs
jayz22: done

dmkozh (approved):

